PR #18, opened 2020-04-08
Request header

Grumpf86:
> 91.61.47.19 - - [16/Mar/2020:14:47:40 +0100] “GET /api/v0/product/4311501642467.json HTTP/1.1” 200 3322 “-” “Dart/2.7 (dart:io)”
> 91.61.47.19 - - [16/Mar/2020:14:47:40 +0100] “GET /api/v0/product/4010355107077%0A.json HTTP/1.1" 200 4517 “-” “Dart/2.7 (dart:io)”
> Can you send the name of Tioli instead of “Dart” ?

Hey, here comes the possibility to add an individual "user-agent" and "from"-field to the request header. So we can send the name of the app (e.g. TIOLI) not only as a product comment, but also as a header value. I've spread these changes over all existing requests.
Now it's mandatory to add a user to each request. the user already has a userId (mailaddress) and a comment-field. These values will be used for the request header.

Can you have a look on the changes?

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -2,9 +2,12 @@ import 'package:openfoodfacts/openfoodfacts.dart';
 
 /// request a product from the OpenFoodFacts database
 Future<Product> getProduct() async {
+  // a registered user login for https://world.openfoodfacts.org/ is required
+  User myUser = new User(userId: "max@off.com", password: "password");
+
   var barcode = "0048151623426";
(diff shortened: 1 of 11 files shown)


Grumpf86:
You are right, I will make the User optional in this case.

PrimaelQuemerais:
Great!
You can merge directly to master

Thanks for this new feature 🙂

PR #19, opened 2020-04-08
parse percent field of ingredients as double not as int

Grumpf86:
we want to parse the percent field of ingredients as double instead of int.
I've implemented this little change.

BUT the mapping of percent is not working at all.  I dont know why. Maybe just my IDE, json_serializer or flutter is broken. or I am missing something.  
test - "product ingredients with percent" is failing. 

Please do not merge until its working and approved.

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -2,12 +2,9 @@ import 'package:openfoodfacts/openfoodfacts.dart';
 
 /// request a product from the OpenFoodFacts database
 Future<Product> getProduct() async {
-  // a registered user login for https://world.openfoodfacts.org/ is required
-  User myUser = new User(userId: "max@off.com", password: "password");
-
   var barcode = "0048151623426";
   ProductResult result =
-      await OpenFoodAPIClient.getProduct(myUser, "barcode", User.LANGUAGE_DE);
+      await OpenFoodAPIClient.getProduct("barcode", User.LANGUAGE_DE);
 
   if (result.status == 1) {
     return result.product;
--- a/lib/openfoodfacts.dart
+++ b/lib/openfoodfacts.dart
@@ -83,7 +83,7 @@ class OpenFoodAPIClient {
   /// No parsing of ingredients.
   /// No adjustment by language.
   static Future<ProductResult> getProductRaw(
-    User user, String barcode, String lang) async {
+      String barcode, String lang, {User user}) async {
     if (barcode == null || barcode.isEmpty) {
       return new ProductResult();
     }
@@ -93,7 +93,7 @@ class OpenFoodAPIClient {
         host: _getHostByLanguage(lang),
         path: 'api/v0/product/' + barcode + '.json');
 
-    String response = await HttpHelper().doGetRequest(productUri, user);
+    String response = await HttpHelper().doGetRequest(productUri, user: user);
     var result = ProductResult.fromJson(json.decode(response));
     return result;
   }
@@ -102,8 +102,8 @@ class OpenFoodAPIClient {
   /// The ProductResult does not contain a product, if the product is not available.
   /// ingredients, images and product name will be prepared for the given language.
   static Future<ProductResult> getProduct(
-      User user, String barcode, String lang) async {
-    ProductResult result = await getProductRaw(user, barcode, lang);
+      String barcode, String lang, {User user}) async {
+    ProductResult result = await getProductRaw(barcode, lang, user: user);
 
     if (result.product != null) {
       ProductHelper.prepareProductName(result.product, lang);
@@ -149,7 +149,7 @@ class OpenFoodAPIClient {
 
     print("URI: " + searchUri.toString());
 
-    String response = await HttpHelper().doGetRequest(searchUri, user);
+    String response = await HttpHelper().doGetRequest(searchUri, user:user);
     var result = SearchResult.fromJson(json.decode(response));
 
     for (Product product in result.products) {
--- a/lib/utils/HttpHelper.dart
+++ b/lib/utils/HttpHelper.dart
@@ -24,7 +24,7 @@ class HttpHelper {
   /// Send a http get request to the specified uri.
   /// The data of the request (if any) has to be provided as parameter within the uri.
   /// The result of the request will be returned as string.
-  Future<String> doGetRequest(Uri uri, User user) async {
+  Future<String> doGetRequest(Uri uri, {User user}) async {
     http.Response response =
         await http.get(uri.toString(), headers: _buildHeaders(user));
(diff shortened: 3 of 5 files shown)


tzehe:
I introduced the following changes and all tests are succeeding besides the api_test_addProductImage.dart. 

Ingredient.dart
`  @JsonKey(includeIfNull: false, fromJson: JsonObject.parseDouble)
  double percent;`

when I run the api_test_addProductImage.dart I get the following errors: 

``` 

../../flutter/packages/flutter_test/lib/src/accessibility.dart:8:8: Error: Not found: 'dart:ui'
import 'dart:ui' as ui;
       ^
../../flutter/packages/flutter_test/lib/src/binding.dart:7:8: Error: Not found: 'dart:ui'
import 'dart:ui' as ui;
       ^
../../flutter/packages/flutter_test/lib/src/matchers.dart:8:8: Error: Not found: 'dart:ui'
import 'dart:ui' as ui;
```

I installed all the packages. Do you have any idea? Also which test case are you referring to? 

Best

Grumpf86:
Hey, thanks for your help :)
I added this test in api_test_getProduct.dart
`  test('product ingredients with percent', () async {
    String barcode = "4388860154948";
    ProductResult result = await OpenFoodAPIClient.getProduct(
        barcode, User.LANGUAGE_DE);

    assert(result != null);
    assert(result.product != null);
    print("number of ingredients: " +
        result.product.ingredients.length.toString());
    assert(result.product.ingredientsText != null);
    assert(result.product.ingredients != null);
    assert(result.product.ingredients.length > 0);
    result.product.ingredients.forEach((Ingredient i) => print(i.toJson()));
    assert(result.product.ingredients.any((Ingredient i) => i.percent != null));
  });`

Can you check if it succeed? 
I will have a look on those tests in api_test_addProductImage.dart

tzehe:
Ok will check and thanks a lot!

PrimaelQuemerais:
> I introduced the following changes and all tests are succeeding besides the api_test_addProductImage.dart.
> 
> Ingredient.dart
> ` @JsonKey(includeIfNull: false, fromJson: JsonObject.parseDouble) double percent;`
> 
> when I run the api_test_addProductImage.dart I get the following errors:
> 
> ```
> 
> ../../flutter/packages/flutter_test/lib/src/accessibility.dart:8:8: Error: Not found: 'dart:ui'
> import 'dart:ui' as ui;
>        ^
> ../../flutter/packages/flutter_test/lib/src/binding.dart:7:8: Error: Not found: 'dart:ui'
> import 'dart:ui' as ui;
>        ^
> ../../flutter/packages/flutter_test/lib/src/matchers.dart:8:8: Error: Not found: 'dart:ui'
> import 'dart:ui' as ui;
> ```
> 
> I installed all the packages. Do you have any idea? Also which test case are you referring to?
> 
> Best

Hi tzehe,
You may be running the test using Dart Test instead of Flutter Test, I ran into this issue once.
Hope that helps

Grumpf86:
thank God, I found the problem. there is a post processing step for all ingredients, which has only handled ingredient text and bold attribute, but not the rank and percentage value.
now ingredients are parsed as double and will be written into the resulting product object, while calling getProduct().

the product upload tests are not designed to run multiple times ... and will fail after one run, cause the image then already exists.

tzehe:
hahah found the same spot and was working on a fix but awesome that you already pushed yours. I found also another issue with product names resulting in an empty string when you select User.Language_Undefined. If there is for example a german name it will default to the english name which is empty therefore the name will be empty. will open another issue for that.

tzehe:
I found another issue. https://world.openfoodfacts.org/api/v0/product/4029764001807.json 

säurungsmittel zitronensäure should be percentage = 0
Mate in 100ml Getränk -> id: id: "de:Mate in 100 ml Getränk" should be 0.4

I guess the order of the ingredients is not the same because of the matching logic so iterating over the product.ingredients list to retrieve the percentage will lead to errors. 

Btw also the name of the ingredient is wrong. 

```
{rank: 1, text: Wasser, bold: false}
{rank: 2, text: Glukose-Fructosesirup, bold: false}
{rank: 3, text: Zucker, bold: false}
{rank: 4, text: Getränkegrundstoff, bold: false}
{rank: 5, text: Mate-Tee-Extrakt mit mind, bold: false}
{rank: 6, text: g Mate in 100 ml Getränk, bold: false} // <-- weird formatting should be: "Mate in 100 ml Getränk"
{rank: 7, text: Säurungsmittel Zitronensäure, percent: 0.4, bold: false}
{rank: 8, text: karamellisierter Zuckersirup, bold: false}
{rank: 9, text: Koffein, bold: false}
{rank: 10, text: natürliches Aroma, bold: false}
{rank: 11, text: Kohlensäure, bold: false}
```
I guess this is not an issue if most of the ingredients no percentage field at all or the first one as a percentage field. But if it's in the middle there will be issues.

I re added your testcase and updated the id.

```
  test('product ingredients with percent', () async {
    String barcode = "4029764001807";
    ProductResult result =
        await OpenFoodAPIClient.getProduct(barcode, User.LANGUAGE_DE);
    assert(result != null);
    assert(result.product != null);
    print("number of ingredients: " +
        result.product.ingredients.length.toString());
    assert(result.product.ingredientsText != null);
    assert(result.product.ingredients != null);
    assert(result.product.ingredients.length > 0);
    result.product.ingredients.forEach((Ingredient i) => print(i.toJson()));
    assert(result.product.ingredients.any((Ingredient i) => i.percent != null));

    // https://world.openfoodfacts.org/api/v0/product/4029764001807.json
    expect(result.product.ingredients[6].percent, 0); // säurungsmittel zitronensäure should be 0
    expect(result.product.ingredients[5].percent, 0.4); // Mate in 100ml Getränk -> id: id: "de:Mate in 100 ml Getränk" should be 0.4
  });
```

Grumpf86:
ohh, I feared this case :) than its not a easy fix, as I supposed.

so  lets do the post processing the "right" way. (the best way I actually know)

1. use the **id** field of the ingredient.
`"id":"en:rice-noodles"`
2. download the ingredient name based on the user language, (may be we can query only one ingredient name instead of all)
https://world.openfoodfacts.org/ingredients.json (125 MB)
https://fr.openfoodfacts.org/ingredients.json
https://de.openfoodfacts.org/ingredients.json
3. override the ingredient name for the given language

Grumpf86:
Solved this issue with commit f9fa8839059a541eb4fc3981046e1794c0e018c6
https://github.com/openfoodfacts/openfoodfacts-dart/commit/f9fa8839059a541eb4fc3981046e1794c0e018c6

PR #21, opened 2020-04-09
product field "images" added. individual method for parsing from json

Grumpf86:
beside the field **selected_images**, the more generel field  **images** contains size information for each image.

the content of **images** will be parsed to the existing **ProductImage** class. 
the tests for getProduct() and getProductRaw() have been extended.


PrimaelQuemerais:
Sounds like a better implementation than the previous one to me :)
Can we consider removing ImageList and put the List<ProductImage> directly in the Product class ?
Is there a use case for an ImageList without Product ?

Grumpf86:
I removed ImageList, as you suggested. :+1: 
Soon I will convert the image string constants **size** and **field** to enum types.

PR #23, opened 2020-05-03
Fix searching errors

fibbers:
Hi,

First of all: I really like this project 😄 

I was trying to perform a search using `OpenFoodAPIClient.searchProducts` and saw that I got several parsing errors. This PR aims to solve that.

I tried the modifications locally and it worked as expected:
- the API gives back JSON
- there's no parsing error anymore when parsing `{ ..., "count": "49", ... }`

### DIFF
--- a/lib/model/SearchResult.dart
+++ b/lib/model/SearchResult.dart
@@ -12,10 +12,10 @@ class SearchResult extends JsonObject {
   @JsonKey(name: "page_size", nullable: false, fromJson: JsonObject.parseInt)
   final int pageSize;
 
-  @JsonKey(name: "count", nullable: false)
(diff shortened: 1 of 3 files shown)


Grumpf86:
Thanks for your contribution @fibbers  :)
Is it necessary to add the "json output format" to other queries, too?

fibbers:
Cool, thanks.
I haven't used that many query-types from the API yet: only search and getting details for one product, and both now retrieve json. So I'm not sure about other queries, but if I come across it, I'll be sure to create a PR for it if necessary :)

_NB: I tried [searching the wiki](https://wiki.openfoodfacts.org/index.php?search=output+format) where I only found examples for `search.pl` which is already addressed now._

PR #24, opened 2020-05-18
removed dart:ui and flutter_foundation dependencies

Grumpf:
Hey :)
there is a constraint of flutter, not allowing UI or widget specific code in integration and unit tests.
When you want to do a server or api call within a test, the openfoodfacts dart api must not contain packages referencing dart:ui like flutter_foundation, flutter_material and so on.

I removed those dependencies. 
Removing flutter_test and replacing it by (dart) test, did not work well (for intellij)
So I keep flutter_test

https://stackoverflow.com/questions/52462646/how-to-solve-not-found-dartui-error-while-running-integration-tests-on-flutt


Grumpf:
@PrimaelQuemerais can you have a short look on this?

Grumpf86:
Ups, I've created this pull - request with an github old account :P

PrimaelQuemerais:
Sounds good to me, any idea why those imports were here in the first place ?

Issue #35, opened 2020-07-11
Constructing Product from JSON fails


peterwvj:
I found an issue with construction of `Product` objects from JSON. The issue can be demonstrated using the following test:

```dart
import 'dart:convert';

import 'package:flutter_test/flutter_test.dart';
import 'package:openfoodfacts/model/Product.dart';
import 'package:openfoodfacts/model/ProductResult.dart';
import 'package:openfoodfacts/openfoodfacts.dart';
import 'package:openfoodfacts/utils/LanguageHelper.dart';
import 'package:openfoodfacts/utils/ProductFields.dart';
import 'package:openfoodfacts/utils/ProductQueryConfigurations.dart';

import 'test_constants.dart';

void main() {
  test('Construct Product from JSON - database example', () async {
    String barcode = '0030000010204';
    ProductQueryConfiguration configurations = ProductQueryConfiguration(
        barcode,
        language: OpenFoodFactsLanguage.ENGLISH,
        fields: [ProductField.ALL]);
    ProductResult result = await OpenFoodAPIClient.getProduct(configurations,
        user: TestConstants.TEST_USER);
    expect(result.status, 1);
    Product product = result.product;
    Map<String, dynamic> productMap = product.toJson();
    String json = jsonEncode(productMap);
    assert(json is String);
    Map<String, dynamic> productRestoredMap = jsonDecode(json);
    Product productRestored = Product.fromJson(productRestoredMap);
    expect(productRestored.productName, "Quaker Oats Old Fashioned Oatmeal 18 Ounce Paper Cannister");
  });
}
```
When you execute this test you get the following error:
```
Converting object to an encodable object failed: _LinkedHashMap len:5
dart:convert               jsonEncode
test/json_test.dart 25:19  main.<fn>
```
The problem is related to the `selectedImages` and `additives` fields in the `Product` class. If you assign these fields to `null` then the test passes:
```dart
...
Product product = result.product;
product.selectedImages = null;
product.additives = null;
...
```

PrimaelQuemerais:
This issue is fixed with version 0.3.6

Issue #34, opened 2020-07-10
JSON generation broken


peterwvj:
There is an issue with generation of JSON from `Product` objects. The issue seems to related to product images.

The issue can be demonstrated using the following test:

```dart
import 'dart:convert';

import 'package:flutter_test/flutter_test.dart';
import 'package:openfoodfacts/model/Product.dart';
import 'package:openfoodfacts/model/ProductResult.dart';
import 'package:openfoodfacts/openfoodfacts.dart';
import 'package:openfoodfacts/utils/LanguageHelper.dart';
import 'package:openfoodfacts/utils/ProductFields.dart';
import 'package:openfoodfacts/utils/ProductQueryConfigurations.dart';

import 'test_constants.dart';

void main() {
  test('Generate JSON - database example', () async {
    String barcode = '0030000010204';
    ProductQueryConfiguration configurations = ProductQueryConfiguration(
        barcode,
        language: OpenFoodFactsLanguage.ENGLISH,
        fields: [ProductField.ALL]);
    ProductResult result = await OpenFoodAPIClient.getProduct(configurations,
        user: TestConstants.TEST_USER);
    expect(result.status, 1);
    Product product = result.product;
    Map<String, dynamic> productMap = product.toJson();
    String json = jsonEncode(productMap);
    assert(json is String);
  });
}
```
This test prduces the following error:
```
Converting object to an encodable object failed: _LinkedHashMap len:5
dart:convert               jsonEncode
test/json_test.dart 25:19  main.<fn>
```
Hint: the problem seems to be the `selectedImages` defined in the `Product` class. In particular, if you remove the content of this field by setting it to `null` then the JSON is generated without errors:
```
....
Product product = result.product;
product.selectedImages = null;
....
```

PrimaelQuemerais:
This issue is fixed with version 0.3.6
I also added your test to the package

PR #36, opened 2020-07-12
Fixed Json product generation (additives, selectedImages)

PrimaelQuemerais:
Addresses issues #34 and #35


PrimaelQuemerais:
Rejected, adding a few corrections before merging

Issue #44, opened 2020-08-01
kcal / kj


adfaure:
Hello,
is there a reason why the kcal value is not present in the nutriments class?

I saw that one can convert from KJ to kcal.
Is it a better option do to the conversion, or to add this value into the nutriment class to access the value returned by OFF API?

Thank you,
Adrien

PrimaelQuemerais:
Hi Adrien,
I think the reason why the field is not present in the nutrition class is sometimes the API doesn't return the value in kcal, in the use-case I was working on at the time it made sens to only use the kj field.

Of course we can add the kcal field as it might be usefull for others. Do you want to make a PR ? Otherwise I'll add the field in the next version of the plugin

adfaure:
Hello, thank you for your reply.
Indeed, it makes sense.

I already did it on a clone, 
I'll do a merge request within the day.

PrimaelQuemerais:
Great! Thank you for helping, I'll merge you'r PR and update the plugin as soon as I can.

PrimaelQuemerais:
Merged

PR #45, opened 2020-08-02
Add kcal values in Nutriments

adfaure:
Hello,

I made the changes to include the `kcal` information into the Nutriment class.
It has been discussed in issue #44 

I am not familiar with OpenFoodFacts, so I am not sure if I used the good fields: `energy-kcal_100g` vs `energy-kcal`...

All tests passe, except for one:

```
00:02 +15 -1: ~/openfoodfacts-dart/test/api_getProductRaw_test.dart: OpenFoodAPIClient get raw products get product test 1 [E]                                                                                                                           
  Expected: <15>
    Actual: <18>
  
  package:test_api                                   expect
  package:flutter_test/src/widget_tester.dart 348:3  expect
  test/api_getProductRaw_test.dart 37:7              main.<fn>.<fn>
```

However, this one seems to be broken on master too.

Thank you,
Adrien

### DIFF
--- a/lib/model/Nutriments.dart
+++ b/lib/model/Nutriments.dart
@@ -43,6 +43,16 @@ class Nutriments extends JsonObject {
       fromJson: JsonObject.parseDouble)
   double energy;
   @JsonKey(
+      name: "energy-kcal",
+      includeIfNull: false,
+      fromJson: JsonObject.parseDouble)
+  double energyKcal;
+  @JsonKey(
+      name: "energy-kcal_100g",
+      includeIfNull: false,
+      fromJson: JsonObject.parseDouble)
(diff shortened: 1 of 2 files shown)


PrimaelQuemerais:
This test failing is normal, the data of products change constantly, we should we-write all our tests to be adaptative.
Looks great thanks! I'm merging

PrimaelQuemerais:
The package is now available on pub.dev with your changes (v0.3.10)
Do you want to be added in the contributors list ? :)

adfaure:
Thank you for merging.

For the moment I don't know if we can say that I am a *contributor*.
I only added what I needed...

So as you wish, thank you :)

Issue #48, opened 2020-08-29
OpenFoodAPIClient.saveProduct fails to transfer nutriments


peterwvj:
I found what seems to be a critical bug in `OpenFoodAPIClient.saveProduct` that prevents nutriments from being saved.

### How to produce

Specify working user credentials and run the following code to add/edit a (real) product to the database:

```dart
import 'package:openfoodfacts/model/Nutriments.dart';
import 'package:openfoodfacts/model/Product.dart';
import 'package:openfoodfacts/model/Status.dart';
import 'package:openfoodfacts/model/User.dart';
import 'package:openfoodfacts/openfoodfacts.dart';

void main() async {
  // TODO: specify user credentials
  User myUser = User(userId: 'user', password: 'secret-password');

  var nutriments = Nutriments();
  nutriments.energy = 365;
  nutriments.carbohydrates = 12;
  nutriments.proteins = 6;
  nutriments.fat = 0.1;

  var newProduct = Product(
      barcode: "7340011364184",
      productName: "Chili beans",
      nutrimentDataPer: "100g",
      nutriments: nutriments);

  Status result = await OpenFoodAPIClient.saveProduct(myUser, newProduct);
  print(result.statusVerbose);
}
```
Next, look up the product by the barcode by entering https://world.openfoodfacts.org/api/v0/product/7340011364184.json into your browser.

### Expected result

I'd expect the values of the `code`,  `product_name`, `nutrition_data_per` and `nutriments` fields in `7340011364184.json` to match those in the code snippet above.

### Actual result

You'll see that the `code`, `product_name` and `nutrition_data_per` fields have the correct values but `nutriments` is empty, i.e. `"nutriments":{}`. So for some reason nutriments are not saved.

### Other information

Produced using version 0.3.10 of the API.

peterwvj:
The body of the request seems to contain the nutriments information:

```
[...]

PR #49, opened 2020-09-09
Fix saveProduct

peterwvj:
Work on #48

### DIFF
--- a/lib/interface/JsonObject.dart
+++ b/lib/interface/JsonObject.dart
@@ -1,11 +1,28 @@
+import 'package:json_annotation/json_annotation.dart';
+import 'package:openfoodfacts/utils/MapUtil.dart';
+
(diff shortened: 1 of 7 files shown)


peterwvj:
I'm having second thoughts about this approach (the PR is a draft to share ideas). Mostly because not all fields in the `Product` class have a counterpart in the write API. For example, the write API has `nutriment_proteins_unit` but this field does not exist in the `Product` or `Nutriments` classes as far as I'm aware. With the write API you specify nutriments and  units separately whereas with the read API this is normalised somehow. As an example, the `Product` class has a `proteins_100g` field. So I'm not sure simply remapping fields using apiFields will work for all fields. So rather than having a `saveProduct(Product product)` function (which we currently have) maybe what we need is a `saveProduct(ProductChange change)` function

PR #51, opened 2020-09-15
Fix OpenFoodAPIClient.addProductImage

peterwvj:
Update SDK to implement [API/Write/Photos](https://wiki.openfoodfacts.org/index.php?title=API/Write/Photos&mobileaction=toggle_view_desktop)

Also update tests to pass.


PrimaelQuemerais (approved):
Looks great

PR #52, opened 2020-09-27
Fix parsing of imageId

peterwvj:
Apparently the `imageId` is sometimes returned as a `String` so it's safer to parse it rather than casting it.


PrimaelQuemerais:
Thank you

PR #47, opened 2020-08-18
Extended micronutrients support

indigothm:
I needed extended support for micronutrients. So additional units and nutrients were added and tested.

PR #56, opened 2020-11-20
Some tests fixed and product attribute added

Grumpf86:
test fixed:
- api_getProduct_test
- api_searchProducts_test

field added for Product.dart 
- last modified_t


Grumpf86:
I will release these changes and the changes of @indigothm . #47 in a new package version 0.3.13

PR #57, opened 2020-11-25
added field ecoscore_grade to product

FredJul:
Hi,

I have added the ecoscore_grade field to Product, which is currently only available when we query with ecoscore_alpha.
Adding this will allows apps (like the smooth app) to prepare the arrival of the ecoscore (which I'm really excited about).

If this is accepted, I will try to add ecoscore_data handling later.

### DIFF
--- a/lib/model/Product.dart
+++ b/lib/model/Product.dart
@@ -153,6 +153,9 @@ class Product extends JsonObject {
     toJson: JsonHelper.dateToTimestamp)
   DateTime lastModified;
 
+  @JsonKey(name: 'ecoscore_grade', includeIfNull: false)
+  String ecoscoreGrade;
+
   Product(
       {this.barcode,
       this.productName,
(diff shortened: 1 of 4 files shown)


teolemon:
Thanks for the PR @FredJul 👍 
Note that we don't want the alpha to be available in the DART package before the synchronized launch. cc @stephanegigandet

FredJul:
Ok, my bad...
But can I ask what are the rational behind this? Does that means clients cannot really prepare until it goes out of alpha?

stephanegigandet (approved):
Looks good, thank you!

stephanegigandet:
"Does that means clients cannot really prepare until it goes out of alpha?"

@FredJul : it's fine, we can merge as-is, the Eco-Score will soon go out of alpha anyway.

FredJul:
I didn't expect the ecoscore to come that soon so that's great news. Expect more PRs on dart ecoscore support in a near future.

PR #58, opened 2020-12-01
product name hotfix

Grumpf86:
- dont send a product name, if its null. (json includeIfNull=false)
- added test case for a product with a name in different languages


Grumpf86:
Its a very small change, but it prevent the API from sending a product name like "null".
And open up the possibility to send a product name only in a specific language. e.g. setting "product_name_fr" and leaving "product_name" empty.

PR #61, opened 2020-12-27
More flexible AttributeGroup's

monsieurtanuki:
Deleted files:
* `AttributeGroups`: replaced by new files `Attribute.dart` and `AttributeGroup.dart`
* `Product.g.dart`: replaced by more readable code (?) in `Product.dart`

New files:
* `Attribute.dart`: inspired by the code in `AttributeGroups`, refactored and added 2 fields
* `AttributeGroup.dart`: inspired by the code in `AttributeGroups`, refactored and added 1 field

Impacted files:
* `api_getProduct_test.dart`: refactoring around `AttributeGroups`; auto reformatting by dart
* `JsonHelper.dart`: minor refactoring; auto reformatting by dart
* `JsonObject.dart`: added cleaning method `removeNullEntries`
* `Product.dart`: new version of `AttributeGroups`; refactored maybe a bit heavily because I could not read those `JsonKey`'s that do not look nice, especially after an automatic dart reformatting, and are not especially relevant given the simplicity of our json's.


lib/model/Attribute.dart
```diff
@@ -0,0 +1,73 @@
...
+  Map<String, dynamic> toJson() => JsonObject.removeNullEntries({
+        _JSON_TAG_ID: id,
+        _JSON_TAG_NAME: name,
+        _JSON_TAG_TITLE: title,
+        _JSON_TAG_ICON_URL: iconUrl,
+        _JSON_TAG_DEFAULT: defaultF,
+        _JSON_TAG_SETTING_NOTE: settingNote,
+        _JSON_TAG_SETTING_NAME: settingName,
+        _JSON_TAG_DESCRIPTION: description,
+        _JSON_TAG_DESCRIPTION_SHORT: shortDescription,
```
stephanegigandet, line 42: I know it was already like this before, but can we change shortDescription to descriptionShort, so that the field names matches the ones we have in the JSON API? The less surprises / differences, the better. :)

peterwvj:
I noticed that the Product.g.dart class was removed and in consequence Product.dart has been updated with manual implementations of `toJson` and `fromJson`, which concerns me quite a lot. The SDK generally uses `json_serializable` for converting between Dart objects and JSON to avoid manual implementation (the *.g.dart files are generated). I'm not saying that the PR is wrong but I think that the manual implementations are costly to maintain given the size of the Product class. I really think this should be avoided unless there is a very good reason for doing this (the PR doesn't mention one). Please let me know if I'm missing something. If the change I'm concerned about isn't related to the topic of the PR as such, i.e. "More flexible AttributeGroup" then perhaps it should not be part this PR.

lib/model/Attribute.dart
peterwvj: `json_serializable` has built-in support for removing field that are `null`, e.g. `@JsonKey(name: 'product_name', includeIfNull: false)`. Like I commented on the PR, I'm worried about the manual implementations of the JSON conversion functions.

peterwvj:
Some additions comments/questions that came up:

 - I should say that my knowledge about attribute groups is fairly limited so could you please explain the contribution of the PR. Does the PR fix a problem/bug in the SDK? Or is the contribution something else? I think explaining this in broad terms will help the reviewers.
 - Does this PR break compatibility? That is, if the PR is merged and a new version of the SDK is released, can I upgrade my app to use the latest SDK without any concerns?

peterwvj (commented):
The concerns I have are added as comments on the PR https://github.com/openfoodfacts/openfoodfacts-dart/pull/61

monsieurtanuki:
@peterwvj Here's the story (I'll try to keep it short):
* I started contributing to [smooth-app](https://github.com/openfoodfacts/smooth-app) about two weeks ago
* On one of the issues I was working on (possibly https://github.com/openfoodfacts/smooth-app/issues/30) I needed to work on the product's attribute groups
* And I was limited in that by the available version of `Product` (`0.3.13+2`) regarding the field `attributeGroups`
* I needed to have a look at the code of `Product`, and I was a bit surprised, as I don't use `json_serializable` myself. I found the code hard to read (I prefer to see clearly all the fields, there are 2 files to read), and given the simplicity of the json encoding/decoding in flutter I was not convinced it was relevant to use those annotations and code generation. But in first approach that's just a matter of taste.
* I didn't know if someone was actively responsible for the code - in the doubt, I changed it my way

The most important change I need in `Product` is `final List<AttributeGroup> attributeGroups` instead of `AttributeGroups attributeGroups`. And @stephanegigandet asked if we could rename `shortDescription` to `descriptionShort` in `Attribute`. Both changes may impact the parts of your code that rely on it.

Beyond that, I'm not a big fan of `json_serializable` and I prefer `final` fields (at least for `String barcode`, no?), but it's OK with me if the `json_serializable` style is kept for `Product`).

What's the next step?

stephanegigandet:
Hi @peterwvj ,
Here is a wiki page that describes the product attributes: https://wiki.openfoodfacts.org/Product_Attributes
It is something we introduced a few months ago to allow the creation of a personalized search feature on the OFF web site and on mobile apps, and to simplify how apps can access product features such as Nutriscore, allergens, NOVA, vegeterian diets etc. by always returning the same structure regardless of the product feature.

> The most important change I need in `Product` is `final List<AttributeGroup> attributeGroups` instead of `AttributeGroups attributeGroups`. And @stephanegigandet asked if we could rename `shortDescription` to `descriptionShort` in `Attribute`. Both changes may impact the parts of your code that rely on it.

I do not know of anyone using the Attribute Groups through Flutter yet.

lib/model/Product.dart
```diff
@@ -1,188 +1,274 @@
...
-  @JsonKey(name: 'nutrition_grade_fr', includeIfNull: false)
-  String nutriscore;
-
-  @JsonKey(includeIfNull: false)
-  String categories;
-
-  @JsonKey(name: 'categories_tags', includeIfNull: false)
-  List<String> categoriesTags;
-  @JsonKey(name: 'categories_tags_translated', includeIfNull: false)
+  final String barcode;
```
peterwvj, line 167: After spending some more time testing this PR with my app I'm getting several compilation errors because fields have been changed to final, which breaks compatibility. I rely a lot on this in my app tests, i.e. instantiating a `Product`, setting some fields, performing some tests, then updating some additional fields for the same object and running some more tests. I think the `final` fields should be avoided.

lib/model/Attribute.dart
```diff
@@ -0,0 +1,73 @@
...
+  final String title;
+  final String iconUrl;
+  final String defaultF;
+  final String settingNote;
+  final String settingName;
+  final String description;
+  final String shortDescription;
+  final double match;
+  final String status;
+
```
peterwvj, line 70: `final` fields. See my other comment regarding this https://github.com/openfoodfacts/openfoodfacts-dart/pull/61/files#r550243068

lib/model/AttributeGroup.dart
```diff
@@ -0,0 +1,52 @@
...
+  static const String _JSON_TAG_ID = 'id';
+  static const String _JSON_TAG_NAME = 'name';
+  static const String _JSON_TAG_WARNING = 'warning';
+  static const String _JSON_TAG_ATTRIBUTES = 'attributes';
+
+  final String id;
+  final String name;
+  final String warning;
+  final List<Attribute> attributes;
+
```
peterwvj, line 38: `final` fields. See my other comment regarding this https://github.com/openfoodfacts/openfoodfacts-dart/pull/61/files#r550243068

peterwvj:
Hi @monsieurtanuki and thanks for explaining.

As stated in the comments above, I did a bit of testing using your PR and found minor changes that break compatibility. Fortunately, the changes seem easy to revert.

Regarding the json_serializable changes. Based on your comment I don't see a good reason for manually implementing JSON conversion. As far as I can tell it doesn't relate to the topic of the PR as such so I'd prefer not to include these changes in this PR. If we agree to include these changes at a later point then they should be submitted as a seperate PR to keep PR 61 less busy (and simpler to review).

monsieurtanuki:
@peterwvj My refactoring of `Product` was disproportionate, I'll give you that.
I'll send an alternate and more reasonable PR within a week.

PR #63, opened 2020-12-29
added field ecoscore_score and ecoscore_data to product

FredJul:
Hi,

I have added the ecoscore_score and ecoscore_data field to Product, which are currently only available when we query with ecoscore_alpha.

All data sub-fields are not handled yet.

### DIFF
--- a/lib/model/EcoscoreAdjustments.dart
+++ b/lib/model/EcoscoreAdjustments.dart
@@ -1,6 +1,6 @@
 import 'package:json_annotation/json_annotation.dart';
 import '../interface/JsonObject.dart';
-import 'IngredientsOrigins.dart';
+import 'OriginsOfIngredients.dart';
 import 'Packaging.dart';
 
 part 'EcoscoreAdjustments.g.dart';
@@ -10,9 +10,9 @@ class EcoscoreAdjustments extends JsonObject {
   @JsonKey(includeIfNull: false)
   Packaging packaging;
   @JsonKey(name: "origins_of_ingredients", includeIfNull: false)
-  IngredientsOrigins ingredientsOrigins;
+  OriginsOfIngredients originsOfIngredients;
 
-  EcoscoreAdjustments({this.packaging, this.ingredientsOrigins});
+  EcoscoreAdjustments({this.packaging, this.originsOfIngredients});
 
   factory EcoscoreAdjustments.fromJson(Map<String, dynamic> json) =>
       _$EcoscoreAdjustmentsFromJson(json);
--- a/lib/model/EcoscoreAdjustments.dart
+++ b/lib/model/EcoscoreAdjustments.dart
@@ -1,0 +1,22 @@
+import 'package:json_annotation/json_annotation.dart';
+import '../interface/JsonObject.dart';
+import 'IngredientsOrigins.dart';
+import 'Packaging.dart';
+
+part 'EcoscoreAdjustments.g.dart';
+
+@JsonSerializable()
+class EcoscoreAdjustments extends JsonObject {
+  @JsonKey(includeIfNull: false)
+  Packaging packaging;
+  @JsonKey(name: "origins_of_ingredients", includeIfNull: false)
+  IngredientsOrigins ingredientsOrigins;
+
+  EcoscoreAdjustments({this.packaging, this.ingredientsOrigins});
+
+  factory EcoscoreAdjustments.fromJson(Map<String, dynamic> json) =>
+      _$EcoscoreAdjustmentsFromJson(json);
+
+  @override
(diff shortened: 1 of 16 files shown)


lib/model/EcoscoreAdjustments.dart
```diff
@@ -0,0 +1,22 @@
...
+part 'EcoscoreAdjustments.g.dart';
+
+@JsonSerializable()
+class EcoscoreAdjustments extends JsonObject {
+  @JsonKey(includeIfNull: false)
+  Packaging packaging;
+  @JsonKey(name: "origins_of_ingredients", includeIfNull: false)
+  IngredientsOrigins ingredientsOrigins;
+
+  EcoscoreAdjustments({this.packaging, this.ingredientsOrigins});
```
stephanegigandet, line 15: There are 2 more adjustments, for the production system (e.g. organic labels) and for the threatened species (e.g. palm oil).
example products:

https://world.openfoodfacts.org/api/v0/product/3478820023184?fields=product_name,ecoscore_data,ecoscore_alpha (organic)

https://world.openfoodfacts.org/api/v0/product/3017620422003?fields=product_name,ecoscore_data,ecoscore_alpha (palm oil)

lib/model/EcoscoreAdjustments.dart
```diff
@@ -0,0 +1,22 @@
...
+import 'Packaging.dart';
+
+part 'EcoscoreAdjustments.g.dart';
+
+@JsonSerializable()
+class EcoscoreAdjustments extends JsonObject {
+  @JsonKey(includeIfNull: false)
+  Packaging packaging;
+  @JsonKey(name: "origins_of_ingredients", includeIfNull: false)
+  IngredientsOrigins ingredientsOrigins;
```
stephanegigandet, line 13: Can we keep the field names the same as in the API? e.g. OriginsOfIngredients instead of IngredientsOrigins
It's easier if the field names are the same as other places on OFF.

FredJul: ok, no problem

lib/model/EcoscoreAdjustments.dart
FredJul: You're right, but I believe there is plenty of other missing fields as well (packagings, agribalyse stuff, ...).

I did not put everything for several reasons:
- it's a first step, I can continue to improve it later, especially is there is a need for these fields. Adding everything in one single PR may be a bit harder to code review.
- I'm not sure you want to be exhaustive in this dart lib or if you want to simply match flutter clients needs.
- I didn't find a documentation of all possible fields and wanted to ask you for that a bit later. Maybe I simply didn't look at the right place.
- I'm not sure how to handle the "en:XXX" strings (like the ones for labels and threatened species). Is is always english or can we sometimes have a map with several language instead?

So, do you still think I should add these two fields? If yes, I can do it.

stephanegigandet: The Eco-Score is very new, there is no documentation for it yet (one reason is that the Eco-Score has not launched yet, so we have been asked not to make it publicly visible. That will change soon though. :) )

en:something fields are identifiers for tags, we have translations for them in different languages. But there isn't an API to retrieve the translations yet. We'll need to make one.

stephanegigandet: It's fine to have a PR with partial support.

FredJul: Then if you agree I suggest to do that in another PR. Also, I don't fully understand how we can have only one production_system's label and only one threatened_species's ingredient, and I wonder if it shouldn't be an array. So here is another question: can we consider the current API to be stable?

stephanegigandet: Another PR is fine, thanks.

For the production system label, for the Eco-Score, only the most beneficial label is kept. e.g. if a product is both organic and "Label Rouge", then the bonus for organic is applied, not two bonuses.

For threatened species, at this point only palm oil is considered. There are talks to add some fishes. It might be the same as the production system: we might consider only the worse ingredient.

So, no, the current API is not 100% stable at this point.

stephanegigandet (approved):
Thanks!

PrimaelQuemerais:
Hello, I took the time to pull and check the PR today.
Code is excellent, all tests pass, all I did was correct errors in prior test (unrelated to this PR).
I've merged your pull request and released the version 0.3.14 of the plugin on pub.dev

Thank you very much for the contribution @FredJul 

PS: I've added you to the credits of the plugin let me know if you want it in another format or removed :wink:

FredJul:
Thank you @PrimaelQuemerais for the credits. FYI my current goal is to add the ecoscore to the smooth app, so I'll certainly do PRs for that as well.

PR #64, opened 2021-01-02
fix 'match' attribute parsing

FredJul:
When the match attribute is null, we can have a parsing exception. Using JsonObject.parseDouble allows to handle all cases.


FredJul:
I just noticed this is also fixed in #61, so I'm closing that PR.

PR #71, opened 2021-01-23
fix: product JSON generation (EcoscoreData related)

### LINKED ISSUE #70: Product JSON generation broken due to problem with EcoscoreData
The Product JSON generation is broken. In particular, the `ecoscore_data` field is not converted into JSON when the `Product.toJson` method is called.

The problem can be demonstrated using the following test, which is expected to pass but currently doesn't with version 0.3.14+1 of the openfoodfacts-dart SDK.

```dart
import 'package:openfoodfacts/model/EcoscoreData.dart';
import 'package:openfoodfacts/model/Product.dart';

void main() {
  test('Product/EcoscoreData Json bug', () {
    final product = Product();
    product.productName = 'TestProduct';
    final ecoscoreData = EcoscoreData();
    ecoscoreData.grade = 'x';
    product.ecoscoreData = ecoscoreData;
    final productJson = product.toJson();
    assert(!(productJson['ecoscore_data'] is EcoscoreData));
  });
}
```
I expect this to be related to #63. @FredJul do you know what is up?

peterwvj:
Closes #70

### DIFF
--- a/lib/model/EcoscoreData.dart
+++ b/lib/model/EcoscoreData.dart
@@ -27,4 +27,7 @@ class EcoscoreData extends JsonObject {
 
   @override
   Map<String, dynamic> toJson() => _$EcoscoreDataToJson(this);
+
+  static Map<String, dynamic> toJsonHelper(EcoscoreData d) =>
+      d != null ? d.toJson() : {};
 }
--- a/lib/model/EcoscoreData.dart
+++ b/lib/model/EcoscoreData.dart
@@ -29,5 +29,5 @@ class EcoscoreData extends JsonObject {
   Map<String, dynamic> toJson() => _$EcoscoreDataToJson(this);
 
   static Map<String, dynamic> toJsonHelper(EcoscoreData d) =>
-      d != null ? d.toJson() : {};
+      d != null ? d.toJson() : null;
 }
--- a/test/json_product_generation_test.dart
+++ b/test/json_product_generation_test.dart
@@ -1,6 +1,7 @@
 import 'dart:convert';
 
 import 'package:flutter_test/flutter_test.dart';
+import 'package:openfoodfacts/model/EcoscoreData.dart';
 import 'package:openfoodfacts/model/Product.dart';
 import 'package:openfoodfacts/model/ProductResult.dart';
 import 'package:openfoodfacts/openfoodfacts.dart';
@@ -22,8 +23,17 @@ void main() {
     expect(result.status, 1);
     Product product = result.product;
     Map<String, dynamic> productMap = product.toJson();
-    print(JsonHelper.selectedImagesToJson(product.selectedImages));
     String json = jsonEncode(productMap);
     assert(json is String);
+  });
+
+  test('EcoscoreData JSON generation', () {
+    final product = Product();
+    product.productName = 'TestProduct';
+    final ecoscoreData = EcoscoreData();
+    ecoscoreData.grade = 'x';
+    product.ecoscoreData = ecoscoreData;
+    final productJson = product.toJson();
+    assert(!(productJson['ecoscore_data'] is EcoscoreData));
   });
 }
--- a/test/json_product_generation_test.dart
+++ b/test/json_product_generation_test.dart
@@ -28,10 +28,8 @@ void main() {
   });
 
   test('EcoscoreData JSON generation', () {
-    final product = Product();
-    product.productName = 'TestProduct';
-    final ecoscoreData = EcoscoreData();
-    ecoscoreData.grade = 'x';
+    final product = Product(productName: 'TestProduct');
+    final ecoscoreData = EcoscoreData(grade: 'x');
     product.ecoscoreData = ecoscoreData;
     var productJson = product.toJson();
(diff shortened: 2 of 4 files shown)


lib/model/EcoscoreData.dart
```diff
@@ -27,4 +27,7 @@ class EcoscoreData extends JsonObject {
 
   @override
   Map<String, dynamic> toJson() => _$EcoscoreDataToJson(this);
+
+  static Map<String, dynamic> toJsonHelper(EcoscoreData d) =>
+      d != null ? d.toJson() : {};
```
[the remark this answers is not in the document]
peterwvj: Given that the `ecoscore_data` field is only translated to JSON when it is different from `null` them "empty map" case should never occur:

```dart
writeNotNull(
      'ecoscore_data', EcoscoreData.toJsonHelper(instance.ecoscoreData));
```
... where `writeNotNull` is defined as follows:
```dart
writeNotNull(
      'ecoscore_data', EcoscoreData.toJsonHelper(instance.ecoscoreData));
```
So even though we change it to return `null` rather than the empty map it would not make a difference.

So I think it is fine to keep it as it is.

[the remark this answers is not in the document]
peterwvj: As a site note, I deleted my previous comment because I misunderstood but you were saying to begin with :) Sorry for the confusion. Anyway, I'll mark this as resolved. Let me know if you disagree.

lib/model/EcoscoreData.dart
```diff
@@ -27,4 +27,7 @@ class EcoscoreData extends JsonObject {
 
   @override
   Map<String, dynamic> toJson() => _$EcoscoreDataToJson(this);
+
+  static Map<String, dynamic> toJsonHelper(EcoscoreData d) =>
```
FredJul, line 31: I see your problem and agree that it does not look normal to have a EcoscoreData instance in a json serialization output.

However, I'm a bit puzzled because: 
1/ the official sample seems to behave the same (line 39 of https://github.com/google/json_serializable.dart/blob/master/example/lib/example.g.dart we can see that it directly put the 'instance.orders' variable, without using toJson on it)

2/ I have an app which has the same "issue", but when I give the object to retrofit it will simply work. At the end, and I'm not sure why, the server receives a correct json.

So I wonder if it has not been designed this way, even if I find that quite surprising as well... Does someone has an answer?

test/json_product_generation_test.dart
```diff
@@ -22,8 +23,20 @@ void main() {
...
-    print(JsonHelper.selectedImagesToJson(product.selectedImages));
     String json = jsonEncode(productMap);
     assert(json is String);
   });
-}
\ No newline at end of file
+
+  test('EcoscoreData JSON generation', () {
+    final product = Product();
+    product.productName = 'TestProduct';
```
FredJul, line 32: For the readability I would suggest to use the cascade operator (see https://dart.dev/guides/language/language-tour#cascade-notation- )

lib/model/EcoscoreData.dart
FredJul: Ok, I just found an explanation there: https://github.com/google/json_serializable.dart/issues/702

We need to use 'explicitToJson: true' on all @JsonSerializable classes (including Product) to "fix" this issue.
But in explicitToJson documentation they say it's not really needed to do so since dart:convert will handle it.

So either we do nothing and accept this behavior, either we use explicitToJson everywhere.

peterwvj: Hi @FredJul , I am aware of the `explicitToJson` flag I intentionally did not add it. Currently, the SDK simply handles JSON generation differently (using the `toJson` helper function) and I wanted to fix the bug in a way that is consistent with respect to how JSON generation is handled for other classes/fields in the SDK.

Note, I'm not saying the way the SDK handles JSON generation is "right". In fact, using the `explicitToJson` could be a better way going forward. If we agree, at a later point, to use `explicitToJson` then we can add that using a separate pull request. That would require addressing this problem for all JSON classes in the SDK, which is outside the scope of this particular, PR.

For what is worth this PR is consistent with the way JSON generation is handled for other classes and fields in this SDK.

test/json_product_generation_test.dart
peterwvj: I cleaned up the test. I was able to simply pass the field values via constructors.

lib/model/EcoscoreData.dart
FredJul: Well, this PR seems to be consistent with the rest of the app, but I'm not fan of the added complexity (compared of doing nothing, or using `explicitToJson` if that's necessary).
However, I'm quite a new guy here and I may have missed something. Maybe others reviewers like @PrimaelQuemerais will agree with you.

PrimaelQuemerais (approved):

lib/model/EcoscoreData.dart
PrimaelQuemerais: I approved it for now, we still need to write the best practices document for the project, we should probably discuss this at some point. I will take a look at the `explicitToJson` parameter

peterwvj:
Thanks everyone for the feedback.

PR #73, opened 2021-01-29
allow anonymous photo upload

### LINKED ISSUE #72: Allow anonymous image uploads (without identification with an OFF account)
The OFF API, website and apps allow users to upload product photos even if they are not logged in.

In the Flutter package, we currently require an OFF username and password for all edits, including uploading photos. It would be a good thing to remove that restriction.

stephanegigandet:
Allows passing a null value for the user, to upload product photos without an OFF account.

fixes #72


teolemon:
How about requiring Global Accounts for apps ?

peterwvj:
Looks good, thanks!

PR #74, opened 2021-01-29
Updating product results (Adding countries and countries_tags to the product results)

MohamedFBoussaid:
Adding countries and countries_tags to the product result.


stephanegigandet (approved):
It looks good to me.

PR #75, opened 2021-01-29
Setting default language to OpenFoodFactsLanguage.WORLD and fields to…

MohamedFBoussaid:
Setting default language to OpenFoodFactsLanguage.WORLD and fields to ProductField.ALL if user does not provide them. 
Adding tests to cover changes. 
Updating tests.

### DIFF
--- a/example/pubspec.lock
+++ b/example/pubspec.lock
@@ -215,7 +215,7 @@ packages:
       path: ".."
       relative: true
     source: path
-    version: "0.3.13+2"
+    version: "0.3.14+1"
   package_config:
     dependency: transitive
     description:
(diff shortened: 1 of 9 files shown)


stephanegigandet (approved):
Looks good, thank you!

stephanegigandet (changes requested):
I approved this too fast. In fact the OFF API has no "world" language. There is only a "world" country.

We should either force the apps to specify a language, or use English as the default. (which is what the API assumes if no language is passed).

MohamedFBoussaid:
@stephanegigandet , my bad. I will fix this and set it to EN.

MohamedFBoussaid:
@stephanegigandet I have rebased my branch on the latest changes on the master. And with your changes on  #76, no need anymore to set a default value for the language. I am only setting the field to ALL in case this was not provided on the configuration.

stephanegigandet (approved):
Thank you!

Issue #76, opened 2021-01-30
Allow setting the cc (country code) field in API queries (product read and search)


stephanegigandet:
Some values returned by the API depend on the country of the user (e.g. the availability of the Eco-Score product attribute). We currently can set the language code (lc) parameter for the query, but not the country code (cc).

PR #82, opened 2021-02-04
add cc parameter to ProductSearchQueryConfiguration, #76

stephanegigandet:
I had forgotten to also add the cc parameter for ProductSearchQueryConfiguration (and not only ProductQueryConfiguration) when fixing #76.


MohamedFBoussaid (approved):
Can be merged to master.

PR #83, opened 2021-02-04
Using test db

MohamedFBoussaid:
Adding QueryType handler (PROD or TEST)
Updating tests
Adding tests using the test-db : world.openfoodfacts.net


stephanegigandet (approved):
Looks good to me, thank you very much!

peterwvj:
Great initiative. Didn't have the time to look into details yet. Could you explain how this test database work? Does it make the "actual" values fixed or is it more stable?

peterwvj:
Do all tests pass?

MohamedFBoussaid:
@peterwvj the idea is to save first the product (or create another state) on the test-db and then read the data of that product. And this can make it stable, and we can set what ever data we want even if it does not make sense without having the risk that those data will be used with any app on PROD. 
Other thing I have noticed is that the products on the test-db are not often changed, so we have more stable tests. 
But the tests can be always improved. 
P.S: The only cases that are not covered with the test-db are the robotoff EPs, bcs they don't existe on that db.

peterwvj:
@MohamedFBoussaid great to see that the tests have been improved. Thanks a lot for doing this.

The Robotoff tests seem to always fail anyway. I'm not familiar with those tests but right now they don't bring any value. Maybe we should just delete them for now? What do you think? @MohamedFBoussaid @stephanegigandet

Issue #84, opened 2021-02-05
All the images url are returning https://static.openfoodfacts.org/images/products/


MohamedFBoussaid:
Whatever the image requested on the PROD query, the returned URL is https://static.openfoodfacts.org/images/products/

This should be fixed on the ImageHelper here: 
    return queryType == QueryType.PROD
        ? IMAGE_PROD_URL_BASE
        : IMAGE_TEST_URL_BASE +
            barcodeUrl +
            "/" +
            image.field.value +
            "_" +
            image.language.code +
            "." +
            image.rev.toString() +
            "." +
            image.size.toNumber() +
            ".jpg";

MohamedFBoussaid:
Fixed and merged.

PR #85, opened 2021-02-05
Image url fix

MohamedFBoussaid:
Hot fix for the issue #84


stephanegigandet (approved):

monsieurtanuki:
@MohamedFBoussaid The tests don't pass anymore with some of your latest changes (e.g. `api_getProduct_test`)
The problem is that you try to deal with asset files (like `'assets/front_coca_light_de.jpg'`) as if they were normal files, and in flutter they are not. `Uri` is not adapted for that.

That leads me to another question, which can help us earn 10 points (cf. #90): what kind of files do you expect to use as front picture and so on? Is that going to be always `http`/`https` pictures?
I ask you that because `HttpHelper.doMultipartRequest` uses a syntax (`await File.fromUri(entry.value).readAsBytes()`) that prevents us from making the library available for flutter web. And if we're only dealing with `http`/`https` pictures, we could use a piece of code that would work also with flutter web (and earn 10 points).

MohamedFBoussaid:
@monsieurtanuki the image sendig to OFF was not implemented by me, on this fix, I have juste made some changes to switch between PROD and TEST DBs, so when devs are working on their apps they can use the TEST db instead of the PROD, to avoid having fake products on the real DB. 

And on the api_getProduct, I was following the examples that were made before. But I guess the expected is to have a pic from a given path (I dont think that people will uses assets on their apps, it is just here for the test purpose). But maybe instead we can get the pic on the test using a url, store it on a file and then send it to OFF using the file.path ?

monsieurtanuki:
@MohamedFBoussaid Exactly, here asset files don't make sense - what's the use-case?
And the JPEGs in assets take more than 3Mb, just for tests...
@stephanegigandet Isn't there a place on the server where we could put those files, like https://openfoodfacts.org/testFiles/front_coca_light_de.jpg?

MohamedFBoussaid:
@monsieurtanuki I agree, I have created [this issue](https://github.com/openfoodfacts/openfoodfacts-dart/issues/101) to update the assets with images from the OFF server. I will update them ASAP

monsieurtanuki:
@MohamedFBoussaid Perfect! Then the tests will pass - after a minor fine-tuning.

PR #88, opened 2021-02-05
Improve test output

### LINKED ISSUE #87: Clean up test output
Generally chatty tests/code should be avoided. This issue suggests getting rid of all the print statements in order to improve the test output.

peterwvj:
Closes #87

### DIFF
--- a/lib/utils/HttpHelper.dart
+++ b/lib/utils/HttpHelper.dart
@@ -63,8 +63,6 @@ class HttpHelper {
     // add all file entries to the request
     for (MapEntry<String, Uri> entry in files.entries) {
       List<int> fileBytes = await File.fromUri(entry.value).readAsBytes();
-      //print(fileBytes.toString());
-
       var multipartFile = new http.MultipartFile.fromBytes(entry.key, fileBytes,
           filename: basename(entry.value.toString()));
       request.files.add(multipartFile);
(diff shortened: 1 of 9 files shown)


lib/utils/HttpHelper.dart
```diff
@@ -99,7 +94,6 @@ class HttpHelper {
     if (isTestModeActive) {
       var token = 'Basic ' + base64Encode(utf8.encode('off:off'));
       headers.addAll({'authorization': token});
-      print("TEST-MODE");
```
MohamedFBoussaid, line 102: Maybe we can keep this log to confirm to the user that he is on test mode. 
it is not that big deal, and this is not a blocker to merge.

MohamedFBoussaid (approved):
Can be merged.

lib/utils/HttpHelper.dart
peterwvj: Hi @MohamedFBoussaid thanks for taking a look. If this print statement is there it is printed with every test, see
image
Would you be OK with leaving it out?
If it is not there the output looks more pleasant
image

If you still prefer to keep it there then I can add the statement back in.
What do you think?

MohamedFBoussaid: In that case let's just delete it, it does not worth it.

peterwvj: Will do, thanks!

Issue #93, opened 2021-02-07
Allow to get ingredient OCR output


teolemon:
Open Food Facts uses optical character recognition (OCR) to retrieve nutritional data and other information from the product labels.

Process

- [x] Capture the barcode of the product where you want to perform the OCR.
- [x] The Product Opener server software opens the image (process_image=1)
- [x] Product Opener returns a JSON response. Processing is done using Tesseract or Google Cloud Vision (recommended). The result is often cripped with errors with Tesseract, less with Google Cloud Vision.

Notes: * The OCR may contain errors. Encourage your users to correct the output using the ingredients WRITE API. * You can also use your own OCR, especially if to plan to send a high number of queries.

Parameters
Test server: https://world.openfoodfacts.org/cgi/ingredients.pl
code=code
id=imagefield
process_image=1

MohamedFBoussaid:
@teolemon, for the id parameter this will be only for ingredients ? (ingredients_en, ingredients_fr, ingredients_de ...) or it can be something else? an if we don't set the ocr_engine parameter this will use automatically the Google Cloud Vision, or it should be specified in all cases?

MohamedFBoussaid:
Implemented with #94

MohamedFBoussaid:
Closed the issue by mistake. 
@teolemon can you please confirm that we can close the issue. 
Thanks.

teolemon:
yes of course

PR #94, opened 2021-02-08
Adding ocr ingredients

MohamedFBoussaid:
Adding Extract ingredients using OCR. 
Disabling unstable tests.
Fix for #93


test/api_getRobotoff_test.dart
```diff
@@ -1,16 +1,17 @@
...
+//import 'package:openfoodfacts/utils/QueryType.dart';
+//import 'test_constants.dart';
+
 
 void main() {
 
   group('$OpenFoodAPIClient get robotoff questions', () {
     test('get questions for Noix de Saint-Jacques EN', () async {
-      RobotoffQuestionResult result =
+      /* RobotoffQuestionResult result =
```
peterwvj, line 14: I agree that these tests should be skipped. However, rather than commenting them out, I think it is better to use either the `skip` parameter or the `@Skip` annotation, see https://pub.dev/packages/test#skipping-tests

test/api_postRobotoff_test.dart
```diff
@@ -1,14 +1,15 @@
+
 import 'package:flutter_test/flutter_test.dart';
-import 'package:openfoodfacts/model/Insight.dart';
-import 'package:openfoodfacts/model/RobotoffQuestion.dart';
+//import 'package:openfoodfacts/model/Insight.dart';
```
peterwvj, line 3: Similar comment regarding skipping tests https://pub.dev/packages/test#skipping-tests

test/api_searchProducts_test.dart
```diff
@@ -98,7 +98,7 @@ void main() {
       expect(result.count > 30000, true);
     });
 
-    test('search products by keywords', () async {
+    /*test('search products by keywords', () async {
```
peterwvj, line 101: Similar comments about skipping tests, https://pub.dev/packages/test#skipping-tests
Alternatively, we can just completely delete the test

peterwvj (commented):
Overall this looks very good, thanks! My only remark is that it would be better to skip the tests rather than commenting them out (as per my other comments).

peterwvj:
And just for the sake of being clear: feel free to merge :)

test/api_searchProducts_test.dart
MohamedFBoussaid: Thanks @peterwvj, yeah you have right I will use the skip annotation, it is better and cleaner

test/api_postRobotoff_test.dart
MohamedFBoussaid: Thanks @peterwvj, yeah you have right I will use the skip annotation, it is better and cleaner

test/api_getRobotoff_test.dart
MohamedFBoussaid: Thanks @peterwvj, yeah you have right I will use the skip annotation, it is better and cleaner

lib/utils/OcrIngredientsField.dart
```diff
@@ -0,0 +1,21 @@
+enum OcrIngredientsField {
```
stephanegigandet, line 1: I think we should remove this enum, there are 200 languages, we cannot and should not list them all.

So instead of having the user specify "INGREDIENTS_DE" etc. the user should only specify the language code (lc) and set it to the string "de".

lib/openfoodfacts.dart
```diff
@@ -428,14 +432,44 @@ class OpenFoodAPIClient {
...
   }
 
+  /// Extract the ingredients from image with the given parameters.
+  /// The ingredients language should be given (ingredients_fr, ingredients_de, ingredients_en)
+  /// Returns the ingredients using OCR.
+  /// By default the query will use the Google Cloud Vision.
+  /// By default the query will hit the PROD DB
+
+  static Future<OcrIngredientsResult> extractIngredients(
+      User user, String barcode, OcrIngredientsField ocrIngredientsField,
```
stephanegigandet, line 450: Instead of a OcrIngredientsField, we can just ask for the 2 letter language code of the ingredients. Or for the corresponding OpenFoodFactsLanguage. That way we can completely remove the class for OcrIngredientsField. Less code.

stephanegigandet (changes requested):
It looks good, thank you!

The only thing is that I think we should not create a new OcrIngredientsField enum. Instead we can just ask to specify the language as a parameter, for instance as a OpenFoodFactsLanguage parameter. Then constructing the image id is just "ingredients_" + the 2 letter language code.

peterwvj (approved):

stephanegigandet (approved):
Thank you!

PR #100, opened 2021-02-13
fix/#90_2 - +10 points - conditional imports for "not web" and "web" apps

monsieurtanuki:
Web is now supported!

New files:
* `UriReader.dart`: Abstract reader of URI data, declined in "not web" and "web" versions
* `UriReader_io.dart`: Reader of URI data, "not web" version (i.e. supports `File`)
* `UriReader_js.dart`: Reader of URI data, "web" version
* `UriReader_stub.dart`: Needed for conditional imports

Impacted file:
* `HttpHelper.dart`: now using a UriReader

### DIFF
--- a/lib/utils/HttpHelper.dart
+++ b/lib/utils/HttpHelper.dart
@@ -1,7 +1,7 @@
+import 'package:openfoodfacts/utils/UriReader.dart';
 import 'package:path/path.dart';
 import 'package:http/http.dart' as http;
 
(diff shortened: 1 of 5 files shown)


MohamedFBoussaid (approved):
Some tests are failing bcs of the assets files, and those tests will be fixed by https://github.com/openfoodfacts/openfoodfacts-dart/issues/101

Thanks @monsieurtanuki

PR #102, opened 2021-02-14
Updating tests with the new test images assets folder

MohamedFBoussaid:
All the tests are passing now : Here is the build https://github.com/openfoodfacts/openfoodfacts-dart/runs/1897679261?check_suite_focus=false


example/pubspec.lock
```diff
@@ -418,4 +418,4 @@ packages:
     source: hosted
     version: "2.2.1"
 sdks:
-  dart: ">=2.10.0-110 <2.11.0"
+  dart: ">=2.12.0-0 <3.0.0"
```
monsieurtanuki, line 421: Careful, I think 2.12 is beta.
Maybe better to stay in 2.10 (stable) for the moment.

pubspec.lock
```diff
@@ -521,4 +521,4 @@ packages:
     source: hosted
     version: "2.2.1"
 sdks:
-  dart: ">=2.10.0 <2.11.0"
+  dart: ">=2.12.0-0.0 <3.0.0"
```
monsieurtanuki, line 524: Same 2.12 beta vs. 2.10 stable comment.

test/test_assets/json/recommended_daily_intakes_source _eu.json
```diff
@@ -0,0 +1,142 @@
+{
```
monsieurtanuki, line 1: I guess you should remove the same named file in `assets/json` then?

monsieurtanuki (commented):
It's OK @MohamedFBoussaid, except:
* I'm a bit surprised by the new dart version 2.12 (beta)
* there's a file to be deleted from `assets/json`
* there's a possible improvement in `UriReader`, in order to accept relative `File`s

Tell me what you think about that.

MohamedFBoussaid:
@monsieurtanuki for the dart version during the local build it was changed automatically, I will see how I can overwrite it.

For the json file, I didn't know what is made for, so I was affraid of deleting it, I will delete it. 

And I will have a look to the URI

monsieurtanuki:
About the dart version, here's what I read in https://dart.dev/guides/libraries/private-files:
> For library packages, don’t commit the pubspec.lock file. Regenerating the pubspec.lock file lets you test your package against the latest compatible versions of its dependencies.

MohamedFBoussaid:
@monsieurtanuki In that case I will add it to the ignore

lib/utils/HttpHelper.dart
```diff
@@ -61,7 +61,7 @@ class HttpHelper {
 
     // add all file entries to the request
     for (MapEntry<String, Uri> entry in files.entries) {
-      List<int> fileBytes = await UriReader.instance.readAsBytes(entry.value);
+      List<int> fileBytes = await UriReader.instance.readFileAsBytes(entry.value);
```
monsieurtanuki, line 64: Please stick to `readAsBytes`, we need to be able to read http resources too.

monsieurtanuki (changes requested):
We need to stay with `readAsBytes`, regardless of it is a File or a http that we're reading. Or even base64 data.

lib/utils/HttpHelper.dart
MohamedFBoussaid: @monsieurtanuki In that case the relative path will not work on tests, this is why in the first play I have used File().absolut.uri. 

I will change this back to readAsByte, and on the test I will make it again File().absolut.uri.

monsieurtanuki: That's why I suggested to change to algorithm in `UriReader` in a previous comment, so that relative `File`s work as well.
I tried that successfully with the `flutter test test` command - they say that it does not work that well with `flutter test`.
In the worst case, please rollback to your initial code with `absolute.uri` and `readAsBytes`.

MohamedFBoussaid: @monsieurtanuki I made the rollback.

monsieurtanuki: Thank you @MohamedFBoussaid! Sorry about these code changes back and forth.
I still think we would be better off with a `UriReader` that works with relative files - I'll fix that in another PR today.

monsieurtanuki (approved):

peterwvj:
Thanks for all the good work, @MohamedFBoussaid and @monsieurtanuki 

I was wondering whether the changes to the GithubAction are necessary. It seems like it was run twice
image

MohamedFBoussaid:
@peterwvj I have added it, bcs I hv noticed that if the PR is coming from a forked repo, It will be displayed on the checker tab, or the PR status.

MohamedFBoussaid:
Here two examples





The action will work only on the forked repo

PR #103, opened 2021-02-14
fix_3/#90 - support of relative files in UriReader

monsieurtanuki:
Impacted files:
* `api_addProductImage_test.dart`: `File`s are now relative
* `api_getProduct_test.dart`: `File`s are now relative
* `api_ocrIngredients_test.dart`: `File`s are now relative
* `UriReader.dart`: now support relative `File`s

### DIFF
--- a/lib/utils/UriReader.dart
+++ b/lib/utils/UriReader.dart
@@ -20,11 +20,12 @@ abstract class UriReader {
       return content;
     }
     switch (uri.scheme) {
+      case '':
       case 'file':
         return await readFileAsBytes(uri);
       case 'http':
       case 'https':
-        http.Response response = await http.get(uri);
+        final http.Response response = await http.get(uri);
         return response?.bodyBytes;
     }
     throw Exception('Unknown uri scheme for $uri');
--- a/test/api_addProductImage_test.dart
+++ b/test/api_addProductImage_test.dart
@@ -1,5 +1,3 @@
-import 'dart:io';
-
 import 'package:flutter_test/flutter_test.dart';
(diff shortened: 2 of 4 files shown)


MohamedFBoussaid:
You have right @monsieurtanuki it is better this way. 

Can you also please update example/main.dart ? Thanks

MohamedFBoussaid (approved):

monsieurtanuki:
Thank you @MohamedFBoussaid!
Unfortunately I'm not sure if I can change `example/lib/main.dart`, where we're not exactly dealing with tests that have access to files. And asset files is not supported either by `UriReader`.
I'll try today.

MohamedFBoussaid:
@monsieurtanuki I don't know if we are supposed to put a real file there, but I was thinking about a fake file just to show how to use the package

```
 SendImage image = SendImage(
    lang: OpenFoodFactsLanguage.FRENCH,
    barcode: "3613042717385",
    imageField: ImageField.INGREDIENTS,
    imageUrl: **File("Path to your image").uri**,
  ); 
```

monsieurtanuki:
@MohamedFBoussaid I've just had a look at the code of `main.dart` in details: it's full of hard-coded values, probably not usable without minor changes. And the message is explicit enough I think: `Uri.parse("Path to you image")`.
Unless we receive tons of mails of developers complaining that they have no idea what to put there, I think we can skip this final fix. Therefore, I'm finished :)

MohamedFBoussaid:
Yeah I guess it is fine to go without it. 

Because even the change that I am proposing, will not change anything just to make it clear how to use it. 

We can ignore that for the moment. 

Thanks @monsieurtanuki

PR #106, opened 2021-02-15
fix_4/#90 - better space management and code readability for SearchResult

monsieurtanuki:
Impacted files:
* `openfoodfacts.dart`: removed any reference to translated fields, that are now automatically computed
* `Product.dart`: automatically computed the translated categories and labels
* `ProductHelper.dart`: deprecated the method to translate fields, which is automatic now
* `SearchResult.dart`: removed the confusing and space-consuming raw `jsonProducts` field; let the `fromJson` method point to the json annotation generated method
* `SearchResult.g.dart`: removed the confusing and space-consuming raw `jsonProducts` field

### DIFF
--- a/lib/model/Product.dart
+++ b/lib/model/Product.dart
@@ -199,8 +199,20 @@ class Product extends JsonObject {
       this.ecoscoreGrade,
       this.ecoscoreScore});
 
-  factory Product.fromJson(Map<String, dynamic> json) =>
-      _$ProductFromJson(json);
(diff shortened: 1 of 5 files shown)


monsieurtanuki:
@peterwvj I've managed to keep the json annotations in `Product`, and just put the fix on top of it.

peterwvj:
Nice addition. Thanks a lot for doing this.

The PR looks good to me.

peterwvj (approved):

Issue #90, opened 2021-02-05
Improve "Pub Points" on pub.dev 


MohamedFBoussaid:
The current version has the following [score](https://pub.dev/packages/openfoodfacts/score)

![Screenshot 2021-02-05 at 17 38 58](https://user-images.githubusercontent.com/52571029/107062118-3b933780-67d9-11eb-9a3d-fd97d7a9a576.png)

Maybe we can improve the score, by following pub.dev recommandations.

stephanegigandet:
That's a great idea.

monsieurtanuki:
Currently working on the "Pass static analysis" score (for `0.3.15+2`: 20/30)

MohamedFBoussaid:
Thanks @monsieurtanuki

monsieurtanuki:
Hi @PrimaelQuemerais!
There's something fishy in your code in `SearchResult.dart` (compared to the generated `SearchResult.g.dart`)
The expected (and previous version) of method `fromJson` is:
```dart
factory SearchResult.fromJson(Map<String, dynamic> json) =>
    _$SearchResultFromJson(json);
```
But you changed it months ago, and now `_$SearchResultFromJson(json)` is never called. And its code is slightly different from the current implementation of `factory SearchResult.fromJson(Map<String, dynamic> json)`.
In the pub.dev score, we get a 10 point penalty because `_$SearchResultFromJson(json)` is never used. And it's also strange that the code is different. Could you please have a look at it? If you do that I'll give you 10 points :)

monsieurtanuki:
Hi @Grumpf!
In `HttpHelper.dart` you use `List<int> fileBytes = await File.fromUri(entry.value).readAsBytes();` to get the content of a `Uri`. Nothing that really shocks me, except that using `File` means using `dart.io`, which prevents developer from using openfoodfacts with flutter _web_. And it costs us 10 points.
I'm sure there's another way to do that; what about this:
```dart
http.Response response = await http.get(entry.value);
List<int> fileBytes = response?.bodyBytes;
```
Could you please have a look at it? If you do that I'll give you 10 points, fair is fair :)

monsieurtanuki:
Well, not everything is clear in my mind now, but we're getting closer to 110 points.

We lose 10 points because of this: ["At least 20% of the public API members contain API documentation."](https://dart.dev/guides/libraries/create-library-packages#documenting-a-library)

[...]

PR #110, opened 2021-02-28
Web support for more Pub Points

M123-dev:
For #90
We used `dart:io` which is not compatible with web. I switched to a third party package [universal_io ](https://pub.dev/packages/universal_io) which works the same but supports web.

> Support multiple platforms
> 10/20 points: Supports 2 of 3 possible platforms (iOS, Android, Web)
> 
> Consider supporting multiple platforms:
> Package not compatible with runtime flutter-web on Web
> 
> Because:
> 
>     package:openfoodfacts/openfoodfacts.dart that imports:
>     package:openfoodfacts/utils/HttpHelper.dart that imports:
>     dart:io

### DIFF
--- a/lib/utils/UriReader_io.dart
+++ b/lib/utils/UriReader_io.dart
@@ -1,4 +1,4 @@
-import 'dart:io';
+import 'package:universal_io/io.dart';
 import 'UriReader.dart';
 
 UriReader getUriReaderInstance() => UriReaderIo();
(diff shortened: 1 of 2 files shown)


monsieurtanuki (changes requested):
Hi @M123-dev!
We precisely don't need that anymore, since I split `UriReader` in 4 distinct files, the "web" version not using `dart:io`.

I've just run `pana`, and according to it:
```
## ✓ Support multiple platforms (20 / 20)
### [*] 20/20 points: Supports 3 of 3 possible platforms (**iOS**, **Android**, **Web**)
```

Therefore I suggest that we keep in mind that if one day we're in trouble, we can use the `universal_io` package. But for the moment, that is pointless and would add another dependency.

Issue #113, opened 2021-03-03
Null safety


M123-dev:
Big news. Flutter 2.0.0 has just entered the stable channel.
There are a lot of big changes. 💯 
Here all the [release notes](https://flutter.dev/docs/development/tools/sdk/release-notes) and the [annauncment](https://medium.com/flutter/whats-new-in-flutter-2-0-fe8e95ecc65).

The most important thing for the package, null safety, is now also stable. As I already noted in #90 , all packages we use are ready for null safety. That means we are now ready to upgrade.

MohamedFBoussaid:
@M123-dev The show was interesting, I have already give it a try during the show to migrate to null-safety, but it seems harder then showed.  I will give it a second try to see how it goes. 
But they did great job with the dart migrate tool
https://www.youtube.com/watch?v=bvq7wbn4VAA&ab_channel=Flutter

M123-dev:
@MohamedFBoussaid, how did your second try went

MohamedFBoussaid:
@M123-dev I will check it tonight and let you know. Didn't get time during the week end

M123-dev:
No need to rush

MohamedFBoussaid:
@M123-dev I guess we are blocked till json_serializable release a null safety version 

Example of errors that I am getting.
[SEVERE] json_serializable:json_serializable on lib/model/Ingredient.dart (cached):
Generator cannot target libraries that have not been migrated to null-safety.

Let's wait for until build_runner and json_serializable  release null safety version

M123-dev:
Where did this error occured?
I think this repo will take longer, but they said it should work. https://github.com/google/json_serializable.dart/issues/821#issuecomment-791787354

MohamedFBoussaid:
I am getting it when I try to generate the g files using : 
flutter packages pub run  build_runner build lib --delete-conflicting-outputs

But even using this does not help, I am still getting the errors: 
flutter packages pub run --mode=--no-sound-null-safety  build_runner build lib --delete-conflicting-outputs

stephanegigandet:
Same for me, I can't generate the g files using "flutter packages pub run build_runner build lib --delete-conflicting-outputs" , and I get this error: https://github.com/google/json_serializable.dart/issues/823

M123-dev:
I am not a core contributor to the plugin. What are the g files for and in which step are they generated?
Is the error message incorrect or can we  just generation either after the migration or beforehand with an older version of build_runner

[...]

PR #115, opened 2021-03-05
Upgraded dependencies

M123-dev:
Upgraded all the dependencies. The only one not supporting null safety is `build_runner `(dev_dependencies)

We have 7x `'nullable' is deprecated and shouldn't be used. Has no effect. (deprecated_member_use at [openfoodfacts] lib\model\Ingredient.dart:12` in the `@JsonKey()`. I left them in, for clearity when switching to null safety


M123-dev:
The test worfklow is still using flutter 1.22.x

monsieurtanuki (approved):
Hi @M123-dev!
Nothing that shocks me when reading the code.
Feel free to merge.

M123-dev:
Related: #113

PR #116, opened 2021-03-05
Update test-sdk.yml

M123-dev:


peterwvj:
Thanks, the new GH action picks up some new problems though

Screenshot_20210306-064442_Chrome

peterwvj (approved):

monsieurtanuki (commented):
Hi @M123-dev!
Disclaimer: I have no idea what this file is made for, and I don't know the impact of your changes either.
AFAIK your change looks reasonable.

monsieurtanuki:
@peterwvj Regarding your comment, it looks like the problem comes from code like this:
```dart
@JsonKey(nullable: false)
  String text;
```
Maybe `String? text` would be enough, while removing the `nullable: false` parameter?

M123-dev:
@monsieurtanuki @peterwvj Thanks for your review. The new infos in the analyzer are coming from the upgraded packages. I left them inside bc I guess they will be an important info when upgrading to null-safety.

PR #117, opened 2021-03-06
Fix analysis errors

peterwvj:
Get CI to work again

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Regarding this package, we accept pull requests as well feature requests.
 Execute the following command from the root of the repository to run the tests:
 
 ```
-flutter test test/
+flutter test
 ```
 
 ## How to use ?
--- a/lib/model/Ingredient.dart
+++ b/lib/model/Ingredient.dart
@@ -9,7 +9,7 @@ class Ingredient extends JsonObject {
   int rank;
   @JsonKey(includeIfNull: false)
   String id;
-  @JsonKey(nullable: false)
+  @JsonKey()
   String text;
   @JsonKey(includeIfNull: false, fromJson: JsonObject.parseDouble)
   double percent;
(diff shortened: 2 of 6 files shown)


lib/model/Ingredient.dart
```diff
@@ -9,7 +9,7 @@ class Ingredient extends JsonObject {
   int rank;
   @JsonKey(includeIfNull: false)
   String id;
-  @JsonKey(nullable: false)
+  @JsonKey()
```
M123-dev, line 12: Cant we just delete them if empty?

> Only required when the default behavior is not desired.

https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonKey/JsonKey.html

M123-dev (approved):
Thanks

lib/model/Ingredient.dart
peterwvj: It seems like you are right. There is a problem though. I'm currently not able to run build_runner because `json_serializable` has not been migrated to null safety

```
[SEVERE] json_serializable:json_serializable on lib/model/Nutriments.dart (cached):

Generator cannot target libraries that have not been migrated to null-safety.
package:openfoodfacts/model/Nutriments.dart:8:7

```

README.md
```diff
@@ -82,7 +82,7 @@ Regarding this package, we accept pull requests as well feature requests.
 Execute the following command from the root of the repository to run the tests:
 
 ```
-flutter test test/
+flutter test
```
monsieurtanuki, line 85: @peterwvj I know it's already been merged, but we added the `test/` directory for a good reason, around `UriReader`.
If `flutter test` now works without having to specify the directory, so much the better!

lib/model/Ingredient.dart
monsieurtanuki: @peterwvj For the record, it's possible to run a mix of null safe/unsafe: `flutter run --no-sound-null-safety`
I don't know how this can apply to `build_runner`.

PR #119, opened 2021-03-07
Migrate to null safety

### LINKED ISSUE #113: Null safety

peterwvj:
Migrate SDK to null safety.

Closes #113


M123-dev:
@peterwvj I havent read through the whole code but I noticed one thing. In #117 you deleted the deligate `nullable` from the `JsonKey()`s. They are deligate since this task has now been taken over by dart. However the values ​​that were marked as 
`nullable: false` before #117 are now nullable again.

lib/model/SendImage.dart
```diff
@@ -4,12 +4,12 @@ import '../interface/JsonObject.dart';
 import '../model/ProductImage.dart';
 
 class SendImage extends JsonObject {
-  OpenFoodFactsLanguage lang;
+  OpenFoodFactsLanguage? lang;
 
   // ignored for json
-  Uri imageUrl;
+  Uri? imageUrl;
```
stephanegigandet, line 10: at some point (not necessarily now), we probably should see which of those fields should be non-nullable. e.g. we should not allow SendImage to be called with a null imageUrl or a null barcode.

lib/model/AttributeGroup.dart
```diff
@@ -31,20 +31,20 @@ class AttributeGroup extends JsonObject {
...
+  final String? warning;
+  final List<Attribute>? attributes;
 
   @override
   String toString() => 'AttributeGroup(${toJson()})';
 
-  List<Map<String, dynamic>> _listToJson() {
-    if (attributes == null || attributes.isEmpty) {
+  List<Map<String, dynamic>>? _listToJson() {
+    if (attributes == null || attributes!.isEmpty) {
```
stephanegigandet, line 43: Just curious, what is the meaning of the !. operator?

M123-dev: The ! can be seen as a force push.
 When a function accepts only not nullable but you have to use a nullable you can still parse it with the !

lib/model/SendImage.dart
peterwvj: I agree. The work here has room for improvements but tweaking the code base is outside the scope of the PR. It would be too much work. Let's improve it as we go on.

lib/model/AttributeGroup.dart
peterwvj: Yeah so `attributes.isEmpty` would cause an error because `attributes` could potentially be null. However, in this case we can be certain that it is not so using `!` is fine.

peterwvj:
@M123-dev the #117 changes were made solely to fix analysis errors. Now the the code has been migrated to null safety with this PR, the fields cannot be non-null without making additional changes like requiring the field (using `@required`) which we probably should not. So I think it's OK as it is.

MohamedFBoussaid:
Look good for me. 
Thanks @peterwvj it was a huge task

MohamedFBoussaid (approved):

stephanegigandet (approved):
Thanks!

M123-dev (approved):
@peterwvj Ok, that's understandable. Thanks!

peterwvj:
Thanks for your feedback all. I'll go ahead and merge.

lib/model/AttributeGroup.dart
monsieurtanuki: Of course the game is to almost never use `!`, and use non-nullable fields/parameters instead. Whenever possible.

PR #120, opened 2021-03-07
add images urls, fixes #111

### LINKED ISSUE #111: Add urls for selected ingredients, nutrition, and packaging images (small + medium size)
We currently have this in Product.json

  @JsonKey(name: 'image_small_url', includeIfNull: false)
  String imgSmallUrl;

Which corresponds to the image_small_url field. But the API can return also images for ingredients, nutrition and packaging, in small and normal size (e.g. image_ingredients_small_url / image_ingredients_url).

I'd like to add corresponding fields, and also to rename imgSmallUrl to imageSmallUrl so that the class names match the API names. (we can keep imgSmallUrl for a while and mark it as deprecated).

stephanegigandet:
Added front / ingredients / nutrition / packaging URLs as described in #111 

I tried to mark imgSmallUrl as deprecated, but it doesn't seem to trigger dart analysis warning, maybe I'm not using it right:

 // Images
  @Deprecated('imgSmallUrl is deprecated, use imageFrontSmallUrl instead')
  @JsonKey(name: 'image_small_url', includeIfNull: false)
  String imgSmallUrl;

As remarked by @MohamedFBoussaid , it's not possible to generate the .g.dart files. I had to downgrade json_annotation and json_serializer to generate Product.g.dart from Product.dart:
https://github.com/openfoodfacts/openfoodfacts-dart/issues/113#issuecomment-792073160


lib/model/Product.dart
```diff
@@ -42,8 +42,32 @@ class Product extends JsonObject {
   OpenFoodFactsLanguage lang;
   @JsonKey(includeIfNull: false)
   String quantity;
+
+  // Images
+  @Deprecated('imgSmallUrl is deprecated, use imageFrontSmallUrl instead')
```
monsieurtanuki, line 47: Shouldn't it be `@deprecated` instead of `@Deprecated`?

monsieurtanuki (approved):
Hi @stephanegigandet!
I guess you downgraded the `pubspec.yaml` because we're still in the "null-safe" migration process, with a side-effect due to json code generation.
Beyond that, I approve.

peterwvj:
I don't think we need to downgrade dependencies. See https://github.com/openfoodfacts/openfoodfacts-dart/pull/119 which should probably be merged before changing the SDK further.

peterwvj (changes requested):
I think it's best to wait with this until #119 has completed. Also, I don't think downgrading the dependencies is necessary.

lib/model/Product.dart
stephanegigandet: if I put @deprecated('imgSmallUrl is deprecated, use imageFrontSmallUrl instead') instead of @Deprecated, I get an error: "Annotation must be either a const variable reference or const constructor invocation"

if I just put @deprecated (without a message), there's no error, but I don't see a warning either.

monsieurtanuki: My personal use is a comment (`///`) and then `@deprecated`. Then the deprecated method.
I've never used `@Deprecated`.
Perhaps `@JsonKey` then `@Deprecated`?...

peterwvj:
The null safety migration changes have caused conflicts, which need to be fixed. Sorry for the inconvenience.

stephanegigandet:
Fixed conflicts.

peterwvj:
The Github action complains about parse errors.

stephanegigandet:
Trying again. :)

MohamedFBoussaid (approved):
All good 
Checks pass

peterwvj (approved):

peterwvj:
Looks good, thanks :)

PR #121, opened 2021-03-07
Prepare release of version 1.0.0 of the SDK

peterwvj:


monsieurtanuki (commented):
I understand how important it can be to quickly make available a null-safe version.
But as @M123-dev or @stephanegigandet (can't remember) pointed out, some fields must definitely never be null, but as far as the migration went they're still `?` nullable.
I'm OK with a quick null-safe version (today?) as long as we're comfortable with the following breaking changes that will probably happen: "OK guys, this field is finally mandatory, sorry about that, we won't stay forever with a nullable field".

M123-dev:
We can probably release as a beta with a big disclaimer.

peterwvj:
Hi @monsieurtanuki thanks for your input. We can wait for a bit with releasing a new version of the SDK if people generally agree it needs additional "post migration" adjustments. By language design, all fields were nullable prior to migration. Whether we choose to introduce additional breaking changing now (non-nullable fields) with a fast release or a bit later I don't think makes much of a difference. Let's see what the other guys think.

MohamedFBoussaid (commented):
🎉🎉🎉🎉

MohamedFBoussaid:
> 🎉🎉🎉🎉

I agree with @M123-dev to release on beta, and we can tested and once we are sure of the beta version we can pass to 1.0.0

peterwvj:
That would work for me as well :)

peterwvj:
Alright, I updated the PR to use `1.0.0-beta` instead per your suggestions.

MohamedFBoussaid (approved):

PR #123, opened 2021-03-08
prefer_single_quotes 

M123-dev:
Not that much to explain. Dart prefers to get single quotes.
Is not that important but with the now natively built-in dart fix it was a quick change.


analysis_options.yaml
```diff
@@ -6,5 +6,6 @@ analyzer:
   errors:
     omit_local_variable_types: ignore
     annotate_overrides: ignore
```
monsieurtanuki, line 8: Maybe we would be better off with an explicit `prefer_single_quotes: warning`?

monsieurtanuki (commented):
Hi @M123-dev!
Please have a look at my comment.

analysis_options.yaml
M123-dev: Would make sense, because we don't let prs with analyzer warnings go through anyways and it is a "wrong" spelling and not something that has been forgotten like type annationations which are added when it works.

monsieurtanuki (approved):
Please test if it actually works (e.g. put double-quotes on a string and see if the dart analyzer complains about it).

peterwvj (approved):

M123-dev:
@monsieurtanuki tested

PR #124, opened 2021-03-09
add ProductField names for fields added in #111

stephanegigandet:
I had forgotten to add new entries in the ProductField enum corresponding to the fields added in #111


peterwvj:
CI fails because it now requires single qoutes. Other than that it's looking good.
image

stephanegigandet:
> CI fails because it now requires single qoutes.

Thanks, fixed.

peterwvj (approved):
Thanks!

Issue #125, opened 2021-03-10
Add vegan and vegetarian status to the Ingredient class


blazern:
Hi! :)

When a product is requested from Rest API of OFF, the response contains vegan/vegetarian status for each ingredient of the product.
But the `Ingredient` class from Dart SDK doesn't include the veg-status for some reason.

It seems that adding the fields wouldn't be too hard as Ingredient extends JsonObject and is parsed automatically.

peterwvj:
Thanks for reporting. Would you be willing to give it a try? :) Seems like a good first issue.

MohamedFBoussaid:
Thanks @blazern for the contribution.

peterwvj:
This was fixed with the PR so I'll go ahead and close it.

PR #126, opened 2021-03-10
Add 'vegan', 'vegetarian' and 'fromPalmOil' fields to the Ingredient class

blazern:
Adding 'vegan', 'vegetarian' and 'fromPalmOil' fields to the Ingredient class (for #125).

At first I wanted to use existing enums in `IngredientsAnalysisTags`, but they seem to be not very suitable for the purpose:
- They don't have an `ignore` value and it doesn't make sense for them to have such a value (e.g. a product cannot have an "ignore" vegan status).
- I would have to create 6 distinct `toJson` and `fromJson` functions for 3 distinct enums, which is a lot of boilerplate code.

### DIFF
--- a/lib/model/Ingredient.dart
+++ b/lib/model/Ingredient.dart
@@ -3,6 +3,8 @@ import '../interface/JsonObject.dart';
 
 part 'Ingredient.g.dart';
 
+enum IngredientSpecialPropertyStatus { POSITIVE, NEGATIVE, MAYBE, IGNORE }
+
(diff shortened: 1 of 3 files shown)


MohamedFBoussaid:
Thanks @blazern for the changes, 
can you fix the "Flutter format", it is failing here : https://github.com/openfoodfacts/openfoodfacts-dart/pull/126/checks

blazern:
Ran flutter format

Issue #108, opened 2021-02-23
Move from smoothie: match, ranked product, importance, ...


monsieurtanuki:
cf. https://github.com/openfoodfacts/smooth-app/issues/140

monsieurtanuki:
Done in #138.

PR #138, opened 2021-03-27
Moved product preferences and match from Smoothie, #108

stephanegigandet:
This PR is to make available in openfoodfacts-dart the product attributes system that is used to rank products according to user preferences (or app preferences).

You can see that system in action in the Smoothie app. This PR moves @monsieurtanuki 's code from Smoothie to openfoodfacts-dart.

@monsieurtanuki : I made some modifications so that dart would not complain about null-safety, but I may not have made the best choices to do so.


stephanegigandet:
Corresponding issue in openfoodfacts-dart: #108

Corresponding issue and discussion in smooth-app: https://github.com/openfoodfacts/smooth-app/issues/140

lib/model/Product.dart
```diff
@@ -261,4 +262,27 @@ class Product extends JsonObject {
 
   @override
   Map<String, dynamic> toJson() => _$ProductToJson(this);
+
+  // TODO(stephanegigandet): move to Product as non-static method getAttributes(final List<String> attributeIds)
```
monsieurtanuki, line 266: Remove the `// TODO` line.

monsieurtanuki (approved):
I have only one minor comment: please fix it, but I approve anyway.

lib/model/Product.dart
stephanegigandet: Oops, I forgot that line. Thanks.

monsieurtanuki (approved):

MohamedFBoussaid (approved):

monsieurtanuki:
@stephanegigandet It looks like we should get rid of the old `UserPreferences.dart` as well, in order to avoid confusion.

monsieurtanuki (approved):

peterwvj:
Some tests seem to be failing
Screenshot_20210328-181510_Samsung Internet

lib/personalized_search/available_attribute_groups.dart
```diff
@@ -0,0 +1,53 @@
...
+  static const String ATTRIBUTE_GROUP_INGREDIENT_ANALYSIS =
+      'ingredients_analysis';
+  static const String ATTRIBUTE_GROUP_NUTRITIONAL_QUALITY =
+      'nutritional_quality';
+  static const String ATTRIBUTE_GROUP_PROCESSING = 'processing';
+  static const String ATTRIBUTE_GROUP_ALLERGENS = 'allergens';
+  static const String ATTRIBUTE_GROUP_LABELS = 'labels';
+  static const String ATTRIBUTE_GROUP_ENVIRONMENT = 'environment';
+
+//  static const String ATTRIBUTE_VEGETARIAN = 'vegetarian';
```
peterwvj, line 40: Minor: there is a bit of commented out code. I'd just delete it 😊

monsieurtanuki: @peterwvj Fair enough. The following question is: would that be relevant instead to put a maximum of attribute ids (all of them?), rather than just a limited set?
I know that in smoothie I had to add or use some of those ids, and there's a good probability that someone who cares about sesame seeds or fair trade would be glad to find the ids already there as `const String`s.

peterwvj: Well, if they are helpful why not just comment them back in? 😊 My point is that commented out code rot.

monsieurtanuki: Sure, putting there the complete list would make sense - even uncommenting those lines wouldn't be comprehensive.

stephanegigandet: I uncommented and completed the strings (allergens etc.) and moved them to the Attribute and AttributeGroup classes.

peterwvj: Thanks. I have no further comments 👍

stephanegigandet:
> Some tests seem to be failing
> Screenshot_20210328-181510_Samsung Internet

I don't think I touched related code, and all tests seem to pass now.
It passes locally too:

~/openfoodfacts-dart$ flutter test test/api_ocrIngredients_test.dart 
00:26 +4: All tests passed!              

It may be that some tests don't pass if the server is overloaded or something similar.

peterwvj (approved):

peterwvj:
Approved, thanks!

monsieurtanuki (approved):
I've never managed to decide where those const Strings belong: I'm ok with those ids being moved to Attribute and AttributeGroup files.

PR #141, opened 2021-03-29
feature/#139 - code example around MatchedProduct

monsieurtanuki:
For the moment, only in the unit tests.

### DIFF
--- a/lib/personalized_search/product_preferences_manager.dart
+++ b/lib/personalized_search/product_preferences_manager.dart
@@ -1,4 +1,3 @@
-import 'package:meta/meta.dart';
 import 'package:openfoodfacts/model/Attribute.dart';
 import 'package:openfoodfacts/model/AttributeGroup.dart';
 import 'package:openfoodfacts/personalized_search/preference_importance.dart';
@@ -13,7 +12,6 @@ class ProductPreferencesManager {
   final ProductPreferencesSelection _productPreferencesSelection;
   AvailableProductPreferences? _availableProductPreferences;
 
-  @protected
   set availableProductPreferences(
     AvailableProductPreferences availableProductPreferences,
   ) =>
--- a/test/api_getProduct_test.dart
+++ b/test/api_getProduct_test.dart
@@ -14,6 +14,15 @@ import 'package:openfoodfacts/utils/UnitHelper.dart';
 
 import 'test_constants.dart';
 
+import 'package:http/http.dart' as http;
+import 'package:openfoodfacts/personalized_search/available_attribute_groups.dart';
+import 'package:openfoodfacts/personalized_search/available_preference_importances.dart';
+import 'package:openfoodfacts/personalized_search/available_product_preferences.dart';
+import 'package:openfoodfacts/personalized_search/matched_product.dart';
+import 'package:openfoodfacts/personalized_search/preference_importance.dart';
+import 'package:openfoodfacts/personalized_search/product_preferences_manager.dart';
+import 'package:openfoodfacts/personalized_search/product_preferences_selection.dart';
+
 void main() {
   group('$OpenFoodAPIClient get products', () {
     test('get product Coca Cola Light', () async {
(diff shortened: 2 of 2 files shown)


test/api_getProduct_test.dart
```diff
@@ -763,5 +772,86 @@ void main() {
       expect(vegetableFat.vegetarian, IngredientSpecialPropertyStatus.POSITIVE);
       expect(vegetableFat.fromPalmOil, IngredientSpecialPropertyStatus.MAYBE);
     });
+
+    test('matched product', () async {
+      final Map<String, String> attributeImportances = {};
+      int refreshCounter = 0;
+      final ProductPreferencesManager manager = ProductPreferencesManager(
+        ProductPreferencesSelection(
+          (String attributeId, String importanceIndex) async {
```
stephanegigandet, line 781: I'm thinking it would be good to use named parameters for ProductPreferencesSelection() so that this code can be understood when read, without having to know the parameters. (I already had forgotten what the 3 parameters were).

test/api_getProduct_test.dart
```diff
@@ -763,5 +772,86 @@ void main() {
...
+          AvailableAttributeGroups.getUrl(languageCode);
+      http.Response response;
+      response = await http.get(Uri.parse(importanceUrl));
+      expect(response.statusCode, httpOk);
+      final String preferenceImportancesString = response.body;
+      response = await http.get(Uri.parse(attributeGroupUrl));
+      expect(response.statusCode, httpOk);
+      final String attributeGroupsString = response.body;
+      manager.availableProductPreferences =
+          AvailableProductPreferences.loadFromJSONStrings(
```
stephanegigandet, line 804: Same here, named parameters would be useful to know which string comes first.

test/api_getProduct_test.dart
```diff
@@ -763,5 +772,86 @@ void main() {
...
+      expect(refreshCounter, 3);
+      await manager.setImportance(attributeId2, importanceId1);
+      expect(
+          manager.getImportanceIdForAttributeId(attributeId2), importanceId1);
+      expect(refreshCounter, 4);
+
+      matchedProduct = MatchedProduct(result.product!, manager);
+      expect(matchedProduct.score, 37.5);
+      expect(matchedProduct.status,
+          false); // because FOREST is mandatory, but unknown
```
stephanegigandet, line 854: That makes me realize that status has only 2 values. We need 3 to indicate whether products match preferences : "yes", "no", "unknown". There are some products that we know match all preferences, some that do not match at least one required preference, and others for which we don't have enough data.

On the web site we display the 3 categories differently:

image

I'll change the MatchedProduct code

stephanegigandet: Corresponding PR: https://github.com/openfoodfacts/openfoodfacts-dart/pull/142

lib/personalized_search/product_preferences_manager.dart
```diff
@@ -13,7 +12,6 @@ class ProductPreferencesManager {
   final ProductPreferencesSelection _productPreferencesSelection;
   AvailableProductPreferences? _availableProductPreferences;
 
-  @protected
```
peterwvj, line 16: Did you consider keeping `@protected` and adding `@visibleForTesting` instead?

monsieurtanuki: @peterwvj I did not, because I didn't know `@visibleForTesting` :)
That being said, the fact that I removed `@protected` has nothing to do with the fact that it's a unit test.
In the real world, there's no reason to force developers to create a class that overrides `ProductPreferencesManager` - which is mandatory if we keep the `@protected` annotation.
For the record, initially I added the `@protected` in the smoothie code that @stephanegigandet moved here. I can't remember the good idea behind it - except the joy of using a Java keyword that I thought did not exist in dart.

test/api_getProduct_test.dart
monsieurtanuki: @stephanegigandet Oh, I though `null` was the 3rd choice = "unknown". Isn't it?

test/api_getProduct_test.dart
monsieurtanuki: Good point. I'll do that.

test/api_getProduct_test.dart
monsieurtanuki: Good point. I'll do that too.

test/api_getProduct_test.dart
stephanegigandet: null indicates that the product has not been matched yet. It's different than unknown.

lib/personalized_search/product_preferences_selection.dart
```diff
@@ -5,11 +5,11 @@
...
 class ProductPreferencesSelection {
-  ProductPreferencesSelection(
-    this._setImportance,
-    this._getImportance,
-    this._notify,
-  );
+  ProductPreferencesSelection({
+    required this.setImportance,
+    required this.getImportance,
+    required this.notify,
```
stephanegigandet, line 11: I think the notify callback could be optional.

stephanegigandet (approved):
Thanks a lot for adding a working example in the test, it helps a lot to understand how the different classes fit together.

lib/personalized_search/product_preferences_selection.dart
monsieurtanuki: Good point! `setImportance` as well could be optional, as in my latest commit.

stephanegigandet (approved):
Perfect, thank you!

monsieurtanuki:
@stephanegigandet Another idea, that would hopefully remove some complexity: perhaps we could get rid of `ProductPreferencesSelection` and embed its callback fields into `ProductPreferencesManager`. What do you think of that?

PR #142, opened 2021-03-29
3 values for MatchedProduct status: yes, no, unknown (instead of bool)

stephanegigandet:
We need to differentiate products that we know do not match preferences from products that may or may not match preferences (products for which we don't have enough data to compute a known value for an attribute selected by the user).


peterwvj (approved):
Looks good, thanks!

monsieurtanuki (approved):
Beyond the fact that there's a potential confusion between `null` and `UNKNOWN`, I'm OK with that.

PR #143, opened 2021-03-29
switched to dart

M123-dev:
@teolemon mentioned on [slack](https://openfoodfacts.slack.com/archives/CC668J8AC/p1617001176000200) that a official Dart setup action got release. 

Because we used `flutter_test` for testing, this packages was a flutter package. I switched to the official [dart test package](https://pub.dev/packages/test) and updated the ci workflow. This should now be a flutter + dart package.


peterwvj (approved):
Looks good, thanks!

Issue #144, opened 2021-03-30
Rename Uri imageUrl; to Uri imageUri; in SendImage (breaking change)


stephanegigandet:
the imageUrl parameter for SendImage has confused multiple people. I think we should rename it to imageUri. It is a breaking change, but we should do it sooner rather than later.

blazern:
Seems this issue can be closed :)

PR #145, opened 2021-03-30
Prepare release of v1.1.0-beta (include breaking changes)

stephanegigandet:
The enum values for IngredientsAnalysisTags have changed, so it is a breaking change, this is why I bumped the middle version number.

While we are doing breaking changes, I think we should include a fix for #144 as well.


peterwvj (approved):
Thanks a lot 💪

PR #147, opened 2021-04-01
Rename SendImage.imageUrl -> SendImage.imageUri

blazern:
Renaming SendImage.imageUrl -> SendImage.imageUri for #144

### DIFF
--- a/test/api_getProductRaw_test.dart
+++ b/test/api_getProductRaw_test.dart
@@ -80,7 +80,7 @@ void main() {
           result.product!.nutrientLevels!.levels[NutrientLevels.NUTRIENT_SALT],
           Level.MODERATE);
       expect(result.product!.countries,
-          'Belgique,France,Allemagne,Pays-Bas,Espagne,Suisse');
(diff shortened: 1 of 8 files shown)


test/api_getProductRaw_test.dart
```diff
@@ -80,7 +80,7 @@ void main() {
           result.product!.nutrientLevels!.levels[NutrientLevels.NUTRIENT_SALT],
           Level.MODERATE);
       expect(result.product!.countries,
-          'Belgique,France,Allemagne,Pays-Bas,Espagne,Suisse');
+          'Belgique,France,Allemagne,Pays-Bas,Espagne,Suisse, en:world');
```
blazern, line 83: Doesn't look like it's related to my changes, but the test failed without `en:world`.

peterwvj (approved):
Looks good, thanks.

test/api_getProductRaw_test.dart
stephanegigandet: It's because an unrelated change happened on the dev database. I think we should remove that test.

stephanegigandet (approved):
Thank you!

Issue #151, opened 2021-04-04
no way to save a product with Dart SDK with `product_name_ru` or `ingredients_text_ru`


teolemon:
Reported on Slack by Danil @blazern 

- Seems like currently there's no way to save a product with Dart SDK with `product_name_ru` or `ingredients_text_ru` (only with `product_name` and `ingredients_text`).
- Which causes at least 2 problems:
-- It seems `product_name_en` is filled when `product_name` is filled, so Russian product names are written into `product_name_en` as if they're in English. For example: https://ru.openfoodfacts.org/продукт/4680015760456/
-- When user scans a barcode of a product which has its `product_name` in for example French, the user has no way to add a name in Russian because SDK doesn't support it.
- Would that be ok if I changed the SDK in a way that when a product is saved a language code can be used, and if it's used then its appended to each of the fields?
- So if a Russian user would save a new product, instead of `product_name` and `ingredients_text` fields `product_name_ru` and `ingredients_text_ru` would be sent to the Server.

Also, the same problem is present when a product is requested from the server - seems like currently there's no way to ask for Russian product name or ingredient list.

blazern:
I think it might be a good idea to remove from the Product class all members with lang code postfixes (like `productNameEN`) and for each translatable member add a neighbor-member with `Translated` postfix (some such fields already exist, like `categoriesTagsTranslated`).
So when SDK is asked to get a product with for example Russian translation, both `productName` and `productNameTranslated` would be filled - the first field with the international name and the second with the Russian name.

To request custom number of translated fields when a product is requested, I'd add a `List<ProductField> translatedFields` member to `ProductQueryConfiguration` - all items from this list would be requested from the server with the language code being appended.
This way the SDK users would be able to request both translated and international fields, or just 1 type (translated/international), or neither.

Then, when a new product or a product update is being sent to the server, users of the SDK would be able to provide both international and local fields by filling the `Translated` members.

blazern:
This issue can be closed now :)

MohamedFBoussaid:
Issue fixed, closing it based on @blazern request

PR #152, opened 2021-04-06
Add multiple '_translated' fields to Product

blazern:
Adding more translated product fields for #151

- Added several `Translated`-fields to Product.
- Marked the `de`, `en` and `fr` fields as deprecated because now (almost`*`) same functionality is available by the `translated` fields.
    `*` - "almost" - because now to get a product name in 3 those language the SDK user has to make 3 requests. If that's an issue - the `deprecated` annotation can of course be removed.
- Many fields of Product are reordered so that related fields would be declared close to each other.
- Product's constructor received new arguments it didn't have before for some reason.
- `Product.toJson` now fills translated fields when they're specified
- Product got a new method - `toValidatedData`, it's used instead of `toData` when a product is sent to the server, and it validates the product state before returning a result. This function would allow SDK users to find errors quicker.
- Extracted common logic from `ProductQueryConfigurations` and `ProductSearchQueryConfiguration` into a function in `ProductFields`. Also removed modification of the Configuration objects from this common logic - seemed strange that a `get`-function made significant state mutation.

### DIFF
--- a/test/api_getProduct_test.dart
+++ b/test/api_getProduct_test.dart
@@ -864,7 +864,7 @@ void main() {
     });
 
     test(
-        'translated fields when product is not translated into second language',
+        'translated fields when product is not translated into a second language',
         () async {
       String barcode = '3333333333333';
 
@@ -948,7 +948,7 @@ void main() {
       expect(russianProduct.countriesTagsTranslated, equals(['Россия']));
     });
 
-    test('translated fields when product is translated into second language',
+    test('translated fields when product is translated into a second language',
         () async {
       String barcode = '2222222222222';
 
--- a/test/api_getProduct_test.dart
+++ b/test/api_getProduct_test.dart
@@ -862,5 +862,184 @@ void main() {
       expect(matchedProduct.score, 0.0);
       expect(matchedProduct.status, MatchedProductStatus.YES);
     });
+
+    test(
+        'translated fields when product is not translated into second language',
+        () async {
+      String barcode = '3333333333333';
+
+      Product englishInputProduct = Product(
+        barcode: barcode,
+        lang: OpenFoodFactsLanguage.ENGLISH,
+        productNameTranslated: 'Pancakes',
+        ingredientsTextTranslated: 'Flour, water',
+        categories: 'Beverages',
+        countries: 'Russia',
+      );
+
+      await OpenFoodAPIClient.saveProduct(
+          TestConstants.TEST_USER, englishInputProduct,
+          queryType: QueryType.TEST);
+
+      final fields = [
+        ProductField.NAME,
+        ProductField.NAME_TRANSLATED,
+        ProductField.INGREDIENTS_TEXT,
+        ProductField.INGREDIENTS_TEXT_TRANSLATED,
+        ProductField.INGREDIENTS_TAGS,
+        ProductField.INGREDIENTS_TAGS_TRANSLATED,
+        ProductField.CATEGORIES_TAGS,
+        ProductField.CATEGORIES_TAGS_TRANSLATED,
+        ProductField.COUNTRIES_TAGS,
+        ProductField.COUNTRIES_TAGS_TRANSLATED,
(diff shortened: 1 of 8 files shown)


stephanegigandet (changes requested):
First of all, let me say that this is a very good idea to solve the problem of fields in different languages. Thank you!

Overall, it looks very good, but there is something a bit tricky: the "lang" field in the product indicates which language is the "main language" of the product. But products often have multiple languages. (e.g. in Belgium, they have both Dutch and French).

So we should not rely on the value of the lang field to get / set values for the translated fields.

The API deals with it with the "lc" field: it is used to tell what is the language of the fields sent to the API (which can be a different language than the main language of the product in the "lang" field).

The "lc" field of the API is set by the "language" or "lc" field of

class ProductQueryConfiguration {
  String barcode;
  OpenFoodFactsLanguage? language;
  // Allow apps to directly provide the language code and country code without
  // having to use the OpenFoodFactsLanguage helper.
  String? lc;
  String? cc;

We should also add it to saveProduct.

blazern:
> First of all, let me say that this is a very good idea to solve the problem of fields in different languages. Thank you!
> 
> Overall, it looks very good, but there is something a bit tricky: the "lang" field in the product indicates which language is the "main language" of the product. But products often have multiple languages. (e.g. in Belgium, they have both Dutch and French).
> 
> So we should not rely on the value of the lang field to get / set values for the translated fields.
> 
> The API deals with it with the "lc" field: it is used to tell what is the language of the fields sent to the API (which can be a different language than the main language of the product in the "lang" field).
> 
> The "lc" field of the API is set by the "language" or "lc" field of
> 
> class ProductQueryConfiguration {
> String barcode;
> OpenFoodFactsLanguage? language;
> // Allow apps to directly provide the language code and country code without
> // having to use the OpenFoodFactsLanguage helper.
> String? lc;
> String? cc;
> 
> We should also add it to saveProduct.

From your description it seems like if I changed the code to append `lc` instead of `lang.code`, then the patch overall would be more okayish. And this `lang.code -> lc` change doesn't seem big - just a couple of tweaks here and there.
Or am I getting something wrong?

stephanegigandet:
> From your description it seems like if I changed the code to append `lc` instead of `lang.code`, then the patch overall would be more okayish. And this `lang.code -> lc` change doesn't seem big - just a couple of tweaks here and there.

That's right.

blazern:
@stephanegigandet I think I did what you have asked from me but I'm not sure.
Please take a look :)

test/api_getProduct_test.dart
```diff
@@ -862,5 +862,184 @@ void main() {
...
+      Product englishInputProduct = Product(
+        barcode: barcode,
+        lang: OpenFoodFactsLanguage.ENGLISH,
+        productNameTranslated: 'Pancakes',
+        ingredientsTextTranslated: 'Flour, water',
+        categories: 'Beverages',
+        countries: 'Russia',
+      );
+
+      Product russianInputProduct = Product(
```
stephanegigandet, line 964: Looking at the test code, I'm thinking it would make sense to add a "translatedLang" field directly in the Product class, that would indicate what is the language of all the *Translated fields.

So when we request products with translated fields, we save the corresponding language to translatedLang.

And when we save the product, instead of passing lc to saveProduct, it justs uses the value of translatedLang.

What do you think?

test/api_getProduct_test.dart
```diff
@@ -862,5 +862,184 @@ void main() {
...
+        barcode: barcode,
+        lang: OpenFoodFactsLanguage.RUSSIAN,
```
stephanegigandet, line 966: so when you set lang: , it changes the "main language of the product" field.

test/api_getProduct_test.dart
blazern: Can do!

stephanegigandet (approved):
Perfect, thank you very much!

PR #153, opened 2021-04-07
Make the example subproject to depend on the parent folder

blazern:
The `example` subproject currently has `openfoodfacts: ^1.0.2-beta` as its dependency, which breaks its compilation since the latest openfoodfacts is `^1.1.0-beta`.

Seems that the example project should rather depend on the most up to date code, which is always stored in its parent folder.
This way no one will ever forget to change a version in its pubspec.yaml.


peterwvj (approved):
Thanks!

M123-dev (approved):
Thanks

PR #161, opened 2021-06-03
feat: add EcoscoreData.missingDataWarning field support

FredJul:
I have added the support for the `missing_data_warning` field, which allows the user to know the ecoscore may be improved.

However, I have decided to use a non-nullable boolean field, while the returned json use a `1` or empty value. I'm not sure why the server does not use boolean values but I believe it should be simplified here.

### DIFF
--- a/lib/interface/JsonObject.dart
+++ b/lib/interface/JsonObject.dart
@@ -33,6 +33,18 @@ abstract class JsonObject {
     }
   }
 
+  static bool parseBool(dynamic json) {
+    if (json is String) {
+      return json == '1';
+    } else if (json is int) {
+      return json == 1;
+    } else if (json is bool) {
+      return json;
+    } else {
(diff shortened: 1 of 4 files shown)


MohamedFBoussaid (approved):
All good, Thanks @FredJul

FredJul:
Thanks @MohamedFBoussaid :)

Just wonder if I should wait for another approve before merging it? What is the current process?
Also, do you know when a new version of the lib will be published on pub?

MohamedFBoussaid:
It is always good to have a second review. And regarding the lib release there is no roadmap for the releasing date. 

But I guess @stephanegigandet can help on this. Can you please review the PR ? And maybe we can release a new version with the latest changes?

peterwvj (approved):
Looks good, thanks!

Issue #159, opened 2021-05-22
Use the new auth.pl API to see if login was successful


teolemon:
https://world.openfoodfacts.org/cgi/auth.pl
Use the new auth.pl API to see if login was successful

This endpoint returns status 200 or 403 if the user is authentified, either through the "session" cookie, or with the user_id and password parameters.

M123-dev:
Implemented in #167

PR #167, opened 2021-06-22
Login via the auth.pl api

M123-dev:
For #159 

@stephanegigandet with my own login data it worked perfectly but with the dart test user it did not (of course on the test server), is this api available on there, is the dart test user already inside it? This would be interesting for tests

### DIFF
--- a/lib/openfoodfacts.dart
+++ b/lib/openfoodfacts.dart
@@ -444,6 +444,6 @@ class OpenFoodAPIClient {
     );
     Response response =
         await HttpHelper().doPostRequest(loginUri, user.toData(), user);
-    return response.statusCode == 200 ? true : false;
+    return response.statusCode == 200;
   }
 }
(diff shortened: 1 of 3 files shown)


lib/openfoodfacts.dart
```diff
@@ -435,11 +434,16 @@ class OpenFoodAPIClient {
...
+  static Future<bool> login(User user) async {
+    var loginUri = Uri(
+      scheme: URI_SCHEME,
+      host: URI_PROD_HOST,
+      path: '/cgi/auth.pl',
+    );
     Response response =
         await HttpHelper().doPostRequest(loginUri, user.toData(), user);
-    return response.body;
+    return response.statusCode == 200 ? true : false;
```
peterwvj, line 447: Just 
```dart
return response.statusCode == 200;
```
is cleaner

peterwvj:
Great extension, thanks!

M123-dev:
Thanks @peterwvj

Issue #131, opened 2021-03-14
Add support for Autocompleting values


teolemon:
Getting suggestions for states
https://world.openfoodfacts.org/cgi/suggest.pl?lc=fr&tagtype=states&string=a

Getting suggestions for languages
https://world.openfoodfacts.org/cgi/suggest.pl?lc=fr&tagtype=languages&string=a

Getting suggestions for labels
https://world.openfoodfacts.org/cgi/suggest.pl?lc=fr&tagtype=labels&string=a

Getting suggestions for categories
https://world.openfoodfacts.org/cgi/suggest.pl?lc=fr&tagtype=categories&string=a

Getting suggestions for Countries
https://world.openfoodfacts.org/cgi/suggest.pl?lc=fr&tagtype=countries&string=f

Getting suggestions for ingredients 
https://world.openfoodfacts.org/cgi/suggest.pl?lc=fr&tagtype=ingredients&string=f

Getting suggestions for traces
https://world.openfoodfacts.org/cgi/suggest.pl?lc=fr&tagtype=traces&string=m

Getting suggestions for additives
https://world.openfoodfacts.org/cgi/suggest.pl?lc=fr&tagtype=additives&string=e

Getting suggestions for allergens
https://world.openfoodfacts.org/cgi/suggest.pl?lc=fr&tagtype=allergens&string=a

Getting suggestions for packager codes
https://world.openfoodfacts.org/cgi/suggest.pl?lc=en&tagtype=emb_codes&EM

M123-dev:
Would be great for openfoodfacts/smooth-app/issues/147

teolemon:
Also for a future addition mode.

monsieurtanuki:
@M123-dev For https://github.com/openfoodfacts/smooth-app/issues/147 we would rather need something about "food" itself, I guess. And none of the suggestions is about food name.

M123-dev:
@monsieurtanuki oh didn't noticed, I would ofc suggest that too if the api allows it

MohamedFBoussaid:
@teolemon The changes were merged to master. 
Can we close this issue ?

teolemon:
yes

PR #169, opened 2021-06-24
#131 : Adding suggestion to the SDK

MohamedFBoussaid:
@teolemon I am not sure how to handle the packager codes. 
It always give the same result example: 
https://world.openfoodfacts.org/cgi/suggest.pl?lc=en&tagtype=emb_codes&EM
https://world.openfoodfacts.org/cgi/suggest.pl?lc=en&tagtype=emb_codes&EMS
https://world.openfoodfacts.org/cgi/suggest.pl?lc=en&tagtype=emb_codes&string=EH
https://world.openfoodfacts.org/cgi/suggest.pl?lc=en&tagtype=emb_codes&string=rgf

It always return: 
["BE 1-05 EC","BE 1-11 EC","BE 1-17 EC","BE 1-18 EC","BE 1-21 EC","BE 1-23 EC","BE 1-26 EC","BE 1-28 EC","BE 1-28-H EC","BE 1-31 EC","BE 1-38 EC","BE 1-39 EC","BE 1-42 EC","BE 1 EC","BE 10-09-2008 EC","BE 100783 EC","BE 100811 EC","BE 100904 EC","BE 100926 EC","BE 100939 EC","BE 100960 EC","BE 100963 EC","BE 101070 EC","BE 101082-H EC","BE 101202-H EC"]

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ You can check the terms of use here : [Terms of use](https://world.openfoodfacts
 ### PERSONALIZATION
 - [x] Recommended Daily Intakes
 - [x] Translated, high-level and customizable product information (Attributes)
+- [x] Autocompleting for the user inputs, by giving suggestions for (Labels, categories, ingredients, additives, traces ...)
 
 ### SEARCH
 - [x] search products by tags
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 
 import 'package:openfoodfacts/model/OcrIngredientsResult.dart';
 import 'package:openfoodfacts/openfoodfacts.dart';
+import 'package:openfoodfacts/utils/TagType.dart';
 
 /// request a product from the OpenFoodFacts database
 Future<Product?> getProduct() async {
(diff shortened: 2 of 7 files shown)


teolemon:
EMB and FR / DE / IT are the only things that will return results

teolemon:
@MohamedFBoussaid you're using `string` instead of `term`. This needs to be fixed for the EMB suggestions to work

MohamedFBoussaid:
@teolemon ok I will update it. I have used string instead of term, bcs I was following the examples that you gave on the issue, my bad I didn't search on the doc

MohamedFBoussaid:
> EMB and FR / DE / IT are the only things that will return results
> https://world.openfoodfacts.org/cgi/suggest.pl?tagtype=emb_codes&term=FR

By the way I was able to get results with RS, UK (don't know if there is other term that can be used



> @MohamedFBoussaid you're using `string` instead of `term`. This needs to be fixed for the EMB suggestions to work

the code is now updated.

M123-dev (approved):
Looks good, only a brief mention in the documentation would definitely be helpful

MohamedFBoussaid:
@M123-dev I will try to add something to the documentation before merging.

peterwvj (approved):

PR #177, opened 2021-07-10
fix_5/#90 - added comments to Parameter-related classes

monsieurtanuki:
On the way to a 130/130 pub.dev score...

By the way @teolemon @stephanegigandet: what's the purpose of `search_type=1` or `search_type=0` in the search API URI?

Impacted files:
* `ContainsAdditives.dart`: added a comment on class; refactored as non-nullable
* `OutputFormat.dart`: added a comment on class and enum; refactored
* `Page.dart`: added a comment on class; refactored as non-nullable
* `PageSize.dart`: added a comment on class; refactored as non-nullable
* `Parameter.dart`: added a comment on class; created method `addToMap`
* `ProductSearchQueryConfiguration.dart`: added a comment on class; refactored some fields as non-nullable
* `SearchSimple.dart`: not done much as the purpose of the class is unclear
* `SearchTerms.dart`: added a comment on class; refactored as non-nullable
* `SortBy.dart`: added a comment on class and enum; refactored
* `TagFilter.dart`: added a comment on class and enum; overridden new method `addToMap`; refactored

### DIFF
--- a/lib/interface/Parameter.dart
+++ b/lib/interface/Parameter.dart
@@ -1,4 +1,6 @@
 /// Abstract search API parameter, to be used in the search URI
+///
+/// Typical use will be 'name=value'
 abstract class Parameter {
   /// URI parameter name
   String getName();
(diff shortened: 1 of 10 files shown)


peterwvj (approved):
Looks good, thanks. I see the PR has two commits with the same message. I would suggest addressing this - perhaps by squashing the two commits - but it isn't critical.

monsieurtanuki:
Thank you @peterwvj for your code review!
Both commits have the same title, but not the same message. Let's say that good enough is good enough.

PR #178, opened 2021-07-12
feature/#155 - new method "OpenFoodAPIClient.getEcoscoreHtmlDescription"

monsieurtanuki:
Impacted files:
* `api_getProduct_test.dart`: added tests around the new method
* `openfoodfacts.dart`: created method `getEcoscoreHtmlDescription`

### DIFF
--- a/lib/openfoodfacts.dart
+++ b/lib/openfoodfacts.dart
@@ -542,12 +542,12 @@ class OpenFoodAPIClient {
   /// Returns the Ecoscore description in HTML
   static Future<String?> getEcoscoreHtmlDescription(
     final String barcode,
-    final String language,
+    final OpenFoodFactsLanguage language,
   ) async {
     const String FIELD = 'environment_infocard';
     final Uri uri = Uri(
       scheme: URI_SCHEME,
-      host: 'world-$language.openfoodfacts.org',
+      host: 'world-${language.code}.openfoodfacts.org',
       path: '/api/v0/product/$barcode.json',
       queryParameters: <String, String>{'fields': FIELD},
     );
--- a/lib/openfoodfacts.dart
+++ b/lib/openfoodfacts.dart
@@ -538,4 +538,31 @@ class OpenFoodAPIClient {
       return Status(status: 400, error: 'Unrecognized request error');
     }
   }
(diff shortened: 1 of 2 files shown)


lib/openfoodfacts.dart
```diff
@@ -538,4 +538,31 @@ class OpenFoodAPIClient {
       return Status(status: 400, error: 'Unrecognized request error');
     }
   }
+
+  /// Returns the Ecoscore description in HTML
+  static Future<String?> getEcoscoreHtmlDescription(
+    final String barcode,
+    final String language,
```
M123-dev, line 545: Also most other functions provide the optional `QueryType` parameter but I personally think its not that important

M123-dev (approved):
Thanks @monsieurtanuki, please have a look at my comments, otherwise the pr looks good.

peterwvj:
Besides @M123-dev useful comment I think the PR is looking good - thanks!

peterwvj (approved):

lib/openfoodfacts.dart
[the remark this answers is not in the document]
monsieurtanuki: Fair enough

monsieurtanuki: @MohamedFBoussaid Most other functions also have an optional `User` parameter, that we don't need here: should we put one there too, for consistency between functions? Will that be mandatory in the future?

Regarding `QueryType`, the thing is that the URI doesn't work in test mode. Actually we're not using `'world.openfoodfacts.org'` but `'world-fr.openfoodfacts.org'` (for French). And `'world-fr.openfoodfacts.net' does not seem to work for test environment. @teolemon What is the correct URI for test mode for the Ecoscore Html description?

In the meanwhile I can add an optional `QueryType` parameter and a `TODO` comment.

MohamedFBoussaid: @monsieurtanuki, in that case I agree with you. Let's not add the QueryType. 
Thanks for the changes, all good for me. It can be merged.

MohamedFBoussaid (approved):
Thanks @monsieurtanuki. 
All good.

Issue #107, opened 2021-02-16
Simplify the definition of search queries, remove some unneeded code


stephanegigandet:
There are a few things that could be simplified / removed from the definition of search queries:

e.g.

      var parameters = <Parameter>[
        const OutputFormat(format: Format.JSON),
        const Page(page: 5),
        const PageSize(size: 10),
        const SearchSimple(active: true),
        const SortBy(option: SortOption.PRODUCT_NAME),
        const TagFilter(
            tagType: "categories",
            contains: true,
            tagName: "breakfast_cereals"),
        const TagFilter(
            tagType: "nutrition_grades", contains: true, tagName: "A")
      ];

const OutputFormat(format: Format.JSON)  --> we just need JSON, the other formats like XML or JSONP are not needed, they are only available for clients that do not understand JSON

const SearchSimple(active: true),  --> this has absolutely no effect, it can be removed

I think we should remove all associated code. That means it would be a breaking change, and clients would have to remove it.

monsieurtanuki:
Fixed by #180.

PR #180, opened 2021-07-13
feature/#107 - deprecated useless parameters OutputFormat and SearchSimple

monsieurtanuki:
Impacted files:
* `api_searchProducts_test.dart`: removed all references to `OutputFormat` and `SearchSimple`
* `OutputFormat.dart`: deprecated because in fact always JSON
* `ProductSearchQueryConfiguration.dart`: explicit JSON output
* `SearchSimple.dart`: deprecated because useless

### DIFF
--- a/lib/model/parameter/OutputFormat.dart
+++ b/lib/model/parameter/OutputFormat.dart
@@ -1,21 +1,10 @@
 import 'package:openfoodfacts/interface/Parameter.dart';
 
-/// "Output format" search API parameter
+// TODO(monsieurtanuki): deprecated from 2021-07-13 (#107) because in fact always JSON; remove when old enough
+@deprecated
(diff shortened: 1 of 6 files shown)


M123-dev (commented):
I don't think I can properly review the PR. If the response is always json, it is certainly good to make the code simpler. But I don't know what what SearchSimple is. I think it's best to get a review from others, but if it takes too long, feel free to merge.

monsieurtanuki:
@M123-dev Oh I didn't invent anything, that's from issue #107 itself:

> const OutputFormat(format: Format.JSON) --> we just need JSON
const SearchSimple(active: true), --> this has absolutely no effect, it can be removed

M123-dev (approved):
Missed that, thanks

PR #182, opened 2021-07-13
feature/#105 - support for product list queries + configuration refactoring

monsieurtanuki:
New files:
* `AbstractQueryConfiguration.dart`: Abstract Query Configuration, that helps build API URI
* `ProductListQueryConfiguration.dart`: Query Configuration for multiple barcodes

Impacted files:
* `api_searchProducts_test.dart`: added a multiple product search test
* `ProductQueryConfiguration.dart`: refactored as now it extends new class `AbstractQueryConfiguration`
* `ProductSearchQueryConfiguration.dart`: refactored as now it extends new class `AbstractQueryConfiguration`
* `openfoodfacts.dart`: created method `getProductList`


M123-dev (approved):
Thanks @monsieurtanuki

lib/openfoodfacts.dart
```diff
@@ -201,6 +202,31 @@ class OpenFoodAPIClient {
...
+        path: 'products/${configuration.barcodes.join(',')}.json',
+        queryParameters: configuration.getParametersMap());
+
+    final Response response =
+        await HttpHelper().doGetRequest(uri, user: user, queryType: queryType);
+
+    final String jsonStr = _replaceQuotes(response.body);
+    final SearchResult result = SearchResult.fromJson(json.decode(jsonStr));
+
+    result.products!.asMap().forEach((index, product) {
```
peterwvj, line 223: We should probably guard against `null` rather than using `!`

lib/utils/AbstractQueryConfiguration.dart
```diff
@@ -0,0 +1,52 @@
+import 'package:openfoodfacts/utils/LanguageHelper.dart';
+import 'package:openfoodfacts/utils/ProductFields.dart';
+
+/// Abstract Query Configuration, that helps build API URI
+abstract class AbstractQueryConfiguration {
+  OpenFoodFactsLanguage? language;
+  // Allow apps to directly provide the language code and country code without
+  // having to use the OpenFoodFactsLanguage helper.
+  String? lc;
```
peterwvj, line 9: Rather than using `lc` and `cc` should we spell out the names for clarity? Or is this convention very common?

lib/utils/AbstractQueryConfiguration.dart
```diff
@@ -0,0 +1,52 @@
...
+      result.putIfAbsent('lc', () => lc!);
+    }
+
+    if (cc != null) {
+      result.putIfAbsent('cc', () => cc!);
+    }
+
+    if (fields != null) {
+      bool ignoreFieldsFilter = false;
+      for (ProductField field in fields!) {
```
peterwvj, line 38: Could we use something like `ignoreFieldsFilter = fields.any( (field) => field == ProductField.ALL)` instead to improve readablity?

peterwvj:
Overal, I think the change is useful. I've added a couple of detailed comments that may help you improve the PR. Most importantly, I noticed that CI raises a syntax error

image

lib/openfoodfacts.dart
monsieurtanuki: Fair enough. Initially I just copied the code from other methods. I'll fix all of them.

lib/utils/AbstractQueryConfiguration.dart
monsieurtanuki: I won't address that subject here as 1) I just reproduced was there before, and more important 2) there are currently a lot of talking about languages in #185.

lib/utils/AbstractQueryConfiguration.dart
monsieurtanuki: Fair enough. I'll do that.

lib/utils/AbstractQueryConfiguration.dart
peterwvj: Fair point 👍

PR #184, opened 2021-07-14
feature/#183 - using maps instead of switch for enums (refactoring)

monsieurtanuki:
Impacted files:
* `Ingredient.dart`
* `Insight.dart`
* `LanguageHelper.dart`
* `NutrientLevels.dart`
* `OcrField.dart`
* `PnnsGroups.dart`
* `ProductFields.dart`
* `ProductImage.dart`
* `SortBy.dart`
* `TagType.dart`
* `UnitHelper.dart`


blazern:
Why though?

monsieurtanuki:
> Why though?

@blazern "Why" can lead to an endless question-answer loop. Even children know that! ;)

If the question is more specifically "How relevant is it to refactor the code that way?", the answer can be found for instance in [`InsightTypesExtension`](https://github.com/openfoodfacts/openfoodfacts-dart/blob/master/lib/model/Insight.dart).
There, the "enumToString" method was coded in a `switch` statement. And the "stringToEnum" was coded in a redundant `switch` statement the other way around.
With this PR, the "stringToEnum" method is coded with a simple `Map` access. And the "enumToString" is coded without duplicating correspondances, by using the same `Map`.
Then, for code consistency and code clarity, other classes were impacted by the same kind of refactoring.

M123-dev (approved):
@monsieurtanuki I haven't read every line but cleaning the code is always a good idea.

peterwvj:
I support this kind of refactorng, I think the map definitions and look ups improve readability. There is a conflict that needs to be addressed though.

PR #185, opened 2021-07-14
Add ability to request and set multiple translated product fields at once

blazern:
Adding ability to request and set multiple translated product fields at once.

If some user of an OFF-based application knows multiple languages (e.g. English and German), it might be very useful for the user to be able to request and see a product fields translated into several languages.

For example, a product might have its name set in English but have a German ingredients list (no English ingredients list).
When a user requests such a product and the request's language is set to English, the user won't see ingredients list.
But if the user did let the app know they also know German language and the app would request the product with both English and German translation, then the user would get much more comprehensive data - both name and ingredients list of the product.

This pull request unfortunately has breaking changes (type of the `translated` fields is changed).
At first I tried to make no breaking changes but code started to look very-very clumsy.
The clumsy version of the changes is still stored on my computer so I can remove the breaking changes if necessary... Although I hope it's not necessary since the number of `translated` fields would increase 2x and it would be a bit harder to use the SDK.

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -76,15 +76,6 @@ List<String> miscTags
 List<String> stateTags
 List<String> tracesTags
 
-
----- Available for RAW products ----
-String productNameDE
-String productNameEN
-String productNameFR
-
-String ingredientsTextDE
-String ingredientsTextEN
-String ingredientsTextFR
 ```
 
 #### ProductImage
@@ -525,3 +516,143 @@ if(result.status != 1) {
 print("Upload was successful");
 ```
 
+## Notes on languages mechanics.
+
+TL;DR: use the "..InLangs" fields if you intend to display products data in
+specific language(s). You __must__ use the "..InLangs" fields if you intend
+to update products. Otherwise you might accidentally corrupt products by
+overwriting proper-language data with improper-language data.
+
+See detailed explanation and examples below.
+
+### Detailed explanation of languages mechanics.
+
+Most of end-user facing apps want to display products data to a user in
+the language(s) the user understands.
+In order for a developer to properly implement multilingual Open Food Facts
+usage, the developer might want to understand how Open Food Facts works with
+languages.
+
+Here are main concepts.
+
+Products in Open Food Facts can have some of their fields in multiple
+languages. Such fields are of 2 types:
+1. Fields with info taken directly from product packaging
+   (like product name and ingredients list). Such fields generally are
+   not touched by the backend and are stored in the DB exactly as they
+   were sent.
+2. Fields with meta-information (like list countries where the product is
+   being sold, or like product's categories). Such fields are processed
+   by the backend in different ways and often shouldn't be displayed to
+   the user as-is.
+
+In addition to that, fields of the [Product] class with multilingual
+support also are of 2 types:
+1. Fields with simple value (e.g. [Product.productName]).
+2. Fields with multilingual values (e.g. [Product.productNameInLangs]).
+
+Lastly, products in Open Food Facts have a main language. That's the
+language which will be used to fill fields of [Product] when no specific
+language settings were provided.
+Most likely you don't want to show the data to your users in the product's
+main language. Most likely you want to prefer data in some specific
+language.
+
+Let's say you want a product to be displayed in German and
+so you've set the value of [AbstractQueryConfiguration.language] to German.
+Here's how the described above types of fields work in combinations:
+1. 1-1 (packaging info + simple value).
+   Fields like [Product.productName] will be in German if the German
+   versions of them are available in OFF.
+   If the German version is not available, the fields will be in the main
+   product's language.
+2. 1-2 (packaging info + multilingual value).
+   Fields like [Product.productNameInLangs] will have the German value
+   inside if it's available.
+   If it's not available, `product.productNameInLangs[GERMAN]`
+   will be `null`.
+3. 2-1 (meta info + simple value).
+   Fields like [Product.countriesTags] __will not be in German__, ever.
+   They will have values like `["en:austria", "en:belgium", "en:canada"]`.
+4. 2-2 (meta info + multilingual value).
+   Fields like [Product.countriesTagsInLangs] will have German values
+   when available, if some of the values is not available in German it
+   will have a language prefix: `["Österreich ", "Belgien", "en:canada"]`.
+
+Those 4 combinations can be observed "in the wild" by
+executing next request:
+https://world.openfoodfacts.org/api/v2/product/3017620422003?lc=de&fields=product_name,product_name_de,countries_tags,countries_tags_de
+
+As you can see, the multilingual values should be prefered in most cases
+when product's data is displayed to a user.
+
+#### Example: display a product in German language or rollback to default.
+
+```dart
+final conf = ProductQueryConfiguration(
+          barcode,
+          language: OpenFoodFactsLanguage.GERMAN,
+          fields: [ProductField.NAME]);
+final product = await OpenFoodAPIClient.getProduct(conf);
+var name = product.productName ?? 'No name';
+_displayProductName(name);
+```
+
+#### Example: display a product in German language, or in Russian,
+#### or rollback to default.
+
+```dart
+final conf = ProductQueryConfiguration(
+          barcode,
+          language: OpenFoodFactsLanguage.GERMAN,
+          secondaryLanguages: [OpenFoodFactsLanguage.RUSSIAN],
+          fields: [ProductField.NAME, ProductField.NAME_IN_LANGS]);
+final product = await OpenFoodAPIClient.getProduct(conf);
+var name = product.productNameInLangs[OpenFoodFactsLanguage.GERMAN];
+name ??= product.productNameInLangs[OpenFoodFactsLanguage.RUSSIAN];
+name ??= product.productName;
+name ??= 'No name';
+_displayProductName(name);
+```
+
+#### Example: display product categories in German language.
+
+```dart
+final conf = ProductQueryConfiguration(
+          barcode,
+          language: OpenFoodFactsLanguage.GERMAN,
+          fields: [ProductField.CATEGORIES_TAGS_IN_LANGS]);
+final product = await OpenFoodAPIClient.getProduct(conf);
+var categories = product.categoriesTagsInLangs[OpenFoodFactsLanguage.GERMAN];
+
+// Erase not translated categories (e.g. "en:cake").
+categories = categories.where((c) => !c.startsWith(RegExp('\w+:')));
+
+_displayProductCategories(categories);
+```
+
+#### Example: update German product categories.
+
+```dart
+final conf = ProductQueryConfiguration(
+          barcode,
+          language: OpenFoodFactsLanguage.GERMAN,
+          fields: [ProductField.CATEGORIES_TAGS_IN_LANGS]);
+final product = await OpenFoodAPIClient.getProduct(conf);
+var categories = product.categoriesTagsInLangs[OpenFoodFactsLanguage.GERMAN];
+
+// Will have only categories which start with
+// lang prefixes (e.g. "en:cake").
+final notTranslatedCategories = categories.where((c) => c.startsWith(RegExp('\w+:')));
+// Translated categories only.
+final translatedCategories = categories.where((c) => !c.startsWith(RegExp('\w+:')));
+
+final updatedCategories = await _askUserToUpdateCategories(
+  translatedCategories);
+// Set updated German categories AND avoid
+// erasure of the not translated ones.
+product.categoriesTagsInLangs[OpenFoodFactsLanguage.GERMAN] =
+  updatedCategories + notTranslatedCategories;
+
+await OpenFoodAPIClient.saveProduct(_user, product);
+```
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -518,8 +518,8 @@ print("Upload was successful");
 
 ## Notes on languages mechanics.
 
-TL;DR: use the "..InLangs" fields if you intend to display products data in
-specific language(s). You __must__ use the "..InLangs" fields if you intend
+TL;DR: use the "..InLanguages" fields if you intend to display products data in
+specific language(s). You __must__ use the "..InLanguages" fields if you intend
 to update products. Otherwise you might accidentally corrupt products by
 overwriting proper-language data with improper-language data.
 
@@ -549,7 +549,7 @@ languages. Such fields are of 2 types:
 In addition to that, fields of the [Product] class with multilingual
 support also are of 2 types:
 1. Fields with simple value (e.g. [Product.productName]).
-2. Fields with multilingual values (e.g. [Product.productNameInLangs]).
+2. Fields with multilingual values (e.g. [Product.productNameInLanguages]).
 
 Lastly, products in Open Food Facts have a main language. That's the
 language which will be used to fill fields of [Product] when no specific
@@ -567,15 +567,15 @@ Here's how the described above types of fields work in combinations:
    If the German version is not available, the fields will be in the main
    product's language.
 2. 1-2 (packaging info + multilingual value).
-   Fields like [Product.productNameInLangs] will have the German value
+   Fields like [Product.productNameInLanguages] will have the German value
    inside if it's available.
-   If it's not available, `product.productNameInLangs[GERMAN]`
+   If it's not available, `product.productNameInLanguages[GERMAN]`
    will be `null`.
 3. 2-1 (meta info + simple value).
    Fields like [Product.countriesTags] __will not be in German__, ever.
    They will have values like `["en:austria", "en:belgium", "en:canada"]`.
 4. 2-2 (meta info + multilingual value).
-   Fields like [Product.countriesTagsInLangs] will have German values
+   Fields like [Product.countriesTagsInLanguages] will have German values
    when available, if some of the values is not available in German it
    will have a language prefix: `["Österreich ", "Belgien", "en:canada"]`.
 
@@ -606,11 +606,11 @@ _displayProductName(name);
 final conf = ProductQueryConfiguration(
           barcode,
           language: OpenFoodFactsLanguage.GERMAN,
-          extraLanguages: [OpenFoodFactsLanguage.RUSSIAN],
-          fields: [ProductField.NAME, ProductField.NAME_IN_LANGS]);
+          languages: [OpenFoodFactsLanguage.RUSSIAN],
+          fields: [ProductField.NAME, ProductField.NAME_IN_LANGUAGES]);
 final product = await OpenFoodAPIClient.getProduct(conf);
-var name = product.productNameInLangs[OpenFoodFactsLanguage.GERMAN];
-name ??= product.productNameInLangs[OpenFoodFactsLanguage.RUSSIAN];
+var name = product.productNameInLanguages[OpenFoodFactsLanguage.GERMAN];
+name ??= product.productNameInLanguages[OpenFoodFactsLanguage.RUSSIAN];
 name ??= product.productName;
 name ??= 'No name';
 _displayProductName(name);
@@ -622,9 +622,9 @@ _displayProductName(name);
 final conf = ProductQueryConfiguration(
           barcode,
           language: OpenFoodFactsLanguage.GERMAN,
-          fields: [ProductField.CATEGORIES_TAGS_IN_LANGS]);
+          fields: [ProductField.CATEGORIES_TAGS_IN_LANGUAGES]);
 final product = await OpenFoodAPIClient.getProduct(conf);
-var categories = product.categoriesTagsInLangs[OpenFoodFactsLanguage.GERMAN];
+var categories = product.categoriesTagsInLanguages[OpenFoodFactsLanguage.GERMAN];
 
 // Erase not translated categories (e.g. "en:cake").
 categories = categories.where((c) => !c.startsWith(RegExp('\w+:')));
@@ -638,9 +638,9 @@ _displayProductCategories(categories);
 final conf = ProductQueryConfiguration(
           barcode,
           language: OpenFoodFactsLanguage.GERMAN,
-          fields: [ProductField.CATEGORIES_TAGS_IN_LANGS]);
+          fields: [ProductField.CATEGORIES_TAGS_IN_LANGUAGES]);
 final product = await OpenFoodAPIClient.getProduct(conf);
-var categories = product.categoriesTagsInLangs[OpenFoodFactsLanguage.GERMAN];
+var categories = product.categoriesTagsInLanguages[OpenFoodFactsLanguage.GERMAN];
 
 // Will have only categories which start with
 // lang prefixes (e.g. "en:cake").
@@ -652,7 +652,7 @@ final updatedCategories = await _askUserToUpdateCategories(
   translatedCategories);
 // Set updated German categories AND avoid
 // erasure of the not translated ones.
-product.categoriesTagsInLangs[OpenFoodFactsLanguage.GERMAN] =
+product.categoriesTagsInLanguages[OpenFoodFactsLanguage.GERMAN] =
   updatedCategories + notTranslatedCategories;
 
 await OpenFoodAPIClient.saveProduct(_user, product);
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -537,11 +537,11 @@ Here are main concepts.
 
 Products in Open Food Facts can have some of their fields in multiple
 languages. Such fields are of 2 types:
-1. Fields with info taken directly from product packaging
+1. Text fields with info taken directly from product packaging
    (like product name and ingredients list). Such fields generally are
    not touched by the backend and are stored in the DB exactly as they
    were sent.
-2. Fields with meta-information (like list countries where the product is
+2. Lists of tags with meta-information (like list countries where the product is
    being sold, or like product's categories). Such fields are processed
    by the backend in different ways and often shouldn't be displayed to
    the user as-is.
@@ -594,18 +594,19 @@ final conf = ProductQueryConfiguration(
           language: OpenFoodFactsLanguage.GERMAN,
           fields: [ProductField.NAME]);
 final product = await OpenFoodAPIClient.getProduct(conf);
+// The requested language is German, so `product.productName`
+// either is in German, or in main product's language, or null.
 var name = product.productName ?? 'No name';
 _displayProductName(name);
 ```
 
-#### Example: display a product in German language, or in Russian,
-#### or rollback to default.
+#### Example: display a product in German language, or in Russian, or rollback to default.
 
 ```dart
 final conf = ProductQueryConfiguration(
           barcode,
           language: OpenFoodFactsLanguage.GERMAN,
-          secondaryLanguages: [OpenFoodFactsLanguage.RUSSIAN],
+          languages: [OpenFoodFactsLanguage.RUSSIAN],
           fields: [ProductField.NAME, ProductField.NAME_IN_LANGS]);
 final product = await OpenFoodAPIClient.getProduct(conf);
 var name = product.productNameInLangs[OpenFoodFactsLanguage.GERMAN];
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -606,7 +606,7 @@ _displayProductName(name);
 final conf = ProductQueryConfiguration(
           barcode,
           language: OpenFoodFactsLanguage.GERMAN,
-          languages: [OpenFoodFactsLanguage.RUSSIAN],
+          extraLanguages: [OpenFoodFactsLanguage.RUSSIAN],
           fields: [ProductField.NAME, ProductField.NAME_IN_LANGS]);
 final product = await OpenFoodAPIClient.getProduct(conf);
 var name = product.productNameInLangs[OpenFoodFactsLanguage.GERMAN];
--- a/lib/utils/AbstractQueryConfiguration.dart
+++ b/lib/utils/AbstractQueryConfiguration.dart
@@ -3,38 +3,53 @@ import 'package:openfoodfacts/utils/ProductFields.dart';
 
 /// Abstract Query Configuration, that helps build API URI
 abstract class AbstractQueryConfiguration {
+  /// Use to set the desired product language. Cannot be used together with
+  /// the [languages] field.
+  /// See also the comment to the [languages] field.
   OpenFoodFactsLanguage? language;
 
-  /// The [extraLanguages] field should be used when a product is requested
-  /// with fields in multiple languages. I.e. when some of the
-  /// `IN_LANGS` fields are used (e.g. [ProductField.NAME_IN_LANGS]).
+  /// The [languages] field should be used when a product is requested
+  /// with fields in multiple languages. I.e.:
+  /// - when some of the `IN_LANGUAGES` fields are used
+  ///   (e.g. [ProductField.NAME_IN_LANGUAGES]).
+  /// - when Product's text fields are wanted to be filled with either
+  ///   of several languages in prioritized manner
   ///
-  /// However, the `IN_LANGS` fields are also compatible with the [language]
-  /// and [lc] fields - if only 1 language is needed, [extraLanguages] can
-  /// be omitted.
-  List<OpenFoodFactsLanguage> extraLanguages;
-  // Allow apps to directly provide the language code and country code without
-  // having to use the OpenFoodFactsLanguage helper.
+  /// However, the `IN_LANGUAGES` fields are also compatible with the [language]
+  /// field - if only 1 language is needed, [language] can be used.
+  ///
+  /// Please see https://github.com/openfoodfacts/openfoodfacts-dart/blob/master/DOCUMENTATION.md#about-languages-mechanics
+  /// for detailed explanation on how to work with multiple languages.
+  List<OpenFoodFactsLanguage> languages;
+
+  /// Deprecated: please use [language] or [languages]
+  @deprecated
   String? lc;
+
   String? cc;
   List<ProductField>? fields;
 
   AbstractQueryConfiguration({
     this.language,
-    this.extraLanguages = const [],
+    this.languages = const [],
     this.lc,
     this.cc,
     this.fields,
   }) {
     fields ??= [ProductField.ALL];
+    if ((language != null || lc != null) && languages.isNotEmpty) {
+      throw ArgumentError(
+          '[languages] cannot be used together with [language]/[lc]');
+    }
   }
 
   /// Returns the corresponding API URI parameter map
   Map<String, String> getParametersMap() {
     final Map<String, String> result = {};
 
-    if (language != null) {
-      result.putIfAbsent('lc', () => language.code);
+    final languages = language != null ? [language!] : this.languages.toList();
+    if (languages.isNotEmpty) {
+      result.putIfAbsent('lc', () => languages.map((e) => e.code).join(','));
     } else if (lc != null) {
       result.putIfAbsent('lc', () => lc!);
     }
@@ -48,15 +63,6 @@ abstract class AbstractQueryConfiguration {
         (field) => field == ProductField.ALL,
       );
       if (!ignoreFieldsFilter) {
-        final languages = extraLanguages.toList();
-        if (language != null && !languages.contains(language)) {
-          languages.insert(0, language!);
-        }
-        final lcLanguage = LanguageHelper.fromJson(lc);
-        if (lcLanguage != OpenFoodFactsLanguage.UNDEFINED &&
-            !languages.contains(lcLanguage)) {
-          languages.insert(0, lcLanguage);
-        }
         final fieldsStrings = convertFieldsToStrings(fields!, languages);
         result.putIfAbsent('fields', () => fieldsStrings.join(','));
       }
--- a/lib/utils/AbstractQueryConfiguration.dart
+++ b/lib/utils/AbstractQueryConfiguration.dart
@@ -5,14 +5,14 @@ import 'package:openfoodfacts/utils/ProductFields.dart';
 abstract class AbstractQueryConfiguration {
   OpenFoodFactsLanguage? language;
 
-  /// The [languages] field should be used when a product is requested
+  /// The [extraLanguages] field should be used when a product is requested
   /// with fields in multiple languages. I.e. when some of the
   /// `IN_LANGS` fields are used (e.g. [ProductField.NAME_IN_LANGS]).
   ///
   /// However, the `IN_LANGS` fields are also compatible with the [language]
-  /// and [lc] fields - if only 1 language is needed, [languages] can
+  /// and [lc] fields - if only 1 language is needed, [extraLanguages] can
   /// be omitted.
-  List<OpenFoodFactsLanguage> languages;
+  List<OpenFoodFactsLanguage> extraLanguages;
   // Allow apps to directly provide the language code and country code without
   // having to use the OpenFoodFactsLanguage helper.
   String? lc;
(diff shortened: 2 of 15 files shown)


blazern:
Tests fail because certificate of openfoodfacts.net has expired :(

monsieurtanuki:
@blazern I have the same problem with tests failing because of certificates.

blazern:
> @blazern I have the same problem with tests failing because of certificates.

You can run them locally thought - you just need to set your local date to 12th of July (the last date before the certificate expiration).
But of course that doesn't help with pull requests...

blazern:
Remade the commit in order for the tests to restart

test/json_product_generation_test.dart
```diff
@@ -20,7 +20,7 @@ void main() {
     String barcode = '0030000010204';
     ProductQueryConfiguration configurations = ProductQueryConfiguration(
         barcode,
-        language: OpenFoodFactsLanguage.ENGLISH,
+        languages: [OpenFoodFactsLanguage.ENGLISH],
```
monsieurtanuki, line 23: Too bad none of your tests include at least 2 languages.

lib/utils/ProductSearchQueryConfiguration.dart
```diff
@@ -5,21 +5,40 @@ import 'package:openfoodfacts/utils/ProductFields.dart';
...
+      this.lc,
+      this.cc,
+      this.fields,
+      required this.parametersList}) {
+    if (languages.isNotEmpty && language != null) {
+      throw ArgumentError(
+          'Please provide either language or languages, not both');
+    }
+    if (language != null) {
+      languages = [language];
```
monsieurtanuki, line 39: Things are getting even more confused between `lc`, `languages` and `language`.

lib/utils/LanguageHelper.dart
```diff
@@ -590,4 +590,42 @@ class LanguageHelper {
     }
     return OpenFoodFactsLanguage.UNDEFINED;
   }
+
+  static Map<String, String>? toJsonStringMap(
+      Map<OpenFoodFactsLanguage, String>? map) {
+    return toJsonMap(map);
+  }
+
+  static Map<OpenFoodFactsLanguage, String>? fromJsonStringMap(
```
monsieurtanuki, line 599: I'm not sure I understand the point of creating several methods that do the same thing - except for a simple cast of data. In that case please comment each method.

monsieurtanuki (commented):
Hi @blazern!

I understand the initial need this way: getting the best translations, given the languages spoken by the end-user.
I'm not sure it should impact that much the "interface".

Food for thoughts:
* assuming that I speak French, English and Italian (in that order)
* I would like to say that my main language is French (like I do now)
* and that I also have an ordered list of other spoken languages (`[ENGLISH, ITALIAN]`)
* that would mean that as parameters, I keep the same mandatory "main" language (`language` or `lc`), and that I add an optional "additional languages" (`List<OpenFoodFactsLanguage>`)
* doing so, we don't break the previous usage
* then, when I get the data, I get potentially several translations for each relevant field
* but that could be transparent for the developer too: asking for (French + [English, Italian]) translations, "_translated" fields should be automatically populated with French, or else English, or else Italian translations.
* doing so, we don't break the previous usage either - just using "_translated" fields

What do you think of that?

test/json_product_generation_test.dart
blazern: But... but there's a new test with 3 languages: `multiple translated fields` in `test/api_getProduct_test.dart`

lib/utils/ProductSearchQueryConfiguration.dart
blazern: You're right, it's becoming more confusing.

But while the `Product.*translated` fields were added by me not very long time ago and I felt confident enough to replace the old ones with the new ones because it's likely used by my project only, the `*Configuration.language` field was added long before that, so the field probably is used by a lot of people.

Some solutions I see:
- completely remove the `language` field, make the `languages` field to be the only one.
- return the old form to the `language` field, replace `languages` field with `translationsLanguages` field, which would be used only for the `Product.*translated` fields (the old `language` field will be used to set `'lc'` as before).

What do you think?
Maybe there's a better way I don't see?

lib/utils/LanguageHelper.dart
blazern: Yes, they're for casts.
To be precise, they are for the `fromJson` and `toJson` properties of the `@JsonKey` annotation.
The annotation does not support generic functions, otherwise those extra not-generic functions wouldn't be needed.

I'll add comments :)

blazern:
> Hi @blazern!
> 
> I understand the initial need this way: getting the best translations, given the languages spoken by the end-user.
> I'm not sure it should impact that much the "interface".
> 
> Food for thoughts:
> 
>     * assuming that I speak French, English and Italian (in that order)
> 
>     * I would like to say that my main language is French (like I do now)
> 
>     * and that I also have an ordered list of other spoken languages (`[ENGLISH, ITALIAN]`)
> 
>     * that would mean that as parameters, I keep the same mandatory "main" language (`language` or `lc`), and that I add an optional "additional languages" (`List<OpenFoodFactsLanguage>`)
> 
>     * doing so, we don't break the previous usage
> 
>     * then, when I get the data, I get potentially several translations for each relevant field
> 
>     * but that could be transparent for the developer too: asking for (French + [English, Italian]) translations, "_translated" fields should be automatically populated with French, or else English, or else Italian translations.
> 
>     * doing so, we don't break the previous usage either - just using "_translated" fields
> 
> 
> What do you think of that?

Hello @monsieurtanuki! Thank you for the review! :)
Sorry for the wall of text below - it's a bit hard for me to be not verbose about this complicated topic :)

While I see the benefits of the approach you have described, I think it might do more harm than good in the long run.

First thing is that we definitelly should have Product fields with type `Map<Language, TranslatedField>`, because that's how the backend gives data to us. When we request multiple translated fields (e.g. `product_name_en, product_name_ru`), the backend just gives us the data wihout thinking of what language the end-user would prefer.
If we don't add such but instead keep the old `_translated` fields, sooner or later some of the SDK users will create issues on the issues tracker where they will ask for the ability to access multiple translations at once.
Worse than that, some of the SDK users will simply start making several requests for a single product - a request for the product with English translations, then with French, then with Italian, ...

> but that could be transparent for the developer too: asking for (French + [English, Italian]) translations, "_translated" fields should be automatically populated with French, or else English, or else Italian translations.

I think the SDK should not do somethink like this...
Presently the Product class is more-or-less a POD-type. It represents data as it is on the OFF backend and doesn't try to be clever.
Which is good, because we cannot predict all ways how people could use the SDK and cannot make clever functions for all use-cases.

For example, what if the app which uses the OFF SDK gives to the end-user an ability to manually prioritize the languages they want to see? Some of the users know French better than Italian, others know Italian better than French, and the app would give them an ability to set priorities manually. If we want the `Product` class to be clever, then it would need to have a `List<Language> priorities` field. But OFF products don't have such a concept as "language priorities" on the backend or anywhere else - priorities are for UI, not for the low-level code.

---

Also, it might be worth noting that the old `translated` fields were added only 3 months ago and _probably_ used only by my Flutter app and by no one else.
OFF Dart SDK was created in 2018 and the `translated` fields were not needed by anyone before - while this is a breaking change, its consequences most likely will not affect anyone.

monsieurtanuki:
Regarding the fact that "_translated" fields was a recent change and that it would probably be harmless to transform them, I understand your point but I don't have enough hindsight about it.

What I'm suggesting is actually to rename your `Map<languageString, translatedLabel>`, that currently use the name of fields without the same type.
We keep all your code changes with slightly different field names (like "_translations"), and populate the corresponding already existing Product "_translated" field.
And your "translations" fields are `public` (for an "expert" mode I guess, e.g. an OFF translation app).

The language order is indeed not supposed to be kept inside the `Product`; we just say
1. give me the translated labels for those 3 languages (API call)
2. by the way I prefer French, then English, then Italian, so populate the "_translated" field with a specific method call (inside dart, like `Product.populateTranslated(List<OpenFoodFactsLanguage>)`, just after we receive and decode the json data)

Is that too smart a move for this class, which is supposed to be a plain receiver of API data?
The question is where you put the cursor. In a very lazy version, we could even only use a `Map<String, dynamic>` for the whole `Product` object. It makes more sense to help the developers with `int?` and `double?` fields, no doubt about that.
It would also help developers too not to tip-tap around languages, hence my suggestion.

> what if the app which uses the OFF SDK gives to the end-user an ability to manually prioritize the languages they want to see

Of course the end-user may have the possibility to prioritize the languages. It's not an app designed only for people that speak those languages in that order ;) But that's not a problem: we keep "your" translations fields, and re-populate the "translated" fields each time the end-user change their mind about the language priority (which is not supposed to happen too often either).

lib/utils/ProductSearchQueryConfiguration.dart
monsieurtanuki: Keeping `lc`, `language` and `languages` is a mess.
I guess we could deprecate both `language` (as you suggested) and `lc`.

test/json_product_generation_test.dart
monsieurtanuki: Fair enough. It didn't see it at once: so many changes in that file ("language" into "languages") that github didn't want to display the file unless you ask politely!

blazern:
Github's discussion system are clumsy :(

> Is that too smart a move for this class, which is supposed to be a plain receiver of API data?
> The question is where you put the cursor. In a very lazy version, we could even only use a `Map<String, dynamic>` for the whole `Product` object. It makes more sense to help the developers with `int?` and `double?` fields, no doubt about that.
> It would also help developers too not to tip-tap around languages, hence my suggestion.

My main thought is that the cursor is already put. The API is already stable. The Product class is already what it is - it already converts the backend data into nice readable form, and at the same time it doesn't try to do anything clever.
I'd vote for leaving the abstraction level as it currently is. If we add 1 clever thing to the class now, in a year the class might have 10 clever things and might grow 2x in size.

> What I'm suggesting is actually to rename your Map<languageString, translatedLabel>, that currently use the name of fields without the same type.
> We keep all your code changes with slightly different field names (like "_translations"), and populate the corresponding already existing Product "_translated" field.

Yes that's something I tried locally before making the PR, but then thought of what I made as being too clumsy.
The PR would probably face even more push back if I pushed that first version...
It turned the old fields into getters, removed annotations, the getters had too sophisticated logic for POD getters, the code looked very confusing - many fields had 3 versions of them (e.g. `String productName`, `String productNameTranslated`, `Map<Lang, String> productNameTranslatedAll`). Clumsy.

> Regarding the fact that "_translated" fields was a recent change and that it would probably be harmless to transform them, I understand your point but I don't have enough hindsight about it.

Do you think maybe we should ask someone who has the hindsight then? Maybe @M123-dev?

monsieurtanuki:
Clever, clever: hey, I'm not that clever! ;)

What I suggest is
1. to use your Map<language, label>, but with a different name (e.g. "_translations") - I won't get a Nobel Price for that idea
1. to deprecate the fields we should not use anymore (e.g. "_translated") (like "_de", "_fr" and "_en" before) (instead of replacing their types) - that's safe
1. to add a method to compute the "best" translation of a field for a `Product` - that's useful and if we don't do it, every developer would need to code it anyway

If you prefer to keep the `Product` code as basic as possible, the "get best translation" method can be put in a distinct `ProductHelper` class. And the deprecated "translated" getters won't be refined code but just plain getters, as the fields will be populated after the json decode with a `ProductHelper.populateBestTranslations(Product, List<Language>)` method - in a first phase. And later those fields will be removed - in a second phase.

I would mind the change of types because I don't think it's the right thing to do, that's typically what `@deprecated` is here for, but I have no moral authority to prevent you from merging your code.

blazern:
> Clever, clever: hey, I'm not that clever! ;)
> 
> What I suggest is
> 
>     1. to use your Map<language, label>, but with a different name (e.g. "_translations") - I won't get a Nobel Price for that idea
> 
>     2. to deprecate the fields we should not use anymore (e.g. "_translated") (like "_de", "_fr" and "_en" before) (instead of replacing their types) - that's safe
> 
>     3. to add a method to compute the "best" translation of a field for a `Product` - that's useful and if we don't do it, every developer would need to code it anyway
> 
> 
> If you prefer to keep the `Product` code as basic as possible, the "get best translation" method can be put in a distinct `ProductHelper` class. And the deprecated "translated" getters won't be refined code but just plain getters, as the fields will be populated after the json decode with a `ProductHelper.populateBestTranslations(Product, List<Language>)` method - in a first phase. And later those fields will be removed - in a second phase.
> 
> I would mind the change of types because I don't think it's the right thing to do, that's typically what `@deprecated` is here for, but I have no moral authority to prevent you from merging your code.

What do you think about changing the fields types and changing the SDK version so that breaking of some of the API would be expected?

I really don't like the idea of keeping the old fields and adding the new ones near.
This will not only make the Product class more clumsy, it will also make Products retrieving (which we've discussed a bit) and saving (which we haven't discussed yet) much more complicated.

This PR has translated fields retrieving as simple as possible. You suggest we add a new function `ProductHelper.populateBestTranslations` function which would do what "every developer would need to code it anyway". No developer has asked for it yet, be should not try to guess every need of every developer. Principles like KISS and YAGNI exist for a reason.

The other thing is saving product back to OFF.
If we keep both fields (_translated and _translations), then which of them should be sent to OFF when a product gets saved?
If only `_translations`, it would break the API without compilation errors.
If both `_translations` and `_translated`, then in case of conflicts (when a product has both fields filled for same language) which of the translations do we send? Do we send neither and throw an exception? This might be a rare case and considering Flutter's nature this exception will most likely be silently consumed in most cases, the users won't even know they do something wrong.

Keeping both fields leads to tons of complications.
So... What do you think about changing the fields types and changing the SDK version so that breaking of some of the API would be expected?

monsieurtanuki:
My point is to use "your" new "_translations" fields as the new reference, that will be impacted by all reads and writes.
The legacy deprecated fields will only reflect what is in the new fields, temporarily, before they're removed.

Developers will need a way to get the "best" translation, or just the only translation, using the "_translations" fields and the desired languages. It can be done if they keep passing the languages to the translation arrays every time, like `productName_translations['en'] ?? productName_translations['ru'] ?? productName_translations['fr'] ?? productName_translations['de'] ?? 'unknown product name'`, or use a more simple field `productName_translated` that was computed when they loaded the product. As a developer I do think I would prefer the second option "hey OFF, I asked the best translations among 3 languages, just give me what I asked for". Which is still compatible with "hey OFF, I asked the translations among 3 languages, give me all of them".

That's my point of view. It looks like we disagree. Good news for you: I'm not the only one there to review and approve your code.

M123-dev:
Very interesting discussions here. Now I'll put my two cents in on this discussion.
First of all, I'm on vacation so I'm writing via the Github app, since it's not exactly pleasant to read the code here I just skimmed it. 

I was thinking a bit about the problem and I have a bit different idea. Why do we even have field and fieldTranslated, I would suggest that we completely move away from fieldTranslated. 

Let's say I as an SDK user am building a small app, in 90% of cases I will only ever display one language so I would say that (although it is a bit of logic and not just a wrapper) we take the most appropriate language. If we assume that the list of languages is ordered by priority it shouldn't be that much effort.  

If there is the case where we need to display more than one language we can add a 

`Map<Language, Product>getMultipleLanguges()` function. 


With this method we don't need hardcoded keys and multiple null checks anymore. If I have forgotten or overlooked anything let me know. I  am looking foreward to your feedback and am open for improvement.

@blazern I think you're recommendation is a good idea for RawProduct.

blazern:
Hello @M123-dev! Thank you for getting involved :)

> @blazern I think you're recommendation is a good idea for RawProduct.

I'm not sure I understand what you mean by `RawProduct` since presently the SDK doesn't have such a class.
There's a `OpenFoodAPIClient.getProductRaw` method, but it does almost same thing as `OpenFoodAPIClient.getProduct` and even return type is the same.
Do you mean we should create a new product class with the name `RawProduct`? Or we should change the return type and logic of `OpenFoodAPIClient.getProductRaw`? Or something else?

> I was thinking a bit about the problem and I have a bit different idea. Why do we even have field and fieldTranslated, I would suggest that we completely move away from fieldTranslated.
> 
> Let's say I as an SDK user am building a small app, in 90% of cases I will only ever display one language so I would say that (although it is a bit of logic and not just a wrapper) we take the most appropriate language. If we assume that the list of languages is ordered by priority it shouldn't be that much effort. 

When I added the `fieldTrasnlated` fields I assumed it was the right way to go since that's how the backend gives its responses - it has fields like `product_name`, `product_name_ru`, `product_name_en`.
Since the backend thinks the translated fields have same position and value as not-translated, I assumed the SDK should treat the fields in the same way.
Here's an example response of the backend: `{"code":"3083680015394","product":{"product_name":"Maïs sans sucres ajoutés","product_name_ru":"Кукуруза сладкая"},"status":1,"status_verbose":"product found"}`

And I still believe that's the right way with current backend responses structure AND with how backend saves/updates products.
- I'd argue that the SDK should not try to figure out appropriate languages for its clients, it should just give the clients what they asked for and nothing more than that. But that's subjective and seems already 2 people think the SDK should provide that ability.
- What's not subjective though, is the fact that backend gives us `product_name` as the *main name of the product* (it's not just a translation, it's the main product's name), then it gives us translations as `product_name_ru`, `product_name_en`, etc.
- If we do remove the `translated` fields and fill `productName` and other field with the translations we deem most appropriate, then the SDK users will loose the ability to understand if the `field` they use is in its main language or if it's a translation. But maybe that's not that big of an issue.
- There's also a little issue with saving - if we remove the `translated` fields then it won't be very straight-forward for the SDK clients to save a product with some fields translations. The client definitelly won't be able to save multiple lang translations with 1 request (unless we do some refactoring of the request-building code). It also will be confusing for the client to understand what kinds of products fields they save - do they update the main language field value? Or a translation? Which translation?

> If there is the case where we need to display more than one language we can add a
> 
> `Map<Language, Product>getMultipleLanguges()` function.
> 
> With this method we don't need hardcoded keys and multiple null checks anymore.

I don't see how such a function would help with nullchecks - if there'll be an instance of `Product` in this map it wouldn't mean that it has translated all of its translateable fields. A product could have `ingredientsText` translated but not `productName`. And also we still would have to check whether the map has a product for a language, which also is a null-check.

So there still would code like `langProductMap[Lang.en]?.productName ?? 'some default name'`.

M123-dev:
> There's a OpenFoodAPIClient.getProductRaw method, but it does almost same thing as OpenFoodAPIClient.getProduct and even return type is the same. 

My bad, I thought it would return something different. In the written documentation in the Product Class overview is a 
`---- Available for RAW products ----` devider, we should probably edit this. 

> I don't see how such a function would help with nullchecks 

I meant if we return a product with only one product field and not all the translated ones, then we only have to check if the field is null and not for every translation.



Putting productFieldTranslated in a map is definitely a very good solution, I understand that some sdk users would like to have the flexibility to have all the data(also it's good for editing and adding).  As it has already been mentioned here, we cannot predict all use cases. My problem is only that I in my opinion it is too confusing what exactly the sdk user should / would like to use. 


Let's say I have a small internal app that only supports the German language.  Either I choose productField. With most german products it is not a problem.  But in the case of an international or foreign product, however, the original product field is used, although there may be is a translated version. 

Or I see there is productFieldTranslated and always use the German translation, but how many German products also have a German translation?


One possible solution I can think of which does not require to build a whole new class is that we add something like a `getBestProductName` which returns the best productName depending on the specified languages.  (Of course the same for all translated data)

We could also rename the existing productName but this would again cause edit or adding problems.

When we do it like that we nearly have no breaking change (we could even allow to pass language or languages so that nothing breaks), we have all the data and have a way to get smart data.

What are your opinions on that @blazern @monsieurtanuki

blazern:
> Or I see there is productFieldTranslated and always use the German translation, but how many German products also have a German translation?

All of the German products have German translation though! :)
And the backend seems to know that.

Take for example this recently added product which has only German language texts and its main language is German: https://de.openfoodfacts.org/produkt/4003240314001/karottensaft
When we request `product_name` and `product_name_de`: https://ru.openfoodfacts.org/api/v0/product/4003240314001.json?fields=product_name,product_name_de
The backend gives us same values for both `product_name` and `product_name_de`: `{"code":"4003240314001","product":{"product_name":"Karottensaft","product_name_de":"Karottensaft"},"status":1,"status_verbose":"product found"}`

So if an app wants to prefer German translations of product's texts, it seems the app should *always* use `product_name_de` instead of `product_name`.

---

If I understood your concern correctly, it seems the "always use product_name_de" approach solves the problem and `getBestProductName` wouldn't be needed at all.
Is that correct?

M123-dev:
It indeed solves the before mentioned problem, and simplified my view about this topics.

When I understood the PR correctly it is about multiple languages, when we expand the problem a bit and I want the german name and when it's not available the russian one, when doing this I would have to manually write both of them, doing this for multiple data sources, multiple times is a pain in the neck.  

I was of the opinion that it would be good to include a simplification of logic in the sdk, but if there is a bit of backend logic behind productFieldTranslated, your approch makes more sense to me. I don't think that so many apps will allow to specifiy more then one language they speak. 

If you update the documentation and probably also explain the difference from the normal and the translated fields in the doc comments I would agree with this pr.

stephanegigandet:
I haven't had time to think a lot about this and what would be the best solution, but I can confirm how the backend works:

For fields that have language specific values (product name, generic name, ingredients_text, packaging_text, and probably a few others):

We store the language specific values in [field_name]_[language_code] e.g. product_name_ru.
Then we copy the value for the "main language" of the product (as set in the "lang" field) to "product_name".

M123-dev:
Thanks @stephanegigandet, so it is also not possible to edit product_name and the like via api, right?

@monsieurtanuki do you have any concerns left or probably some new ideas?

stephanegigandet:
> Thanks @stephanegigandet, so it is also not possible to edit product_name and the like via api, right?

Right. The api allows you to specify that you want to update the "product_name" field, or the  "product_name_ru" field. If you just say you update "product_name", in fact it will add the value of the "lc" field you pass, so if you pass "ru", you will in fact update "product_name_ru". (which will then be copied to "product_name" only if "ru" is the main language of the product).

stephanegigandet:
It's a very complex topic, because products are complex (they can have multiple languages) and the API is complex (we have lots of features to handle the complexity of multilingual products, but also multilingual categories, features etc.).

So first of all, thanks a lot for your work on solving this problem @blazern , and thank you @monsieurtanuki and @M123-dev for all the insightful comments.

For reading, the API is in fact able to return you the product name (or ingredients etc.) in the language you prefer:

https://world.openfoodfacts.org/api/v2/product/3017620422003?fields=ingredients_text,ingredients_text_languages&lc=ru,es --> Prefer Russian, and then Spanish. Fallback to main language of the product.

Here ingredients_text will in fact contain the Spanish ingredients (because we don't have a value for Russian).

There is also the "ingredients_text_languages" field which will give you all languages.

So, for reading, I think it makes sense to keep the existing "product_name" field, but to allow to pass a list of languages, ordered by preference.

It also makes sense to allow to get the actual values for specific languages (including multiple languages). That's especially important for apps that are going to update products: we want them to request product_name_ru and to update product_name_ru. They need to know the value for the specific language they will be editing. We don't want them to request product_name, get a Spanish value, and then write back this value to product_name_ru.

One thing though is that we should not call the field with "translated" or "translations" in it. It's not a translation, it is the actual value that is shown on the product. We especially do not want users or apps to add translations, manual or automated, (e.g. for ingredients) that are not printed on the package.

I think the best thing would be to use the name that we have in the API: e.g. ingredients_text_languages (instead of ingredients_text_translated/translations).

One thing to note: currently, ingredients_text_languages is just a field available when reading products, it doesn't really exist in the database (instead we have the individual ingredients_text_en/ru/fr etc.). So it's not supported for writing, for writing we need to send ingredients_text_en etc.

blazern:
- Added comments to new functions of `LanguageHelper`.
- Reverted deprecation of `AbstractQueryConfiguration.language`, replaced `AbstractQueryConfiguration.languages` with `AbstractQueryConfiguration.secondaryLanguages`. Most SDK users probably don't care about the multilingual support, so 1 language is plenty and it should not be deprecated.
- Renamed the `translated` fields according to the information provided by @stephanegigandet (thanks a lot, btw, you clarified a lot of things). The `translated` postfix is replaced with `InLangs` postfix. At first I wanted to use the `product_name_languages` format as advised by @stephanegigandet, but then I saw that not all of the fields have the `_languages` versions. For example, `product_name_languages` exists, but `countries_tags_languages` doesn't. On the other hand, there are both `product_name_en` and `countries_tags_en`, and all the other translatable fields have the `_lang` option.
- Added a long language-related mechanics explanation to DOCUMENTATION.md. Although my understanding of how Open Food Facts works is limited and the text I wrote can have some not correct info - I'd appreciate it if someone proofread it. The doc is more readable in my branch: https://github.com/blazern/openfoodfacts-dart/blob/translations-update/DOCUMENTATION.md

blazern:
`test/api_searchProducts_test.dart: OpenFoodAPIClient search products search products filter additives` fails but it doesn't seem to be related to my commit (reverted it locally and tried to run the tests again).
What should I do? Disable the test?

monsieurtanuki:
> `test/api_searchProducts_test.dart: OpenFoodAPIClient search products search products filter additives` fails but it doesn't seem to be related to my commit (reverted it locally and tried to run the tests again).
What should I do? Disable the test?

I had the same problem this morning, something about a timeout.
I suggest that you add a `TODO` comment, disable the test and create the corresponding issue.

DOCUMENTATION.md
```diff
@@ -525,3 +516,143 @@ if(result.status != 1) {
...
+the language(s) the user understands.
+In order for a developer to properly implement multilingual Open Food Facts
+usage, the developer might want to understand how Open Food Facts works with
+languages.
+
+Here are main concepts.
+
+Products in Open Food Facts can have some of their fields in multiple
+languages. Such fields are of 2 types:
+1. Fields with info taken directly from product packaging
```
stephanegigandet, line 540:
```suggestion
1. Text fields with info taken directly from product packaging
```

DOCUMENTATION.md
```diff
@@ -525,3 +516,143 @@ if(result.status != 1) {
...
+   (like product name and ingredients list). Such fields generally are
+   not touched by the backend and are stored in the DB exactly as they
+   were sent.
+2. Fields with meta-information (like list countries where the product is
```
stephanegigandet, line 544:
```suggestion
2. Lists of tags with meta-information (like list countries where the product is
```

lib/utils/AbstractQueryConfiguration.dart
```diff
@@ -38,7 +48,16 @@ abstract class AbstractQueryConfiguration {
         (field) => field == ProductField.ALL,
       );
       if (!ignoreFieldsFilter) {
-        final fieldsStrings = convertFieldsToStrings(fields!, language);
+        final languages = secondaryLanguages.toList();
+        if (language != null && !languages.contains(language)) {
+          languages.add(language!);
```
stephanegigandet, line 53: If both language and scondaryLanguages are provided, language should be added at the top of the list I think.

stephanegigandet:
> * Reverted deprecation of `AbstractQueryConfiguration.language`, replaced `AbstractQueryConfiguration.languages` with `AbstractQueryConfiguration.secondaryLanguages`. Most SDK users probably don't care about the multilingual support, so 1 language is plenty and it should not be deprecated.

I think we should keep the name "languages" and not secondaryLanguages. As you showed in the examples, for multilingual apps, they will probably just specify "languages" without "language", so it's not "secondary".

It's a bit unclear what happens when both language and languages are specified. Maybe we could just decide to thrown an error if we get both. That way there is no chance for surprises.

I think I'm the one who added lc as a convenience (for myself, just passing language codes seems more clearer to me, but that's just a question of habit). Maybe we could deprecate lc in order to have one less field.

Thanks a lot for adding the long documentation with examples, I think it helps a lot.

lib/utils/AbstractQueryConfiguration.dart
stephanegigandet: Or on second thought, I think we should disallow it and throw an error.

monsieurtanuki:
Maybe it's not relevant, but what about deprecating `language` and `lc` altogether, using `languages` only, with a special process when `(languages.length==1)`?
If we do that, there's no confusion - just one parameter about languages - and it's rather fair for "experts" that use several languages (not the comme case) to expect a different behavior for 1 and for n languages.

blazern:
> Maybe it's not relevant, but what about deprecating `language` and `lc` altogether, using `languages` only, with a special process when `(languages.length==1)`?
> If we do that, there's no confusion - just one parameter about languages - and it's rather fair for "experts" that use several languages (not the comme case) to expect a different behavior for 1 and for n languages.

We probably can't do that.
At least not without another long discussion and I don't feel like I can handle another one :(

The `language` and `lc` fields are used to set the `lc` query parameter: https://world.openfoodfacts.org/api/v2/product/3017620422003?fields=ingredients_text,ingredients_text_en&lc=ru (the last param, `lc=ru`).
The `lc` query parameter and the `_lang` postfixes (e.g. `_ru` or `_de`) affect query execution in different ways:
- `lc` makes fields like `product_name` to be preferably filled with the provided language, or rollback to the main language.
- `_lang` postfixes explicitly specify what language we want, without any rollbacks.
Some SDK users will want one thing, some the other, and some will want both.

We could use the first element of `languages` to specify the value of `lc` in queries to the backend, but it probably is a bad idea since it's a very-very implicit behavior.

lib/utils/AbstractQueryConfiguration.dart
blazern: Are you sure?
The 2 fields regulate different behavior - the `language` field gives a value to the `lc` query-param, and `languages`/`secondaryLanguages` give values to lang-postfix query params (e.g. `product_name_en`).
`lc` and postfixes have different behaviors: https://github.com/openfoodfacts/openfoodfacts-dart/pull/185#issuecomment-882654523

monsieurtanuki:
@blazern I see. Maybe `preferredLanguage` and `languages` then. Just a suggestion.

blazern:
> @blazern I see. Maybe `preferredLanguage` and `languages` then. Just a suggestion.

I like it!
But I think we definitely need approves of both @stephanegigandet and @M123-dev for that, because probably 90% of project which use the OFF Dart SDK also use the `language` property and it will break compilation for them.

monsieurtanuki:
@blazern Fair enough. Then, maybe `language` and `additionalLanguages` (or `extraLanguages`).

M123-dev:
> Then, maybe language and additionalLanguages (or extraLanguages).

I think that would be the easiest to understand and it does not break a lot. Let's go with that.

blazern:
> > Then, maybe language and additionalLanguages (or extraLanguages).
> 
> I think that would be the easiest to understand and it does not break a lot. Let's go with that.

Okay

blazern:
- Renamed `AbstractQueryConfiguration.languages` -> `AbstractQueryConfiguration.extraLanguages`.

Although @stephanegigandet might disagree with this: https://github.com/openfoodfacts/openfoodfacts-dart/pull/185#discussion_r672346540

stephanegigandet:
Hi @blazern,

> The 2 fields regulate different behavior - the language field gives a value to the lc query-param, and languages/secondaryLanguages give values to lang-postfix query params (e.g. product_name_en).
> lc and postfixes have different behaviors

I have a proposal to change that. To me, "language" and "languages" should not have different behaviors. They should both indicate the order of preferred language(s) for fields like productName and ingredientsText + indicate the desired languages for productNameInLang or categoriesInLang.

It's just that if you want to set only one language, you use language=>OpenFoodFactsLanguage.RUSSIAN, and if you want to set multiple languages, you use languages => [[
            OpenFoodFactsLanguage.ENGLISH,
            OpenFoodFactsLanguage.RUSSIAN,
            OpenFoodFactsLanguage.GERMAN
          ],

It's a bit like setting child: or children: properties in widgets: the name makes it very clear if it's only one or multiple.

That way we don't break the existing behavior for apps that currently request "language".

Then depending on whether the app wants the best match for its choice of languages, or the actual values for each of the languages, it then asks for field ingredientsText or for ingredientsTextInLangs (or possibly both: it makes sense to use ingredientsText for displaying a product, and to also get ingredientsTextInLangs to then offer to update it).

I think it's simpler / cleaner that way. We separate setting requested languages from the fields being requested, and setting one or more languages that not change the behavior.

It's also similar to the way the API works today: in the lc field, you can pass one language lc=ru), or multiple (lc=ru,en,de).

stephanegigandet:
Small comment on fields names:

I think it may be best to rename ingredientsTextInLangs to ingredientsTextInLanguages , in order to avoid having fields with "Language" and other with "Lang".

I was reading the Flutter style recommendations, and they recommend not to use abbreviations: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#avoid-abbreviations

stephanegigandet:
So for instance, we would support queries like:

      ProductQueryConfiguration conf = ProductQueryConfiguration(barcode,
          languages: OpenFoodFactsLanguage.ENGLISH,
          fields: [ProductField.INGREDIENTS_TEXT_IN_LANGS]);

blazern:
Okay, so I have to:
- remove the `lc` and `language` fields of `AbstractQueryConfiguration`,
- rename `extraLanguages` to `languages`,
- make the `lc` query param in the SDK to accept a list of languages.
Is that correct?

@M123-dev, @monsieurtanuki, what do you think?

stephanegigandet:
> * remove the `lc` and `language` fields of `AbstractQueryConfiguration`,

We could keep "language" in order not to break existing apps.
passing language : OpenFoodFactsLanguage.ENGLISH  would be exactly the same as passing languages : [ OpenFoodFactsLanguage.ENGLISH ]

We should forbid setting both "language" and "languages" though as it would not be obvious which would have priority.

> * rename `extraLanguages` to `languages`,

yes

> * make the `lc` HTTP-query param in the SDK to accept a list of languages.

lc should be a comma separated list of language codes corresponding to "languages" (or just 1 language code if we have "language")

monsieurtanuki:
If I sum up, about `AbstractQueryConfiguration`:
1. I think that everybody agrees to keep the parameter `OpenFoodFactsLanguage? language`, in order not to break usability.
1. And the purpose of the current PR is to add the parameter `List<OpenFoodFactsLanguage>? languages`.
1. I think we've got enough language parameters, so we can deprecate the parameter `String lc`.
2. Both `language` and `languages` cannot be not null at the same time - all other combinations are ok: (null,null), (null, not null), (not null, null)
1. In all cases, that will be transformed into the API URL as something like `&lc=fr,de,ru` in `getParametersMap()`.
1. And in `ProductFields`
```dart
List<String> convertFieldsToStrings(
  List<ProductField> fields,
  OpenFoodFactsLanguage? language)
```
will probably become something like
```dart
List<String> convertFieldsToStrings(
  List<ProductField> fields,
  List<OpenFoodFactsLanguage> languages)
```

blazern:
- `extraLanguages` -> `languages`
- `in langs` -> `in languages`
- multiple values for the `lc` query param are supported and a corresponding test is added

stephanegigandet (approved):
Looks good to me, thanks a lot @blazern !

M123-dev (approved):
Thank you @blazern , looks good to me too.  I'm used to small discussions here, but how it turned out here amazes me.  I hope it doesn't stop you from contributing in the future.

lib/utils/ProductListQueryConfiguration.dart
```diff
@@ -6,15 +6,19 @@ import 'package:openfoodfacts/utils/ProductFields.dart';
 class ProductListQueryConfiguration extends AbstractQueryConfiguration {
   final List<String> barcodes;
 
+  /// See [AbstractQueryConfiguration.languages] for
+  /// parameter's description.
   ProductListQueryConfiguration(
     this.barcodes, {
     final OpenFoodFactsLanguage? language,
+    final List<OpenFoodFactsLanguage> languages = const [],
```
monsieurtanuki, line 14: Unfortunately I don't think we can deprecate `lc` from there.

monsieurtanuki (approved):
@blazern Looks good to me!

Still, correct me if I'm wrong, it's a breaking change regarding the names of the localized product fields.
And, as developers that used the `language` parameter probably used the localized product fields (that have just been renamed), perhaps it would have been a good opportunity to deprecate the `language` parameter as well and only use the new `languages`. Developers, forced to rename their fields, would also have to rephrase `language` into `languages`, but they needed to edit their code anyway.
It's just an option, if you don't agree just dismiss my comment - especially if my assumptions are wrong ;)

lib/utils/ProductListQueryConfiguration.dart
blazern: I don't understand. Could you please elaborate?

blazern:
> Still, correct me if I'm wrong, it's a breaking change regarding the names of the localized product fields.
> And, as developers that used the `language` parameter probably used the localized product fields (that have just been renamed), perhaps it would have been a good opportunity to deprecate the `language` parameter as well and only use the new `languages`. Developers, forced to rename their fields, would also have to rephrase `language` into `languages`, but they needed to edit their code anyway.
> It's just an option, if you don't agree just dismiss my comment - especially if my assumptions are wrong ;)

Yesterday you've said a [different thing](https://github.com/openfoodfacts/openfoodfacts-dart/pull/185#issuecomment-883388989) though:
> I think that everybody agrees to keep the parameter OpenFoodFactsLanguage? language, in order not to break usability.

Stephane also [have said](https://github.com/openfoodfacts/openfoodfacts-dart/pull/185#issuecomment-883333788) we should keep the old functionality for the `language` param:
> We could keep "language" in order not to break existing apps.
> passing language : OpenFoodFactsLanguage.ENGLISH would be exactly the same as passing languages : [ OpenFoodFactsLanguage.ENGLISH ]
> 
> We should forbid setting both "language" and "languages" though as it would not be obvious which would have priority.

monsieurtanuki:
@blazern I do know I said that yesterday. It's just that reviewing your code this afternoon, I've seen that localized product fields were renamed. If all of them are renamed, then my remark makes sense: all developers with localized fields will be impacted anyway. But if some localized fields are not renamed, like `productName` (I haven't double-checked that), then my remark is not that relevant, because some developers won't be impacted. And there's no reason to impact them.

lib/utils/ProductListQueryConfiguration.dart
monsieurtanuki: I guess I should have selected the line below, with `String lc`.
`lc` is deprecated in the abstract class but I don't think there's a way to tag it as deprecated in the child classes as well. Perhaps dart/flutter does it automagically: "hey you're using a parameter that is used as is in the parent class and is deprecated there"?

blazern:
> @blazern I do know I said that yesterday. It's just that reviewing your code this afternoon, I've seen that localized product fields were renamed. If all of them are renamed, then my remark makes sense: all developers with localized fields will be impacted anyway. But if some localized fields are not renamed, like `productName` (I haven't double-checked that), then my remark is not that relevant, because some developers won't be impacted. And there's no reason to impact them.

The `productName` and `ingredientsList` fields are not affected - SDK users who used the `language` param before had these fields filled with a desired language, and this behavior will stay the same way after the changes from this PR are applied.

Can we please merge this PR already?
I do understand the desire to make things perfect, to remove every single extra thing that is not needed, to propagate the `@deprecated` annotation everywhere where it's appropriate and to document every single public member.
But this PR is being worked on for a week already. I can understand a push back when a pull request has real flaws which would definitely affect the SDK users. But dancing around little things which are not agreed upon only because different people like things in different ways - I don't understand that, and it exhausts me really much.

monsieurtanuki:
@blazern Thank you for your answer, and thank you for your code and your patience.
The topic was important and interesting enough to be reviewed, commented and approved by 3 reviewers, which doesn't happen often.

blazern:
Sorry for being a bit toxic in the end - I expected the patch to be a quick 20 minutes adventure when I started. :)
I do understand that the topic is complex and everybody wanted the patch to not have flaws so that nothing would need to be fixed in the future, and that is the reason why it took it so long to merge.

PR #187, opened 2021-07-16
fix/#48 - now saving nutrients

monsieurtanuki:
Impacted files:
* `api_saveProduct_test.dart`: new test "confirm that nutrient fields are saved"
* `openfoodfacts.dart`: fixed the way nutrient API parameters are set


M123-dev (approved):
Thanks @monsieurtanuki

PR #197, opened 2021-07-21
feature/#194 - explicit parameter WithoutAdditives replaces ContainsAdditives

monsieurtanuki:
New file:
* `WithoutAdditives.dart`: "Without Additives" search API parameter

Impacted files:
* `api_searchProducts_test.dart`: replaced `ContainsAdditives` with `WithoutAdditives`
* `ContainsAdditives.dart`: deprecated in favor of new class `WithoutAdditives`

### DIFF
--- a/lib/model/parameter/ContainsAdditives.dart
+++ b/lib/model/parameter/ContainsAdditives.dart
@@ -1,6 +1,8 @@
 import 'package:openfoodfacts/interface/Parameter.dart';
 
 /// "Contains additives?" search API parameter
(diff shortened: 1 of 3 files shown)


MohamedFBoussaid (commented):
@monsieurtanuki I know the code was merged. 
But can we use @Deprecated('Message') instead ? It will display the 'Message' in the IDE warning message to make it easy using the SDK, instead of going to the code and read the comments. https://api.dart.dev/stable/2.13.4/dart-core/Deprecated-class.html

Issue #201, opened 2021-07-22
The new product_name / ingredients_text / categories "in_languages" field break Product.fromJson()


stephanegigandet:
In Smoothie we save the products as Json and then load them back, but it now fails with version 1.3.0 with the new in languages fields from https://github.com/openfoodfacts/openfoodfacts-dart/pull/185

Discussion in https://github.com/openfoodfacts/smooth-app/pull/442#issuecomment-884819542

I will add a test case in openfoodfacts-dart that exhibits the problem, and @monsieurtanuki said he's working on a solution.

monsieurtanuki:
I don't think there should be a breaking change: the format of what we get from the server did not change, did it?
Or is it rather the fields that we're asking for that are different?

stephanegigandet:
That's correct, the JSON sent back by the server is exactly the same as before.

But the Dart package's toJson produces a different JSON than what the server returns.

Instead of having product_name_en, product_name_fr etc. they are combined in product_name_in_languages:

{"code":null,"product_name_in_languages":{"en":"Quaker Oats Old Fashioned Oatmeal 18 Ounce Paper Cannister"},"lang":"-","selected_images":{},"images":{},"ingredients_text_in_languages":{"en":"whole grain rolled oats,"},"ingredients_analysis_tags":[],"nutriments":{},"additives_tags":[],"allergens_tags":[],"nutrient_levels":{},"categories_tags_in_languages":{"en":["Plant-based foods and beverages","Plant-based foods","Cereals and potatoes","Cereals and their products"],"fr":["Aliments et boissons à base de végétaux","Aliments d'origine végétale","Céréales et pommes de terre","Céréales et dérivés"]}}

monsieurtanuki:
@stephanegigandet That's the problem, then. It should be the same json format we get from the server and we send to the server. Am I right?

stephanegigandet:
> @stephanegigandet That's the problem, then. It should be the same json format we get from the server and we send to the server. Am I right?

Right, we currently have a Product.toServerData() function that changes the Product.toJson output to change back the keys to the format used by the server:

```
  Map<String, String> toServerData() {
    final json = toJson();
[..]
```

I'm thinking we could do the reverse, we change Product.toJson() to add directly in it the extra processing that is currently in toServerData().

monsieurtanuki:
@stephanegigandet The thing is that we cannot do exactly what we want because of JsonAnnotation's generated code.
[...]

PR #202, opened 2021-07-22
Test to reproduce fromJson bug #201

stephanegigandet:
Test to reproduce #201

Unfortunately no solution yet.

### DIFF
--- a/lib/model/Product.g.dart
+++ b/lib/model/Product.g.dart
@@ -10,8 +10,8 @@ Product _$ProductFromJson(Map<String, dynamic> json) {
   return Product(
     barcode: json['code'] as String?,
     productName: json['product_name'] as String?,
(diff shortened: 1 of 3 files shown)


monsieurtanuki (commented):
@stephanegigandet I would recommend to comment all the `languages` fields until we find a solution.

monsieurtanuki:
@stephanegigandet I've just fixed it. Please have a look at my code, and merge if relevant.

stephanegigandet:
> @stephanegigandet I've just fixed it. Please have a look at my code, and merge if relevant.

I can't approve my own PR but your fix looks good to me and it makes the test pass. :)

monsieurtanuki:
@stephanegigandet You cannot approve your own code but you can merge it all the same.
Assuming it makes sense compared to #203.

PR #203, opened 2021-07-22
Make toJson use the same format as the server for in_languages fields

### LINKED ISSUE #201: The new product_name / ingredients_text / categories "in_languages" field break Product.fromJson()

stephanegigandet:
fixes #201 in a different way (complementary I think) than https://github.com/openfoodfacts/openfoodfacts-dart/pull/202/commits/05da4bb66e114b60a29b3690c66251c08b2e3436

### DIFF
--- a/lib/model/Product.dart
+++ b/lib/model/Product.dart
@@ -359,37 +359,6 @@ class Product extends JsonObject {
 
   Map<String, String> toServerData() {
     final json = toJson();
-
-// Defensive keys copy to modify map while iterating
-    final keys = json.keys.toList();
-
-    for (final key in keys) {
-      // NOTE: '_tags_in_languages' are not supported because tags translation
-      // is done automatically on the server.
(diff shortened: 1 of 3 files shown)


monsieurtanuki (approved):
Honestly I'm a bit lost now with all this ping-pong about languages and json.

That said:
* I did not see anything wrong in this review
* apparently it solves the issue
* if the question is: "would you be angry if this PR was merged and not #202, the answer is that I won't be angry"
* if it makes sense to merge both, why not

PR #207, opened 2021-07-23
Fix issue for fromJson with categories_tags_in_languages + tests

stephanegigandet:
one more test for #201 (currently failing).

The part that makes it fail is the tags_in_languages fields:

"categories_tags_in_languages":{"fr":["Aliments et boissons à base de végétaux","Aliments d\'origine végétale","Céréales et pommes de terre","Aliments à base de fruits et de légumes","Céréales et dérivés","Plats préparés","Sandwichs","Substituts de viande","Plats préparés d\'origine végétale","Céréales préparées","Steaks végétaux","Plats à base de boulgour","Plats préparés à faible teneur en matières grasses","Hamburgers végétariens","Galettes de céréales aux légumes","Plats-prepares-au-fromage"]},

"labels_tags":["en:organic","en:vegetarian","en:eu-organic","en:non-eu-agriculture","en:es-eco-019-ct","en:eu-agriculture","en:eu-non-eu-agriculture","en:fr-bio-01","en:high-fibres","en:made-in-france","en:no-additives","en:no-colorings","en:no-preservatives","en:nutriscore","en:nutriscore-grade-a","fr:ab-agriculture-biologique","en:es-eco-019"],"labels_tags_in_languages":{"fr":["Bio","Végétarien","Bio européen","Agriculture non UE","ES-ECO-019-CT","Agriculture UE","Agriculture UE/Non UE","FR-BIO-01","Riche en fibres","Fabriqué en France","Sans additifs","Sans colorants","Sans conservateurs","Nutriscore","Nutriscore A","AB Agriculture Biologique","en:es-eco-019"]},


stephanegigandet:
01:00 +21 ~6 -1: test/product_from_json_test.dart: Load product from JSON - multilingual categories [E]                                                                                                
  type 'List<dynamic>' is not a subtype of type 'List<String>' in type cast
  package:openfoodfacts/utils/LanguageHelper.dart 451:56  LanguageHelper.fromJsonStringsListMap
  package:openfoodfacts/model/Product.g.dart 71:47        _$ProductFromJson
  package:openfoodfacts/model/Product.dart 304:28         new Product.fromJson
  test/product_from_json_test.dart 42:31                  main.<fn>
  test/product_from_json_test.dart 36:60                  main.<fn>

monsieurtanuki:
@stephanegigandet I think the problem is that you populate fields `categories_tags_in_languages` and you decode them in `Product.fromJson` in the following test: `} else if (key.startsWith('categories_tags_')) {`.
In one case (`categories_tags_in_languages`), you expect something like a `Map<String, List<String>>`.
In the other case ('categories_tags_fr`), you expect directly a `List<String>`.
Am I right?

stephanegigandet:
> @stephanegigandet I think the problem is that you populate fields `categories_tags_in_languages` and you decode them in `Product.fromJson` in the following test: `} else if (key.startsWith('categories_tags_')) {`.

It doesn't reach this point in fact, it fails just before, in this call:
final Product result = _$ProductFromJson(json);

_ProductFromJson(json)
package:openfoodfacts/utils/LanguageHelper.dart 451:56  LanguageHelper.fromJsonStringsListMap
package:openfoodfacts/model/Product.g.dart 71:47        _$ProductFromJson
package:openfoodfacts/model/Product.dart 305:28         new Product.fromJson
test/product_from_json_test.dart 42:31                  main.<fn>
test/product_from_json_test.dart 36:60                  main.<fn>

type 'List<dynamic>' is not a subtype of type 'List<String>' in type cast

stephanegigandet:
So this:

result[LanguageHelper.fromJson(key)] = map[key]!.cast<String>();

Seems to solve the first issue I had, but you are right, I do have a further issue with: 

> @stephanegigandet I think the problem is that you populate fields `categories_tags_in_languages` and you decode them in `Product.fromJson` in the following test: `} else if (key.startsWith('categories_tags_')) {`.

I'll skip it for the "_in_languages" fields.

Tests are passing now, which doesn't mean everything works though. :)

test/product_from_json_test.dart
```diff
@@ -32,4 +32,35 @@ void main() {
         json.decode(encodedJson) as Map<String, dynamic>;
     Product product2 = Product.fromJson(decodedJson);
   });
+
+  test('Load product from JSON - multilingual categories', () async {
+    String encodedJson =
```
monsieurtanuki, line 37: I've tried to put the json data in an asset file but I have spent too much time on it without success; it would have been great, though.

lib/model/Product.dart
```diff
@@ -305,45 +305,52 @@ class Product extends JsonObject {
...
-      } else if (key.startsWith('ingredients_text_')) {
-        OpenFoodFactsLanguage lang = _langFrom(key, 'ingredients_text_');
-        if (lang != OpenFoodFactsLanguage.UNDEFINED) {
-          result.ingredientsTextInLanguages ??= {};
-          result.ingredientsTextInLanguages![lang] = json[key];
+      } else {
+        // If the key ends with _[2 characters], it might be a language specific
+        // field.
+        RegExp regexp = RegExp(r'_(\w\w)$');
+        if (regexp.hasMatch((key))) {
```
monsieurtanuki, line 312: I'm not fluent in regexp, therefore I would have coded that differently. There are explicit comments, fair enough.

I still think that you could get rid of your regexp just by saying:
* I compute the language
* if the language is unknown, I do nothing
* else I do something, which includes using the problematic `_jsonValueToList`

Probably would be easier to read. What do you think of that?

monsieurtanuki (changes requested):
Looks good, but I'm not a big fan of regexp.
Please have a look at my comment about it: I think the code would be easier to read and to understand.

lib/model/Product.dart
stephanegigandet: Ah, I use regexps a lot (Perl is my favorite language) so it didn't strike me as bad, but there are other ways certainly.

I'm not sure I understand your suggestion, you want to move the check in each of the conditions?

```
else if (key.startsWith('categories_tags_')) {
            OpenFoodFactsLanguage lang = _langFrom(key, 'categories_tags_');
            final values = _jsonValueToList(json[key]);
            if (lang != OpenFoodFactsLanguage.UNDEFINED && values != null) {
              result.categoriesTagsInLanguages ??= {};
              result.categoriesTagsInLanguages![lang] = values;
            }
          }
```

So something like:

```
else if (key.startsWith('categories_tags_')) {
            OpenFoodFactsLanguage lang = _langFrom(key, 'categories_tags_');
             if (lang != OpenFoodFactsLanguage.UNDEFINED) {
               final values = _jsonValueToList(json[key]);
                if (values != null) {
                 result.categoriesTagsInLanguages ??= {};
                 result.categoriesTagsInLanguages![lang] = values;
               }
            }
          }
```

? (repeated for labels_tags, countries_tags etc.)

It's confusing I think: it would sound as if we somehow get something like  category_tags_zx and zx is not recognized as a language.

Here what we want is to make sure not to match categories_tags_in_languages. I guess we could explicitly exclude it:

else if (key.startsWith('categories_tags_') && key != 'categories_tags_in_languages') {

Then it would be more clear why we have this condition.

monsieurtanuki: Correct, you've understood my point: as anyway we're checking if the language is "defined" (I did not invent it, it's already coded), we can skip the cryptic regexp (at least cryptic enough that you needed to comment it) by checking the language as early as possible.

I agree with the last part of your comment too: it could be more readable to explicitly say "it's not just about wrong language codes, it's also about `_in_languages`".
That would be something like that:
```dart
else if (key.startsWith('categories_tags_')) {
  if (key != 'categories_tags_in_languages') { // more explicit indeed
    final OpenFoodFactsLanguage lang = _langFrom(key, 'categories_tags_');
    if (lang != OpenFoodFactsLanguage.UNDEFINED) {
      final values = _jsonValueToList(json[key]);
      if (values != null) {
        result.categoriesTagsInLanguages ??= {};
        result.categoriesTagsInLanguages![lang] = values;
      }
    }
  }
}
```
or perhaps better
```dart
if (key.contains('debug')) {
   continue;
}
if (key.endsWith('_in_languages')) {
   continue;
}
if (key.startsWith('categories_tags_')) {
  final OpenFoodFactsLanguage lang = _langFrom(key, 'categories_tags_');
  if (lang != OpenFoodFactsLanguage.UNDEFINED) {
    final values = _jsonValueToList(json[key]); 
    if (values != null) {
      result.categoriesTagsInLanguages ??= {};
      result.categoriesTagsInLanguages![lang] = values;
  }
}
```

PS: for the record I hate Perl :)

stephanegigandet: Sounds good, I replaced the regexp with:

```
if (key.endsWith('_in_languages')) {
   continue;
}
```

and added a few comments to explain why we need to do this.

monsieurtanuki (approved):
Perfect @stephanegigandet!

PR #213, opened 2021-08-02
Added User-Agent HTTP Header

### LINKED ISSUE #208: Send user agent HTTP header without User parameter
Hello there!

I would like to conform to: https://wiki.openfoodfacts.org/API/Read and include my App name, version, etc. in the HTTP header when firing request against the API with this dart plugin.

However, I could not find any way to do this. I could only find this pull request but that indicates that one would have to include a user to the request: https://github.com/openfoodfacts/openfoodfacts-dart/pull/18

However, I would like to let the users of my app querying OFF without an account and in the future let them create unique accounts to add or change data. The current implementation on the other hand would let me think that I would have to create a global user for my app or am I mistaken?

Kind regards, Christian

M123-dev:
Fixes #208 

- Added a new `UserAgent` class
- Added a new `toValueString` method to `JsonObject `
- Added `UserAgent` as optional parameter to all get requests


lib/interface/JsonObject.dart
```diff
@@ -59,4 +59,19 @@ abstract class JsonObject {
...
   }
+
+  /// Returns all values as a String separated by a hyphen
+  /// value1 - value2 - value3
+  String toValueString() {
+    return toValueStringStatic(toJson());
+  }
+
+  static String toValueStringStatic(Map<String, dynamic> json) {
+    String result = '';
```
monsieurtanuki, line 70: Please add `///` comment.

lib/model/UserAgent.dart
```diff
@@ -0,0 +1,26 @@
+import 'package:json_annotation/json_annotation.dart';
+import '../interface/JsonObject.dart';
+
+part 'UserAgent.g.dart';
+
+@JsonSerializable()
```
monsieurtanuki, line 6: I do not see the use case that justifies making it `@JsonSerializable`.

monsieurtanuki (commented):
@M123-dev Please have a look at my local comments.
In addition to that, I'm wondering if it makes sense to add `UserAgent` in all methods. I mean, we would perhaps be better off with setting a `static UserAgent` field and using it again and again.
Unless it makes sense to use different `UserAgent`s all along the app. And an _optional_ `UserAgent` makes it easy to forget it once in a while.

M123-dev:
Thanks for your review @monsieurtanuki, indeed it's probably better to do it that way. I updated the code.

monsieurtanuki (approved):
Thank you @M123-dev for editing your code!
I prefer it that way, but actually my previous comment was more an open question than a change request.

If we follow "my" suggestion (that you've just implemented in the latest commit), the idea is that somewhere in `main.dart` the developer would call once something like `OpenFoodFacts.userAgent = UserAgent(name: 'my wonderful app', version: '1.0');`, and this is it, from now on all the OFF http requests will use that user agent, is that correct?

@teolemon @christianfl Does that look good to you, too, or is there a use case that is not covered?

M123-dev:
@monsieurtanuki, yeah that is correct. Currently we pass it to the HttpHelper in every time, this probably can be done better later,

Also its currently only in the get requests, @teolemon what about adding it to post and multipart as well. 
Another thing I noticed while browsing the code, sometimes a user is requiered eventhough it is not needed. (at least from the documentation)  

[addProductImage](https://github.com/openfoodfacts/openfoodfacts-dart/blob/master/lib/openfoodfacts.dart#L109)
[getProductInsights](https://github.com/openfoodfacts/openfoodfacts-dart/blob/master/lib/openfoodfacts.dart#L306)

christianfl:
Wow, that was really fast, thanks for that fix. Observing the code without testing it yet, it seems to do the job fine for the cases I had in mind @monsieurtanuki !

monsieurtanuki:
@M123-dev I guess we're OK with this one, aren't we?

lib/interface/JsonObject.dart
```diff
@@ -59,4 +59,21 @@ abstract class JsonObject {
...
+
+  /// Returns all values as a String separated by a hyphen
+  /// value1 - value2 - value3
+  String toValueString() {
+    return toValueStringStatic(toJson());
+  }
+
+  /// Returns all values as a String separated by a hyphen
+  /// value1 - value2 - value3
+  static String toValueStringStatic(Map<String, dynamic> json) {
```
jasmeet0817, line 71: I would just call it toValueString, static is an identifier, usually we don't put identifiers in method names.

lib/interface/JsonObject.dart
```diff
@@ -59,4 +59,21 @@ abstract class JsonObject {
     });
     return result;
   }
+
+  /// Returns all values as a String separated by a hyphen
+  /// value1 - value2 - value3
+  String toValueString() {
```
jasmeet0817, line 65: is this proxy method needed? The callers can just call JsonObject.toValueString(json) ?

jasmeet0817:
Just a couple nits I found, you don't have to change anything since the code is already merged but I thought I could still add the comments for later.

lib/interface/JsonObject.dart
M123-dev: Thanks for your review @jasmeet0817, I just added it that way because its implemented the same for `toData` a few lines above, but I have no problem if we update it later on.

Issue #193, opened 2021-07-20
Add a Nutri-Score helper


teolemon:
Add a Nutri-Score helper to help apps get to the Nutri-Score instantly for their users

![image](https://user-images.githubusercontent.com/1689815/126289174-1217cbdb-59b7-447b-acb3-dbab67877859.png)

https://docs.google.com/document/d/1_Y3tdgB8w3VkL6tXgjzPVmkmFiXsBfy0UfS6Mg3MkLs/edit#heading=h.ftzz6kpf06gw

monsieurtanuki:
Product improvements are now computed in `product.getProductImprovements()` (#217).
Beyond the `enum` values, I don't know how the translations are supposed to be integrated, and how.

monsieurtanuki:
@teolemon Should we provide messages and translations, or are we done now thanks to `enum ProductImprovement` and `product.getProductImprovements()`?
Same question for #191.

teolemon:
Translations can be done in smoothie @monsieurtanuki

monsieurtanuki:
> Translations can be done in smoothie @monsieurtanuki

Then we can close this issue, can't we? And we should also ping both https://github.com/openfoodfacts/smooth-app/issues/518 and https://github.com/openfoodfacts/smooth-app/issues/519, shouldn't we?

teolemon:
yes, it's fixed by the actions API

PR #217, opened 2021-08-20
feature/#191 - added and computed possible product improvements

monsieurtanuki:
Impacted files:
* `api_getProduct_test.dart`: added checks for `getProductImprovements`
* `Product.dart`: create `enum ProductImprovement` and `Product.getProductImprovements()`

### DIFF
--- a/lib/model/Product.dart
+++ b/lib/model/Product.dart
@@ -19,14 +19,6 @@ part 'Product.g.dart';
 
 /// Possible improvements on a [Product] given its current data
 enum ProductImprovement {
-  /// * Possible message:
-  /// “The Eco-Score takes into account environmental labels. Please take them
-  /// into photo or edit the product so that they can be taken into account.”
-  /// * Asking your users for a photo should be enough.
-  /// * You can otherwise add toggles for explicit labels (please add a photo
-  /// of them to avoid mistakes)
-  LABELS_TO_BE_COMPLETED,
-
   /// Possible message:
   /// “The Eco-Score takes into account the origins of the ingredients.
   /// Please take them into a photo (ingredient list and/or any geographic claim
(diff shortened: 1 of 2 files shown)


teolemon:
Is the scope the implementation of the Open Food Facts states ?

lib/model/Product.dart
```diff
@@ -461,4 +479,23 @@ class Product extends JsonObject {
...
+
+  /// Returns all the potential improvements given the quality of the data
+  ///
+  /// For apps with contribution mode.
+  /// A typical use-case is to alert the end-users that they can improve
+  /// the quality of the OFF data by taking pictures (or something like that),
+  /// when displaying a [Product].
+  Set<ProductImprovement> getProductImprovements() {
+    final Set<ProductImprovement> result = {};
+    if (statesTags != null) {
```
peterwvj, line 491: Minor but I'd avoid nested if structures like this, if possible. You could change it by adding something like
```
if (statestags == null) {
    return result;
}
```

monsieurtanuki: Thank you @peterwvj for your comment.
I agree with you, I'm a big fan myself of `return`, `break` or `continue` statements instead of `else`s (nested or not).
The thing is that I also had in mind #193, that will impact the very same method.
When I implement #193 (next week), I'll take your remark into account if still relevant.

monsieurtanuki:
> Is the scope the implementation of the Open Food Facts states ?

@teolemon If the question is for me, could you please elaborate on it?

teolemon:
So far, the PR is implementing some of the available states (https://world.openfoodfacts.org/states, https://github.com/openfoodfacts/openfoodfacts-server/blob/main/taxonomies/states.txt)
So perhaps I would name them that way ?

peterwvj (approved):

M123-dev (approved):
Thanks @monsieurtanuki

Issue #216, opened 2021-08-18
Allow clients to pass the country as a parameter request while getting a product's details


jasmeet0817:
Currently openfoodfacts.dart queries "world.openfoodfacts.org" directly. See here: https://github.com/openfoodfacts/openfoodfacts-dart/blob/master/lib/openfoodfacts.dart#L66

The functionality of the app is quite different depending upon the country from where you are accessing it (specially when it comes to eco score details card display). So this should be configurable in the package. 

For testing, we should have a way of modifying the host url locally, to be able to test things without having to change the code.

M123-dev:
I am going to add some kind of settings, which should be set once and used everywhere similar to the  User-Agent HTTP Header (#213)
Including:
- All currently hardcoded url hosts
- The QueryType
- A way to overwrite the cc field for debug purposes (to be able to see the eco score from other countries)

Should the User-Agent HTTP Header or maybe something other added as well

monsieurtanuki:
@M123-dev I was thinking about `static` fields again, indeed like #213.
It should be the same for "country": we should not presume what the country is, but instead let the developer set the correct value (in a `static` field). If "country" is not set, it's going to be "https://world..." as it is now.

jasmeet0817:
Are there a set of flags that can be overridden by developers locally? Or static fields that the developer could change in order to have different functional behavior in the app ? Usually we use flags for these things, but I guess static fields are fine for now. 

I would encourage adding all fields together in a file (settings.dart?) that a developer can modify to their liking, and clearly document each field and respective app behavior on it's modification. This is to ensure new developers are able to quickly find and modify settings.dart without having to look deep into the code.

stephanegigandet:
Sorry I was on vacations and did not notice this issue before.

API queries should always go to the "world" subdomain, and language and country should be passed in the "cc" (country code) and "lc" (language code) parameter. Queries to subdomains [country code]-[language code].openfoodfacts.org do work today, but should not be encouraged.

stephanegigandet:
Note that we already do have the "cc" parameter in product query configurations.

PR #219, opened 2021-08-23
Added initial Settings.dart

M123-dev:
This is the first PR for #216 

Moved the uri generating progress to Settings.dart

Also fixed a bug that `getAutocompletedSuggestions` always took the production server and redirected the robotoff test requests to the now available robotoff test server and adapted the tests appropriately


lib/model/Settings.dart
```diff
@@ -0,0 +1,40 @@
+import 'package:openfoodfacts/utils/QueryType.dart';
+
+class Settings {
```
jasmeet0817, line 3: This change makes sense, thanks for adding. Question: Can we later add other settings like language etc to this file ? Currently it's passed into the getUri method. What's the plan for expanding the this file to include more settings, what else will be here etc (just curious).

M123-dev: Thanks for your review Jasmeet, exactly I am going to expand that pr later on, but since it's going to bring a lot refactoring code changes I decided to split the PR so that's it stays reviewabls.

The next thing will be a `globalQueryType` and the depreciation of passing them to every method manually and also creating a global user and moving the user agent into the settings.

I am also planning to add some way to choose the queried language from inside the settings but for that I have to find a reliable way which still customizable enough for use in translation apps.

Documentation will be added down the line.

lib/openfoodfacts.dart
```diff
@@ -77,16 +78,20 @@ class OpenFoodAPIClient {
...
+  }) async {
     var parameterMap = <String, String>{};
     parameterMap.addAll(user.toData());
     parameterMap.addAll(product.toServerData());
 
-    var productUri = Uri(
-        scheme: URI_SCHEME,
-        host: queryType == QueryType.PROD ? URI_PROD_HOST : URI_TEST_HOST,
-        path: '/cgi/product_jqm2.pl');
+    var productUri = Settings.getUri(
```
monsieurtanuki, line 90: You should deprecate `URI_SCHEME`, `URI_PROD_HOST`, `URI_TEST_HOST`, `URI_PROD_HOST_ROBOTOFF` and `URI_TEST_HOST_ROBOTOFF`.

monsieurtanuki (changes requested):
Hi @M123-dev! Minor fixes needed - tell me when you don't agree.

lib/model/Settings.dart
jasmeet0817: monsieurtanuki@ I think the idea is to turn this file into a generic settings/config class, changing these parameters can change the app's behavior and the idea is that a dev can update these parameters to test different behaviors locally. 

Currently it does look like UriHelper but in the future it would turn into AppSettings or AppConfig, we can call it one of those? Any other ideas ?

lib/model/Settings.dart
```diff
@@ -0,0 +1,40 @@
...
+  Settings._();
+
+  static String uriScheme = 'https';
+  static String uriProdHost = 'world.openfoodfacts.org';
+  static String uriTestHost = 'world.openfoodfacts.net';
+
+  static String uriProdHostRobotoff = 'robotoff.openfoodfacts.org';
+  static String uriTestHostRobotoff = 'robotoff.openfoodfacts.net';
+
+  static Uri getUri({
```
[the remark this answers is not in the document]
jasmeet0817: good comments monsieurtanuki@.

For this one, I think static methods make sense here, specially because we may never want to create any instance of "Settings", since the app should only have 1 version of "settings". 

If there will never be more than one instance of "settings", I think it makes sense to use static methods rather than instance methods.

lib/model/Settings.dart
monsieurtanuki: @jasmeet0817 Fair enough but for the moment I cannot visualize other usages of this class: can you please name one - in the next future?

lib/model/Settings.dart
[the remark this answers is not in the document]
monsieurtanuki: @jasmeet0817 As I said before, I don't know how far this class is supposed to go.

That being said, I can still imagine cases where it's healthier to use rather `Settings.getInstance().getUri(...)` than `Settings.getUri(...)`.
E.g. if you want to run at the same time two queries: one with your standard settings ("your" current static version) and the other one with fallback settings (alternate server, fallback language like `'world'` or `'en'`, smaller page size). I don't say it's the most typical use case, but with the `static` syntax you just can't do that, and with the `getInstance()` syntax you can, and you can use the `copyWith` syntax.

Not sure what the standards on flutter/dart say about it, though.

lib/openfoodfacts.dart
```diff
@@ -140,11 +145,11 @@ class OpenFoodAPIClient {
     User? user,
     QueryType queryType = QueryType.PROD,
   }) async {
-    var productUri = Uri(
-        scheme: URI_SCHEME,
-        host: queryType == QueryType.PROD ? URI_PROD_HOST : URI_TEST_HOST,
-        path: 'api/v0/product/' + barcode + '.json',
-        queryParameters: {'lc': language.code});
+    var productUri = Settings.getUri(
+      path: 'api/v0/product/' + barcode + '.json',
```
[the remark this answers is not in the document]
M123-dev: Yeah I've also noticed it, but forgot to re-edit after I discarded some changes.

lib/openfoodfacts.dart
M123-dev: Ohh thanks for the reminder, missed that one

lib/model/Settings.dart
M123-dev: @monsieurtanuki 

> 
> The next thing will be a `globalQueryType` and the depreciation of passing them to every method manually and also creating a global user and moving the user agent into the settings.
> 
> I am also planning to add some way to choose the queried language from inside the settings but for that I have to find a reliable way which still customizable enough for use in translation apps.

---------

I myself am not satasfied with the current name, what about something like `OpenFoodAPIClientSettings` or in short `OpenFoodSettings`

lib/model/Settings.dart
```diff
@@ -0,0 +1,40 @@
+import 'package:openfoodfacts/utils/QueryType.dart';
+
+class Settings {
+  Settings._();
+
+  static String uriScheme = 'https';
```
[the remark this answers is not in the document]
M123-dev: I don't think we should make it const, of course its very likely not getting changed by anyone but why should we stop someone from doing so if he want to.

lib/model/Settings.dart
```diff
@@ -0,0 +1,40 @@
...
+  static String uriProdHost = 'world.openfoodfacts.org';
+  static String uriTestHost = 'world.openfoodfacts.net';
+
+  static String uriProdHostRobotoff = 'robotoff.openfoodfacts.org';
+  static String uriTestHostRobotoff = 'robotoff.openfoodfacts.net';
+
+  static Uri getUri({
+    String? path,
```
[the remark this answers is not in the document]
M123-dev: `Uri` only requieres a nullable String, we can of course change it but I dont think it makes a real difference

lib/model/Settings.dart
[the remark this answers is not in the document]
monsieurtanuki: @M123-dev I see what you mean. But even if we're very open minded, there's a very high probability that the scheme will be either `'http'` or `'https'`, right?
In that case I would rather recommend something like that - assuming we use the "non static" way of writing this class:
```dart
static String _URI_SCHEME_HTTP = 'http';
static String _URI_SCHEME_HTTPS = 'https';
bool secureScheme = true;
String _getScheme() => secureScheme ? _URI_SCHEME_HTTPS : _URI_SCHEME_HTTP;
```
```

File: lib/model/Settings.dart (hunk 6 above)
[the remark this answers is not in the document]
Comment (monsieurtanuki) [reply lib/model/Settings.dart:14]: Our added value would be to say to the developers: put something relevant (= not null) because we process only relevant URIs.

File: lib/model/Settings.dart (hunk 1 above)
Comment (monsieurtanuki) [reply lib/model/Settings.dart:3]: @M123-dev We're still only talking about the URI, am I right?
Anyway, what about `OpenFoodAPISettings`?
Or, we could even move the code of `Settings` into `OpenFoodAPIClient`: there's no other place where this code could be reused. Not convinced it's a good idea, though, and that would probably thwart my idea of `getInstance()` vs. `static`.

Comment (jasmeet0817) [reply lib/model/Settings.dart:3]: So the general idea here (correct me if I'm wrong) is to come up with a set of settings that can change the behavior of the app. For example if you wanted to test the eco score in Switzerland, there is no way of doing it right now except changing the getProduct method and passing the 'cc' query parameter set as 'DE' or "FR' etc, but in the future we can force set the setting for "cc" to allow eco scores to be shown in a local version of the app. This then becomes the poor man's version of flags for the repo.

The reason it makes sense to extract this into a class of it's own rather than inside an existing one like `OpenFoodAPIClient` is because it should be easy for new developers to come and see where all such settings live, so they can easily override something they need to.

I would actually advice not having any methods in this class (although upto you guys), but just a bunch of overridable settings. And then you could have two classes:

1. `OpenFoodAPISettings` with all the settings.
2. (optional - if needed) UriHelper with the helper methods, this can use the variables from `OpenFoodAPISettings` as needed.

What do you guys think?

File: lib/model/Settings.dart (hunk 5 above)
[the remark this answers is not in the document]
Comment (jasmeet0817) [reply lib/model/Settings.dart:6]: +1 monsieurtanuki@ good idea.

And to my above comment, if we wanted to have a UriHelper class separate from Setting class, the setting class would contain `secureScheme` (or `useSecureScheme`) defaulting to true and UriHelper class would contain _getScheme() mehtod.

File: lib/model/Settings.dart (hunk 3 above)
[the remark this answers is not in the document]
Comment (M123-dev) [reply lib/model/Settings.dart:13]: I think I understand what you mean but to achieve what you are writing about we would have to pass Settings to every method, this would again lead to much more comlexity which is probably not worth it,

File: lib/model/Settings.dart (hunk 1 above)
Comment (M123-dev) [reply lib/model/Settings.dart:3]: @monsieurtanuki I actually have thought about this when starting, this would also allow us to make getUri private but would again add more code to the already 600 line long OpenFoodAPIClient

File: lib/model/Settings.dart (hunk 3 above)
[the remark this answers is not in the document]
Comment (monsieurtanuki) [reply lib/model/Settings.dart:13]: Not exactly, as `Settings.getInstance()` does not need to be passed as a parameter.
Maybe __later__ an optional `Settings` parameter, when the developer wants to override the "global" `Settings`.

Anyway, let's not spend too much time on this `getInstance` / `static` question for the moment; just finish your "static" version (in this PR and later). Then we'll see how important `Settings` are, and how relevant it would be to switch to the "instance" mode. Then, we publish to pub.dev.

File: lib/model/Settings.dart (hunk 1 above)
Comment (M123-dev) [reply lib/model/Settings.dart:3]: @jasmeet0817 thats indeed a good idea in terms of overview and abstraction

Comment (monsieurtanuki) [reply lib/model/Settings.dart:3]: If you ask me this is what I would do. Looks much cleaner to me, but we've already spent a lot of time on the subject, so feel free to ignore this suggestion:
```dart
// default value: prod
OpenFoodAPISettings.instance = OpenFoodAPISettings.fromQueryType(QueryType.PROD);

class OpenFoodAPISettings {
  static OpenFoodAPISettings instance;
  
  OpenFoodAPISettings({
      this.scheme,
      this.host,
      this.robotoffHost,
    });

  OpenFoodAPISettings.fromQueryType(final QueryType queryType)
    : this(
        scheme: _URI_SCHEME_HTTPS,
        host: queryType == QueryType.PROD ? _URI_HOST_PROD : _URI_HOST_TEST,
        robotoffHost: queryType == QueryType.PROD ? _URI_ROBOTOFF_HOST_PROD : _URI_ROBOTOFF_HOST_TEST,
      );

  final String scheme;
  final String host;
  final String robotoffHost;

  static String const _URI_SCHEME_HTTPS = 'https';
  static String const _URI_HOST_PROD = 'world.openfoodfacts.org';
  static String const _URI_HOST_TEST = 'world.openfoodfacts.net';

  static String const _URI_ROBOTOFF_HOST_PROD = 'robotoff.openfoodfacts.org';
  static String const _URI_ROBOTOFF_HOST_TEST = 'robotoff.openfoodfacts.net';

  Uri _getUri({
    String path,
    Map<String, dynamic>? queryParameters,
    bool isRobotoff,
  }) =>
    Uri(
      scheme: scheme,
      host: isRobotoff ? robotoffHost : host,
      path: path,
      queryParameters: queryParameters,
    );
  }

  Uri getUri({
    String path,
    Map<String, dynamic>? queryParameters,
  }) => _getUri(path: path, queryParameters: queryParameters, isRobotoff: false);

  Uri getRobotoffUri({
    String path,
    Map<String, dynamic>? queryParameters,
    }) => _getUri(path: path, queryParameters: queryParameters, isRobotoff: true);
}
```

jasmeet0817: monsieurtanuki@ I agree with most of your code but I still think it would be beneficial (more modular and dev friendly) to abstract out a settings-only class (`OpenFoodAPISettings`) that only contains a bunch of variables that can be changed to modify the app behavior (example: `language`, `country`, `queryType`, `useSecureScheme` etc). And then we can have a `UriHelper` class which would mostly have the code you mention in your comment. It would use `apiSettings.useSecureScheme` and `apiSettings.queryType` inside where `apiSettings = OpenFoodAPISettings.getInstance()`.

For tests you could do `testSettings =OpenFoodAPISettings.getTestInstance()` to return test instance instead.

Example usage: This way when you call a method like `OpenFoodAPIClient.getRobotoffQuestionsForProduct`, you would pass in the `apiSettings` as a parameter, it will then call `uriHelper` with the `apiSettings`. For tests you would pass `testSettings` instead.

Also for example, if you had to run a local instance without https, you would just go to settings file and set useSecureScheme = false. You wouldn't have to deal with any other code which is unrelated to settings. 

Just my 2 cents, I still think you guys have been on the project longer so it should be your call though.

lib/model/OpenFoodAPISettings.dart
```diff
@@ -0,0 +1,63 @@
+import 'package:openfoodfacts/utils/QueryType.dart';
+
+class OpenFoodAPISettings {
+  static OpenFoodAPISettings? instance;
+
+  OpenFoodAPISettings({
+    this.scheme,
```
monsieurtanuki, line 7: Actually they should be `required`, shouldn't they?

lib/model/OpenFoodAPISettings.dart
```diff
@@ -0,0 +1,63 @@
+import 'package:openfoodfacts/utils/QueryType.dart';
+
+class OpenFoodAPISettings {
+  static OpenFoodAPISettings? instance;
```
monsieurtanuki, line 4: Right now we don't need this static field, as you use `OpenFoodAPISettings.fromQueryType(queryType)` in `openfoodfacts.dart`.

lib/model/OpenFoodAPISettings.dart
```diff
@@ -0,0 +1,63 @@
...
+  OpenFoodAPISettings.fromQueryType(final QueryType? queryType)
+      : this(
+          scheme: _URI_SCHEME_HTTPS,
+          host: queryType == QueryType.PROD ? _URI_HOST_PROD : _URI_HOST_TEST,
+          robotoffHost: queryType == QueryType.PROD
+              ? _URI_ROBOTOFF_HOST_PROD
+              : _URI_ROBOTOFF_HOST_TEST,
+        );
+
+  final String? scheme;
```
monsieurtanuki, line 21: `String` instead of `String?` I guess.

lib/openfoodfacts.dart
```diff
@@ -589,7 +577,7 @@ class OpenFoodAPIClient {
     final OpenFoodFactsLanguage language,
   ) async {
     const String FIELD = 'environment_infocard';
-    final Uri uri = Settings.getUri(
+    final Uri uri = OpenFoodAPISettings().getUri(
```
monsieurtanuki, line 580: That can't work. We need to change the constructor of `OpenFoodAPISettings`.
Something like that:
```dart
OpenFoodAPISettings.fromQueryType(
    final QueryType queryType, {
  })
     : this(
         scheme: _URI_SCHEME_HTTPS,
         host: queryType == QueryType.PROD ?
            (language != null ? 'world-${language.code}.openfoodfacts.org' : 'world.openfoodfacts.org')
            : (language != null ? 'world-${language.code}.openfoodfacts.net' : 'world.openfoodfacts.net'),
         robotoffHost: queryType == QueryType.PROD
             ? _URI_ROBOTOFF_HOST_PROD
             : _URI_ROBOTOFF_HOST_TEST,
       );
```

monsieurtanuki (changes requested):
@M123-dev That's indeed more what I had in mind. There are still needed fixes - see my comments.
But it's a nice first step, where we reuse the `QueryType` parameters in `openfoodfacts.dart`.

@jasmeet0817 I read your latest comment but unfortunately I have no more bandwidth for this topic until Thursday morning. I'm not sure I see the added value of separating fields and methods here and anyway it's getting late :)

jasmeet0817:
@monsieurtanuki no probs, how about this, lets go ahead with this change, and also lets continue adding the other settings/methods into this file as planned originally, once the additions saturate, I will send a pull request with what I have in mind, and we can talk on that.

M123-dev:
@jasmeet0817 @monsieurtanuki 

I have updated the code a bit, when we now run:

```dart
  print(OpenFoodAPISettings.fromQueryType(QueryType.PROD).getUri(path: 'foo'));
  print(OpenFoodAPISettings.fromQueryType(QueryType.TEST).getUri(path: 'foo'));

  print('------------');

  OpenFoodAPISettings.instance = OpenFoodAPISettings(host: 'myDomain');

  print(OpenFoodAPISettings.fromQueryType(QueryType.PROD).getUri(path: 'foo'));
  print(OpenFoodAPISettings.fromQueryType(QueryType.TEST).getUri(path: 'foo'));

  print('------------');

  OpenFoodAPISettings.instance =
      OpenFoodAPISettings(globalQueryType: QueryType.TEST);

  //Its null to imitate passing a nullable queryType from every method in OpenFoodAPIClient after deprecating
  print(OpenFoodAPISettings.fromQueryType(null).getUri(path: 'foo'));
  print(OpenFoodAPISettings.fromQueryType(null).getUri(path: 'foo'));
```
We get:

```
https://world.openfoodfacts.org/foo
https://world.openfoodfacts.net/foo
------------
https://mydomain/foo
https://mydomain/foo
------------
https://world.openfoodfacts.net/foo
https://world.openfoodfacts.net/foo
```

I don't know what you think, but I find the static version much nicer and cleaner. 
For reference here with also build in globalQueryType:

https://github.com/openfoodfacts/openfoodfacts-dart/compare/settings-backup#diff-105a353d869ad12536e1f1e6d55da2d48f2924c3dcf59a235aaeaa91647c4798

jasmeet0817:
If you ask me, I would recommend just working with statics if possible.

Using the instance when it's not needed, feels awkward to me. The only time I don't like statics is if it hinders testability, which doesn't seem to be the case here.

monsieurtanuki:
Correction: according to @stephanegigandet in https://github.com/openfoodfacts/openfoodfacts-dart/issues/216#issuecomment-911336577, the syntax `world-de.off.org` is not to be encouraged, and `world.off.org` should always be used instead, with `lc` or/and `cc` URL parameters for language and country.

Which probably simplifies things and means that for both URIs (off and robotoff) we currently just need `final String scheme` and `final String host`. And no other parameter.

M123-dev:
Since this pr is so crowded, I am going to move it to #222 222

PR #220, opened 2021-08-26
JSON object for knowledge panel with tests 

jasmeet0817:
This adds a Json object for Knowledge Panels (KP) that will be served via the new Json API. To begin with, the first version of the KP API will be serving generic Eco score related KPs.

More information can be found in this working doc - https://docs.google.com/document/d/1vJ9gatmv8pCXxyOERmYD16jOKRWJpz1RaQQ5MEcTxms/


monsieurtanuki (dismissed):
@jasmeet0817 Please forgive me for being cynical: I don't care about `//` comments, I need `///` comments as in https://dart.dev/guides/language/effective-dart/documentation#do-use--doc-comments-to-document-members-and-types
The general idea is - beyond adding clarity as with `//` - to give us a better pub.dev score and to have docs being automatically generated.

stephanegigandet:
Here is a test file for the knowledge panel API: https://github.com/openfoodfacts/openfoodfacts-server/pull/5578

https://fr.openfoodfacts.dev/files/knowledge_panel_test.json (login and password: off )

lib/model/KnowledgePanelElement.dart
```diff
@@ -0,0 +1,204 @@
...
+
+  @JsonValue('image')
+  IMAGE,
+
+  /// Disclaimer notes that the client may or may not choose to display.
+  @JsonValue('panel')
+  PANEL,
+
+  /// Disclaimer notes that the client may or may not choose to display.
+  @JsonValue('table')
```
stephanegigandet, line 137: Server side I changed 'table' to 'table_2_columns', to clearly mark that it is expected to have only 2 columns. And we can have a more generic 'table' type with any number of columns.

jasmeet0817: yeah, but can't we fit table_2_columns in the generic table ? client can easily figure out that there are two columns. if we want to pass extra info to the client, we can pass it in the type ("percent" in this case)?

lib/model/KnowledgePanel.dart
```diff
@@ -0,0 +1,109 @@
...
+
+  const KnowledgePanel(
+      {required this.type,
+      required this.level,
+      required this.topics,
+      required this.title,
+      required this.elements,
+      this.subtitle,
+      this.iconUrl,
+      this.grade});
```
slava-sh, line 102: nit: Add trailing comma? 

https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#use-a-trailing-comma-for-arguments-parameters-and-list-items-but-only-if-they-each-have-their-own-line

jasmeet0817: Thanks, done here and in other places I found.

stephanegigandet (approved):
Looks good to me, we can merge, and tweak things later if needed.

PR #222, opened 2021-09-02
Added OpenFoodAPISettings

M123-dev:
Replacement of #219 for #216

@monsieurtanuki @jasmeet0817


M123-dev:
### Current State
I have converted the PR back to static fields since fortunately we don't have to specify the language code in the pre url but in the params. I put all the configurations parameters in `OpenFoodAPIConfiguration` I know it goes against the Dart linting rules but I found it better this way, if you disagree with me I can move it to OpenFoodAPIClient.
Also the getUri methods are now in a separate `UriHelper` 

### Future
Half of the test still show a deprecated info, I will take care of that when I have some more time. The plan for the future is to make it possible to set a global language and country code directly from the configuration.

stephanegigandet (approved):
Looks good to me, thanks!

M123-dev:
I have converted the last tests and removed the deprecated disclaimer so that we can make both, prod and test server requests in some tests.

PR #228, opened 2021-09-14
Added globalLanguages and globalCC to OpenFoodAPIConfiguration 

### LINKED ISSUE #216: Allow clients to pass the country as a parameter request while getting a product's details

M123-dev:
Closes #216 

CC: @blazern

### DIFF
--- a/lib/utils/AbstractQueryConfiguration.dart
+++ b/lib/utils/AbstractQueryConfiguration.dart
@@ -69,6 +69,8 @@ abstract class AbstractQueryConfiguration {
 
     if (cc != null) {
       result.putIfAbsent('cc', () => cc!);
+    } else if (OpenFoodAPIConfiguration.globalCC != null) {
+      result.putIfAbsent('cc', () => OpenFoodAPIConfiguration.globalCC!);
     }
 
     if (fields != null) {
--- a/lib/utils/AbstractQueryConfiguration.dart
+++ b/lib/utils/AbstractQueryConfiguration.dart
@@ -1,4 +1,5 @@
 import 'package:openfoodfacts/utils/LanguageHelper.dart';
+import 'package:openfoodfacts/utils/OpenFoodAPIConfiguration.dart';
 import 'package:openfoodfacts/utils/ProductFields.dart';
 
 /// Abstract Query Configuration, that helps build API URI
@@ -20,7 +21,7 @@ abstract class AbstractQueryConfiguration {
   ///
   /// Please see https://github.com/openfoodfacts/openfoodfacts-dart/blob/master/DOCUMENTATION.md#about-languages-mechanics
   /// for detailed explanation on how to work with multiple languages.
-  List<OpenFoodFactsLanguage> languages;
+  List<OpenFoodFactsLanguage>? languages;
 
   // TODO: deprecated from 2021-07-20 (#185); remove when old enough
   @Deprecated('Use parameters language or languages instead')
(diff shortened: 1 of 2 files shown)


lib/utils/AbstractQueryConfiguration.dart
```diff
@@ -31,23 +32,35 @@ abstract class AbstractQueryConfiguration {
...
     this.lc,
     this.cc,
     this.fields,
   }) {
     fields ??= [ProductField.ALL];
-    if ((language != null || lc != null) && languages.isNotEmpty) {
-      throw ArgumentError(
-          '[languages] cannot be used together with [language]/[lc]');
+    if (languages != null) {
+      if ((language != null || lc != null) && languages!.isNotEmpty) {
```
blazern, line 42: The second check will always be true
image

lib/utils/AbstractQueryConfiguration.dart
```diff
@@ -31,23 +32,35 @@ abstract class AbstractQueryConfiguration {
...
   }
 
   /// Returns the corresponding API URI parameter map
   Map<String, String> getParametersMap() {
     final Map<String, String> result = {};
+    late List<OpenFoodFactsLanguage> languages;
+
+    if (language != null) {
+      languages = [language!];
+    } else if (this.languages != null && this.languages!.isNotEmpty) {
```
blazern, line 56: `this.languages!.isNotEmpty`
Are you sure it should be illegal to pass an empty languages list?
Can't an SDK user request no languages?

lib/utils/AbstractQueryConfiguration.dart
M123-dev: It's language**s** and language, or have I missed something

blazern: Oh, you didn't miss anything, it's me who have missed the fact that these variables are different :sweat_smile: 
Sorry!

lib/utils/AbstractQueryConfiguration.dart
M123-dev: In fact, I only added this during the review reading.

True it would make sense to allow to pass an empty list to specify no language even if you have set a global value.

blazern: Do you mean there's a discussion somewhere which prompted you to add this check?
Could you please give a link to the discussion? :)
I'm curious of the reasoning!

lib/utils/AbstractQueryConfiguration.dart
M123-dev: No problem, they look very similar though, what about renaming the local variable from languages to languageList

lib/utils/AbstractQueryConfiguration.dart
M123-dev: Ohh, there is no other discussion, I just read the whole changed code before opening the pr, I just thought that it makes no sense passing a empty list and that we then shoud rather use the global value. But actually there is no problem to pass an empty list, because that way we create  a way to not add the lc to querie despite the global languages field is set.

M123-dev: @blazern I removed the `&& this.languages!.isNotEmpty`

blazern:
Some tests fail!

M123-dev:
@blazern I think thats because they both run at the same time and this sometimes causes conflicts on the test server, running them one after the other should fix it

blazern (approved):

monsieurtanuki (approved):

PR #236, opened 2021-09-17
Update KnowlededgePanel related objects as per the server response and call the BE server from openfoodfacts.dart

jasmeet0817:


stephanegigandet (approved):
Looks good, but my preference would be to have an exact match (besides the foo_bar vs fooBar) between the variable names in the API and in Flutter.

And if we do that, I would prefer to shorten the variable names.
e.g. instead of having table_id, table_type, cell_label etc. , just have "id", "type", "text" (or "label" but then we should change it in the API). e.g. if we have a class "SomeObject", I don't see much value in having all properties start with someObject (someObjectId, someObjectType etc.).

jasmeet0817:
Agreed, we should probably sync on the naming across FE and BE.

I renamed column -> column_descriptor ... as it is describing the type and label.
and text -> label

what do you think about these suggestions ? let me know your preferences and I can change.

stephanegigandet:
I'd prefer the shortest and simplest names, so just text and column. I'd prefer also to change table_type to just type (just like for the other elements where we have only type and not cell_type etc.). Same for element_type -> type.

jasmeet0817:
Just "type" definitely makes sense. I just think "text" is not a future proof name, but since you have a strong preference I'll modify to it for now.

jasmeet0817:
> Looks good, but my preference would be to have an exact match (besides the foo_bar vs fooBar) between the variable names in the API and in Flutter.

Made changes so that we have exact matches.

PR #240, opened 2021-09-21
Add isAttributeImportant method to ProductPreferencesManager

jasmeet0817:


M123-dev (approved):
I don't know what this will be used for or if something similar already exists but the code looks good

jasmeet0817:
> I don't know what this will be used for or if something similar already exists but the code looks good

Thanks Marvin, something similar doesn't exist AFAICT and this is for a new PR for displaying important attributes on the summary card, the current mechanism for ordering will change

lib/personalized_search/product_preferences_manager.dart
```diff
@@ -102,6 +102,18 @@ class ProductPreferencesManager {
     return result;
   }
 
+  /// Returns whether an attribute is important as per the user preferences.
+  bool isAttributeImportant(String attributeId) {
+    final String importanceId = getImportanceIdForAttributeId(attributeId);
+    final int? importanceIndex =
+        _availablePreferenceImportances?.getImportanceIndex(importanceId);
+    if (importanceIndex == null || importanceIndex == 0) {
```
slava-sh, line 110: Use `PreferenceImportance.INDEX_NOT_IMPORTANT`?

lib/personalized_search/product_preferences_manager.dart
```diff
@@ -102,6 +102,18 @@ class ProductPreferencesManager {
     return result;
   }
 
+  /// Returns whether an attribute is important as per the user preferences.
+  bool isAttributeImportant(String attributeId) {
```
slava-sh, line 106: Should this return a `bool?`, like the other methods relying on `_availablePreferenceImportances` in this file?

jasmeet0817: It depends upon whether we want to continue having 4 levels of "PreferenceImportance" or turn it into a boolean, my understanding is that we want to turn it into a boolean as of now, but if we later decide to continue having the 4 levels I would  modify the method to getAttributeImportance and return PreferenceImportance, but as of now, I think this is fine

lib/personalized_search/product_preferences_manager.dart
jasmeet0817: Good call

lib/personalized_search/product_preferences_manager.dart
slava-sh: This is just a question of consistency of the API. Take a look at `getImportanceIndex()`, for example. It returns `null` if and only if `_availablePreferenceImportances` is `null`. I don't really know what that means, but I think it would be more consistent for `isAttributeImportant()` to implement the same behaviour.

jasmeet0817: I see what you mean, I mistook your original comment to be "Should this return a bool?" :D 

you're right, I'll send a follow up PR.

PR #243, opened 2021-09-23
Return "bool?" from isAttributeImportant

jasmeet0817:


M123-dev (approved):

lib/personalized_search/product_preferences_manager.dart
```diff
@@ -103,16 +103,16 @@ class ProductPreferencesManager {
...
         _availablePreferenceImportances?.getImportanceIndex(importanceId);
-    if (importanceIndex == null ||
-        importanceIndex == PreferenceImportance.INDEX_NOT_IMPORTANT) {
-      // Not important.
-      return false;
+    if (importanceIndex == null) {
+      return null;
     }
-    return true;
+    return importanceIndex == PreferenceImportance.INDEX_NOT_IMPORTANT
```
slava-sh, line 113: sugg: `importanceIndex != PreferenceImportance.INDEX_NOT_IMPORTANT`

lib/personalized_search/product_preferences_manager.dart
```diff
@@ -103,16 +103,16 @@ class ProductPreferencesManager {
...
-  bool isAttributeImportant(String attributeId) {
+  bool? isAttributeImportant(String attributeId) {
     final String importanceId = getImportanceIdForAttributeId(attributeId);
     final int? importanceIndex =
         _availablePreferenceImportances?.getImportanceIndex(importanceId);
-    if (importanceIndex == null ||
-        importanceIndex == PreferenceImportance.INDEX_NOT_IMPORTANT) {
-      // Not important.
-      return false;
+    if (importanceIndex == null) {
```
slava-sh, line 110: sugg: Check `_availablePreferenceImportances == null` instead, since that's the only reason why `importanceIndex` might be null.

lib/personalized_search/product_preferences_manager.dart
jasmeet0817: I think this is just a matter of taste, I see where you're coming from "returning `true first `false` later" but I'm avoiding negation in the statement

lib/personalized_search/product_preferences_manager.dart
jasmeet0817: Ah but we don't know if that's true in the future, maybe `getImportanceIndex` will return `null` in another case later.

I think the point of the cl was to stay consistent with the other APIs that only return null if `getImportanceIndex` returns null

lib/personalized_search/product_preferences_manager.dart
slava-sh: I agree double negation doesn't look good

lib/personalized_search/product_preferences_manager.dart
slava-sh: `int getImportanceIndex(...)` can never return `null` :)

jasmeet0817: Ack, feel free to send a follow up PR, I'm not sure it matters much :)

PR #251, opened 2021-10-01
Add Permanent_attributes to Attribute.dart, user will always see these attributes on product page and can't toggle them off

jasmeet0817:


jasmeet0817:
Usage: https://github.com/openfoodfacts/smooth-app/pull/606

M123-dev (approved):
The code has logically no problems, but is it good to put something like this in the public package. If we leave it like this, we should definitely add a note to the comment that this is used for smoothie and has no further influence for the package users.

monsieurtanuki:
> The code has logically no problems, but is it good to put something like this in the public package. If we leave it like this, we should definitely add a note to the comment that this is used for smoothie and has no further influence for the package users.

My 2 cents: I agree with @M123-dev - it looks like it should apply only to smoothie and I would not pollute the library with irrelevant code. I would put the permanent attributes directly in smoothie instead.

stephanegigandet:
Agreed, this behaviour is specific to Smoothie, it should not be hardcoded in the Dart package.

peterwvj:
I agree with the above comments. Please don't merge.

jasmeet0817:
Oh yes, most definitely a good point, sorry I missed it, will revert

PR #252, opened 2021-10-02
New recommended linting rules

M123-dev:
[pedantic](https://pub.dev/packages/pedantic) is discontinued and deprecated, insted we should now use `package:lints` (or `package:flutter_lints`)

The analyzer complained that it couldn't find `package:lints` in the example folder, so I added it there too.

I renamed `requested_org` to `orgName` to fit the linting rules in register without deprecating it, also this name makes more sense this way

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,4 +1,4 @@
-include: package:pedantic/analysis_options.yaml
+include: package:lints/recommended.yaml
 
 analyzer:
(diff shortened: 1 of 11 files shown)


M123-dev:
I have re-run the test from the last merged PR and there were also 22 failing now
cf: https://github.com/openfoodfacts/openfoodfacts-dart/runs/3775898775

PR #253, opened 2021-10-03
Add List<Parameter> parametersList to ProductListQueryConfiguration

blazern:
Add `List<Parameter> parametersList` to `ProductListQueryConfiguration`.
It's not possible to use pagination without the field and requesting more than 24 barcodes becomes much harder.


monsieurtanuki (commented):
Hi @blazern!
Nothing wrong with your code, but I think that would make more sense to create the `List<Parameter>` in the abstract class only.
Then we don't need method `parametersListToParametersMap`, whose code would be moved to `getParametersMap` (in the abstract class).
What do you think of that?

lib/model/parameter/SortBy.dart
```diff
@@ -27,3 +20,18 @@ enum SortOption {
...
+  String get name {
+    switch (this) {
+      case SortOption.PRODUCT_NAME:
+        return 'product_name';
+      case SortOption.CREATED:
+        return 'created_t';
+      case SortOption.EDIT:
+        return 'last_modified_t';
+      case SortOption.POPULARITY:
+        return 'unique_scans_n';
```
teolemon, line 34: There's a sort by nothing option
https://github.com/openfoodfacts/openfoodfacts-server/pull/5779

blazern:
> Hi @blazern! Nothing wrong with your code, but I think that would make more sense to create the `List<Parameter>` in the abstract class only. Then we don't need method `parametersListToParametersMap`, whose code would be moved to `getParametersMap` (in the abstract class). What do you think of that?

Hello @monsieurtanuki!
I thought a little bit about your comment and about how the base abstract class doesn't really need a `List<Parameter>` and then I realized that most of the Parameter subtypes don't make any sense for `ProductListQueryConfiguration` either.
So I've pushed a commit which adds `page`, `pageSize` and `sortOption` fields to `ProductListQueryConfiguration` instead of `List<Parameter>`.

I know that's not what you've asked to do, but I think it's a bit better this way.
What do you think?

lib/model/parameter/SortBy.dart
blazern: Hello!
Do you want me to add it in this pull request?

lib/model/parameter/SortBy.dart
```diff
@@ -27,3 +20,18 @@ enum SortOption {
   CREATED,
   EDIT,
 }
+
+extension SortOptionExt on SortOption {
+  String get name {
```
monsieurtanuki, line 25: @blazern I know that you did not agree, but from #184 there's a sort of consensus about using `Map<MyEnum, String>`. The way it was coded in this file just before your PR ;)

monsieurtanuki (commented):
@blazern Adding `page`, `pageSize` and `sortOption` to `ProductListQueryConfiguration` probably makes sense if we cannot use the other `Parameter`s for this type of query (is that a fact?), and it makes those useful parameters more visible, that's cool.

That being said, you duplicate code in `getParametersMap`, code that was in subclasses of `Parameter`. And with a slight difference: in `ProductSearchQueryConfiguration` we use `result.putIfAbsent`.

My suggestion would be to:
* keep your `page`, `pageSize` and `sortOption` parameters in the constructor of `ProductListQueryConfiguration` - provided that the other `Parameter`s have no effect for this query
* don't use those parameters as is but transform them into `List<Parameter>` when needed
* that would probably mean moving the `List<Parameter>` field to the abstract class
* the abstract class would do all the work, with a `List<Parameter>` either provided by `ProductSearchQueryConfiguration`, build by `ProductListQueryConfiguration` from the `page`, `pageSize` and `sortOption` parameters, or empty for `ProductQueryConfiguration`

What's your opinion on that?

lib/model/parameter/SortBy.dart
blazern: Done

lib/model/parameter/SortBy.dart
blazern: Returned back to how it was before

blazern:
> @blazern Adding `page`, `pageSize` and `sortOption` to `ProductListQueryConfiguration` probably makes sense if we cannot use the other `Parameter`s for this type of query (is that a fact?), and it makes those useful parameters more visible, that's cool.
> 
> That being said, you duplicate code in `getParametersMap`, code that was in subclasses of `Parameter`. And with a slight difference: in `ProductSearchQueryConfiguration` we use `result.putIfAbsent`.
> 
> My suggestion would be to:
> 
>     * keep your `page`, `pageSize` and `sortOption` parameters in the constructor of `ProductListQueryConfiguration` - provided that the other `Parameter`s have no effect for this query
> 
>     * don't use those parameters as is but transform them into `List<Parameter>` when needed
> 
>     * that would probably mean moving the `List<Parameter>` field to the abstract class
> 
>     * the abstract class would do all the work, with a `List<Parameter>` either provided by `ProductSearchQueryConfiguration`, build by `ProductListQueryConfiguration` from the `page`, `pageSize` and `sortOption` parameters, or empty for `ProductQueryConfiguration`
> 
> 
> What's your opinion on that?

Mainly my opinion is that it's too hard to contribute to this repo.
I've applied all of your suggestions to the commit.

monsieurtanuki (commented):
Thank you @blazern!
There's just a tiny probable confusion with `import`/`export` (see my comment): once it's fixed I'll approve this PR.

lib/utils/ProductListQueryConfiguration.dart
```diff
@@ -1,3 +1,7 @@
+import 'package:openfoodfacts/openfoodfacts.dart';
+export 'package:openfoodfacts/model/parameter/Page.dart';
```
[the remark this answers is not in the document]
blazern: Thanks! Fixed!

monsieurtanuki (approved):
Thank you @blazern!

lib/utils/AbstractQueryConfiguration.dart
```diff
@@ -30,12 +32,15 @@ abstract class AbstractQueryConfiguration {
...
 
+  List<Parameter> additionalParameters;
+
   AbstractQueryConfiguration({
     this.language,
     this.languages,
     this.lc,
     this.cc,
     this.fields,
+    this.additionalParameters = const [],
```
M123-dev, line 43: I am not sure since the analyzer doesn't complain but don't we have to specify the list type before the `[]`

M123-dev (approved):
Thanks @blazern just one thing I noticed,  otherwise everything looks good

lib/utils/AbstractQueryConfiguration.dart
blazern: Hmmmm I never do and never had any problems with the `const []` declarations so far.
Do you want me to put a type here?

monsieurtanuki: As long as the analyzer does not complain, let's say it's OK.
I've just double-checked and both syntaxes are accepted (`const []` and `const <Parameter>[]`).
@M123-dev Probably `dart` knowing that it's a `const` and that the field is a `List<Parameter>` understands that it's implicitly a `const <Parameter>[]`. I don't understand either in which specific cases the analyzer actually complains.

blazern:
Pushed a commit with same changes to restart checks

PR #260, opened 2021-10-06
Added the API Parameter 'energy-kj'

ShriAgashe:
Added the missing energy-kj API parameter in the Nutriments.dart file.


ShriAgashe:
I changed the code wherever i found the fields. 
please check it and connfirm
thanks!

PR #261, opened 2021-10-07
Fix lint warnings

gspencergoog:
This just fixes all the lint warnings: no functionality changes.

PR #262, opened 2021-10-07
Add Taxonomy fetching for Categories, Allergens, and Ingredients.

gspencergoog:
## Description

This adds a generic taxonomy fetching API, and a specific one for categories, allergens, and ingredients.

(Currently also includes the contents of https://github.com/openfoodfacts/openfoodfacts-dart/pull/261, which will be merged out when that is submitted).

## Tests

- Added tests for the generic taxonomy API and the specific APIs.

## Related Issues

- https://github.com/openfoodfacts/openfoodfacts-dart/issues/263 (now resolved)


gspencergoog:
OK, @stephanegigandet, I've renamed the new classes with the "Taxonomy" prefix, and put the old names back to what they were, except for `TagType` is still renamed to "TaxonomyType".  Since enums can't be subclassed, and we don't have "real" typedefs (yet), I can't keep both the new and the old name and deprecate the old one. Let me know if you'd rather I just leave it as `TagType` and try and fix the confusing dual use of the word "tag" some other way.

I also updated the tests to use a new `FakeHttpHelper` which can reply with pre-assigned responses.  Check out the tests for the new taxonomy classes to see how it can be used. It means that tests no longer need to hit the test server at all.

Can you please take another look at this, since I've rearranged so much since you saw it before?

PR #267, opened 2021-10-10
fix: Nutriments JSON conversion bug related to micro gram unit

### LINKED ISSUE #265: Nutriments JSON conversion bug related to the micro gram units
**Describe the bug**

JSON serialization of nutriments units is broken for micro grams. Concretely, `Unit.MICRO_G` is not map to any of the valid write API values `'mcg'` or `'µg'` but instead to `'micro-gram'`, which is wrong. This is not a problem when the unit represents grams or milligrams.

**To Reproduce**

Run this test:

```dart
import 'package:openfoodfacts/model/Nutriments.dart';
import 'package:openfoodfacts/utils/UnitHelper.dart';

void main() {
  test('Nutrimens unit bug', () {
    final n = Nutriments(
      fatUnit: Unit.MICRO_G
    );
    final json = n.toJson();
    final dynamic fatUnit = json['fat_unit'];
    expect(fatUnit, anyOf('mcg', 'µg'));
  });
}
```

**Expected behavior**
Above test should pass.

**Actual behaviour**
Above test fails.

image

**package information:**
Openfoodfacts package version:

1.3.5, specifically this commit 099b57df745d698daafa4717a1ac51c5bc222189

**Additional info**

This is also a problem for units (carbohydrates, proteins) but it all boils down to the mapping used by `UnitHelper.unitToString` being wrong so the fix should be straight-forward. Specifically, `Unit.MICRO_G` should be mapped to `'mcg'` or `'µg'`.

peterwvj:
Closes #265


peterwvj:
Hi @monsieurtanuki as always thanks for your great feedbck. I'll go ahead and merge this PR.

PR #269, opened 2021-10-11
Comply to latest BE changes to KnowledgePanels

jasmeet0817:
1. Remove name from KnowledgePanel and add a KnowledgePanelGroupElement to panels
2. Make `Elements` nullable
3. ...and add `type` = `Card` to knowledgePanel

This is to comply to latest BE changes: https://github.com/openfoodfacts/openfoodfacts-server/pull/5958

### DIFF
--- a/lib/model/KnowledgePanels.dart
+++ b/lib/model/KnowledgePanels.dart
@@ -8,7 +8,6 @@ class KnowledgePanels {
   const KnowledgePanels({required this.panelIdToPanelMap});
 
   factory KnowledgePanels.fromJson(Map<String, dynamic> json) {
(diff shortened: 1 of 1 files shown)


M123-dev:
@jasmeet0817 the analyzer is complaining, it seems like the removed `name` is still in KnowledgePanel.dart (maybe the constructor) and is used in tests

jasmeet0817:
> @jasmeet0817 the analyzer is complaining, it seems like the removed `name` is still in KnowledgePanel.dart (maybe the constructor) and is used in tests

Thanks Marvin, indeed you are right, fixed.

PR #287, opened 2021-10-29
Add queries to get the roots from the server

gspencergoog:
## Description

This adds parameters to the Taxonomy queries that have hierarchies (Category, Ingredient, Label) that allow fetching of the root entries of the taxonomy (the entries without parents).

### DIFF
--- a/lib/model/TaxonomyCategory.dart
+++ b/lib/model/TaxonomyCategory.dart
@@ -348,6 +348,21 @@ class TaxonomyCategoryQueryConfiguration extends TaxonomyQueryConfiguration<
           additionalParameters: additionalParameters,
         );
 
+  TaxonomyCategoryQueryConfiguration.roots({
+    List<OpenFoodFactsLanguage>? languages = const [],
+    String? cc,
+    bool includeChildren = false,
+    List<TaxonomyCategoryField> fields = const [],
+    List<Parameter> additionalParameters = const [],
+  }) : super.roots(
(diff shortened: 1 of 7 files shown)


monsieurtanuki (approved):
Thank you @gspencergoog for this clean code, as usual.

Instead of repeating 4 times the new `roots` constructor I would have changed the constructor parameter as a bare `List<String>? tags`. And compute the `_isRootConfiguration` bool from this parameter being `null` or empty. But I admit the choice of `roots` constructors may make things more explicit.

At least, you should probably change the current comment:
```dart
/// The tags to request.
///
/// If empty, no results will be returned. // not 100% true anymore
final List<String> tags;
```

gspencergoog:
> Instead of repeating 4 times the new roots constructor I would have changed the constructor parameter as a bare List<String>? tags.

I did consider that, but I tried to think the way a newbie might think when confronted with the constructors available: if they wanted to get the roots, it wouldn't be obvious that supplying no tags would do that: it's a non-intuitive side effect. Sure, it fits into the existing parameters, but it's just not something I would think of to do if I didn't read all the documentation.

I decided it would be better to be explicit, and so I also considered having a new `bool` argument called `fetchRoots` that was mutually exclusive with the `tags` argument, but that seemed overly complicated, and since Dart has the nice feature of named constructors, it made sense to use a named constructor instead that just didn't take any tags.

Issue #191, opened 2021-07-20
Add an Eco-Score helper


teolemon:
![image](https://user-images.githubusercontent.com/1689815/126279965-a378a16a-4edc-4eae-9341-29400ad80e73.png)
I specced a system to help 3rd party apps ask for relevant Eco-Score data. How about creating the logic within openfoodfacts-dart so that it can be reused easily ?

https://docs.google.com/document/d/1_5AeofpXbaKY9Rd3eeWmHIrhJE8GiPQ-Mfx1SCvpzME/edit#

A similar system for Eco-Score, and another one for Ingredient related analysis (NOVA, allergens, gluten…) would unlock more reuses, and more data for Open Food Facts

monsieurtanuki:
My suggestion would be to:
1. create an `enum` with all the possible message cases (`ProductImprovement`?)
2. for each message case, provide a (default) message in English
1. add a method in `Product` that returns a `List<ProductImprovement>`
2. the developer would then be able to ignore those possible improvements (same as today), to display them with the default English message, or to localize them according to their own user target and their own UI/UX

Same method for #193 of course.

@teolemon Does that make sense?

teolemon:
That looks like a rather good plan. I would also try to mutualize translations, or possibly move them to something on the server (a file the plugin could query ?)

monsieurtanuki:
@teolemon Are you sure about the `'en:labels-to-be-completed'` string?
In the unit tests I could not find a product with something similar. I found something like that for `product.statesTags`:
* en:to-be-completed
* en:nutrition-facts-completed
* en:ingredients-completed
* en:expiration-date-completed
* en:packaging-code-to-be-completed
* en:characteristics-to-be-completed
* en:origins-to-be-completed
* en:categories-completed
* en:brands-completed
* en:packaging-completed
* en:quantity-completed
* en:product-name-completed
* en:photos-to-be-validated
* en:packaging-photo-to-be-selected
* en:nutrition-photo-selected
* en:ingredients-photo-selected
* en:front-photo-selected
* en:photos-uploaded

stephanegigandet:
There's no label to be completed state because some products don't have labels.

monsieurtanuki:
@stephanegigandet I've just removed `ProductImprovement.LABELS_TO_BE_COMPLETED` from #217.

monsieurtanuki:
Product improvements are now computed in `product.getProductImprovements()` (#217).
[...]

PR #288, opened 2021-11-04
feature_2/#191 - added eco-score and nutri-score categories (ProductImprovement)

monsieurtanuki:
Impacted file:
* `Product.dart`

### DIFF
--- a/lib/model/Product.dart
+++ b/lib/model/Product.dart
@@ -60,6 +60,30 @@ enum ProductImprovement {
   OBSOLETE_NUTRITION_IMAGE,
 }
 
+/// Category: what would this [ProductImprovement] help compute?
+enum ProductImprovementCategory {
+  /// Eco-Score category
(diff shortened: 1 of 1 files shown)


M123-dev (approved):
👍🏼

M123-dev:
Just one sidenote @monsieurtanuki, the changelog is going to be created by [release-please](https://github.com/google-github-actions/release-please-action) and it doesn't leave much tolerance for commit names (in my testing, even a capital F at fix: was not accepted), so I assume that such names as this will not be automatically included.

monsieurtanuki:
@M123-dev Oh, I wasn't aware of that. In that case a correct title would have been "feat: #191 - added eco-score and nutri-score categories (ProductImprovement)", am I right?

M123-dev:
@monsieurtanuki according to the dokumentaion it looks right. With a little testing, we will quickly find the right syntax.

PR #294, opened 2021-11-16
fix: Update to latest KnowledgePanel API changes 

jasmeet0817:


M123-dev:
We now create the release with release please, in order for it to add the commit messages to the changelog they need a certain scheme. It would be good if you could add the prefix when merging the PR.

It would be either 
feat: ...
or
fix: ...
and more but I am not sure which are supported. 

For breaking changes a ! does the trick (`fix!:`)

lib/model/KnowledgePanel.dart
```diff
@@ -60,65 +40,57 @@ enum Grade {
 enum Evaluation {
   @JsonValue('good')
   GOOD,
-  @JsonValue('neutral')
-  NEUTRAL,
+  @JsonValue('average')
+  AVERAGE,
   @JsonValue('bad')
   BAD,
   UNKNOWN,
```
stephanegigandet, line 47: why no @JsonValue('unknown') ?

stephanegigandet: right: environment/packaging_recycling.tt.json:                    "evaluation": "neutral",

stephanegigandet (approved):

lib/model/KnowledgePanel.dart
jasmeet0817: what's the difference between "average" and "neutral"? Can't we collapse them into one ? 


Unknown: the object gets a value of enum type UNKNOWN when the json parser encounters a value from the server that isn't mappable to anything else in this enum (specifically if server responds with something other than Good, bad, average, the variable is assigned UNKNOWN).

peterwvj (approved):

peterwvj:
Thanks.

@jasmeet0817 please branch off master before working on your changes. This way we can away cluttering the PR with old commits.

jasmeet0817:
> Thanks.
> 
> @jasmeet0817 please branch off master before working on your changes. This way we can away cluttering the PR with old commits.

Sure thing

Issue #225, opened 2021-09-12
Create github releases for new releases


christianfl:
Hey guys! Just an idea, would love if you "implement" that but could also understand when it doesn't fit in your workflow.

As I'm still slowly developing an Open Source food tracker for Android I'm depending on this dart plugin. I would like to get informed for new releases so that I can test and use them for my app - So I'm subscribing to this and other plugins I'm using (with Watch > Custom > Releases and Security alerts) here on Github.

**Could you maybe think of adding new releases here on Github or is there any other possibility to get informed for new releases? I'm relatively new to Flutter, Dart, pub.dev and so on.**

Thanks in advance.
-Christian

M123-dev:
I think that would be a good idea, but then I would directly suggest that we automate the entire release process. I'll have a look at that

M123-dev:
@christianfl, just a quick reminder we published a new version with some big changes yesterday

christianfl:
@M123-dev Hehe, thank you! This is some first class service. 🥇

M123-dev:
Blocked by #250

PR #312, opened 2021-12-11
fix: Release workflow

M123-dev:
Finally working release workflow + changelog order fix
Finally closes: #225

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,6 +1,6 @@
 name: openfoodfacts
 description: Dart package for the Open Food Facts API, a food products database made by everyone, for everyone.
-# This is automatically taken care of, no need to increase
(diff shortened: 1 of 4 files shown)


pubspec.yaml
```diff
@@ -1,6 +1,7 @@
 name: openfoodfacts
 description: Dart package for the Open Food Facts API, a food products database made by everyone, for everyone.
-version: 1.3.9
+# This is automatically taken care of, no need to increase
```
jasmeet0817, line 3: can you add some more details? Where is it taken care of, if this version line is a no-op, is it possible to live without a version here?

jasmeet0817 (approved):

pubspec.yaml
M123-dev: Updated the comment, is it better now

Issue #307, opened 2021-11-29
Allow for advanced search queries


teolemon:
Ensure the search.pl route is able to handle all of those parameters and values for tag_N
https://openfoodfacts.github.io/api-documentation/#jump-5Filtering-SEARCHparameters

### List of TagTypes
```

brands
categories
packaging
labels
origins
manufacturing_places
emb_codes
purchase_places
stores
countries
additives
allergens
traces
nutrition_grades
states
contains
does_not_contain

  | <option value="ingredients" label="ingredients">ingredients</option>

  | <option value="nova_groups" label="NOVA groups">NOVA groups</option>
  |  
  | <option value="languages" label="languages">languages</option>
  |  
  | <option value="creator" label="contributors">contributors</option>
  |  
  | <option value="editors" label="editors">editors</option>
  |  
  | <option value="lang" label="main language">main language</option>


```

monsieurtanuki:
Done by #313.

PR #313, opened 2021-12-12
feat: #307 - added 21 enum values for advanced search parameters

monsieurtanuki:
Impacted files:
* `api_searchProduct_test.dart`: additional tests on all `TagFilterType`s; refactoring
* `TagFilter.dart`: added more explicit `enum TagFilterType` as filter types

### DIFF
--- a/lib/model/parameter/TagFilter.dart
+++ b/lib/model/parameter/TagFilter.dart
@@ -1,5 +1,58 @@
 import 'package:openfoodfacts/interface/Parameter.dart';
 
+/// Filter types for advanced search parameters
+enum TagFilterType {
+  BRANDS,
+  CATEGORIES,
+  PACKAGING,
+  LABELS,
+  ORIGINS,
+  MANUFACTURING_PLACES,
+  EMB_CODES,
+  PURCHASE_PLACES,
+  STORES,
+  COUNTRIES,
+  ADDITIVES,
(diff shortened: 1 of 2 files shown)


M123-dev (approved):
Looks good, thanks @monsieurtanuki 👍🏼

monsieurtanuki:
Thank you @M123-dev!

monsieurtanuki:
Hey @M123-dev, I've just received an email about version 1.6.0. Isn't that a bit generous to go from 1.4.0 to 1.6.0 in just a couple of `feat:`? Will your PR about country code also move us up to 1.7.0?

M123-dev:
@monsieurtanuki we just released 1.5.0 yesterday, with getProductUri and setProductImageAngle (It was the first working automatic release #312) so it is not a jump. 

But yes a already moving up to 1.6.0 is a bit quick but luckily the version increase only happens when we merge the autorelease PR, that means we can merge as many feats and fixes as we want and the PR stays at 1.6.0, only with a breaking change (feat!: or fix!:) it would increase to 2.0.0.


Same goes for fixes, when we have x fixes merges the version goes only up by 0.0.1

monsieurtanuki:
Ok thank you @M123-dev that's clearer. The version increase makes sense, and the fact that we still have control on when to actually release too (with the autorelease PR).

PR #314, opened 2021-12-12
feat: Allow to get OpenFoodFactsCountry from countryCode

M123-dev:


lib/utils/CountryHelper.dart
```diff
@@ -1003,4 +1003,19 @@ extension OpenFoodFactsCoutryExtension on OpenFoodFactsCountry {
...
+  bool containsCountryCode(String countryCode) =>
+      _ISO_2_CODES.containsValue(countryCode);
+
+  OpenFoodFactsCountry? fromCountryCode(String countryCode) {
+    OpenFoodFactsCountry? country;
+
+    _ISO_2_CODES.forEach((key, value) {
+      if (value == countryCode) {
+        country = key;
+        return;
```
monsieurtanuki, line 1016: The `return` here is misleading I think.

monsieurtanuki (approved):

lib/utils/CountryHelper.dart
M123-dev: It just stops the for loop from running further than it has to, adding a comment

monsieurtanuki: I don't think it does stop the loop. The next loop will be for the next (key, value). Add a print, just for tests, in order to double-check that.

M123-dev: Thats interesting it actually doesn't work, (It works in normal for loops)

Also while testing I found out that the extension doesn't work on the enum itself 

So not `OpenFoodFactsCountry.fromCountryCode`
but `OpenFoodFactsCountry.FRANCE.fromCountryCode` works.

monsieurtanuki: The `return` exits from the anonymous function, that is called for all (key, value)s.
And the method should be `static`.

M123-dev: Just created a new CountryHelper class and put it in there

lib/utils/CountryHelper.dart
```diff
@@ -1004,3 +1004,12 @@ extension OpenFoodFactsCoutryExtension on OpenFoodFactsCountry {
...
   String get iso2Code => _ISO_2_CODES[this]!;
 }
+
+/// Helper class around [OpenFoodFactsCountry]
+class CountryHelper {
+  /// Converts an ISO 2 code into an [OpenFoodFactsCountry]
+  static OpenFoodFactsCountry? fromJson(String? code) =>
+      OpenFoodFactsCountry.values.firstWhere(
+        (final OpenFoodFactsCountry country) => country.iso2Code == code,
+      );
```
monsieurtanuki, line 1014: Does not work for an unknown `code`: crashes, instead of an expected `null`.
`firstWhere` is not adequate, as it never returns null.
Your previous syntax was correct (it's just that the `return;` was useless).

monsieurtanuki (changes requested):

lib/utils/CountryHelper.dart
M123-dev: Yes I know that it theoretically worked but it didn't make sense since the extension only worked for specific values of the enum, not for the enum itself, so there you would have to write `OpenFoodFactsCountry.FRANCE.fromCountryCode('de')` which would return germany, that's why I moved it and now the structure is similar to the one in the OpenFoodFactsLanguage.

Regarding the null problem, it's not possible to return null from `orElse` so we could either add a UNKNOW value to the enum as in the OpenFoodFactsLanguage but I would rather just switch back to the old forEach loop.

monsieurtanuki: I see, `static` methods don't seem to be allowed in `extension`s.

Similarly to what you wrote initially, and which seems to work:
```dart
class CountryHelper {
  static OpenFoodFactsCountry? fromJson(String? code) {
    OpenFoodFactsCountry? result;
    OpenFoodFactsCountry.values.forEach((final OpenFoodFactsCountry key) {
      if (key.iso2Code == code) {
        result = key;
      }
    });
    return result;
  }
}
```

monsieurtanuki: Or even better (I don't find the dart map methods always 100% relevant):
```dart
class CountryHelper {
  static OpenFoodFactsCountry? fromJson(String? code) {
    if (code == null) {
      return null;
    }
    for (final OpenFoodFactsCountry key in OpenFoodFactsCountry.values) {
      if (key.iso2Code == code) {
        return key;
      }
    }
    return null;
  }
}
```

M123-dev: Changed it, good suggestion

monsieurtanuki (approved):
Should be OK now @M123-dev!

M123-dev:
Thanks @monsieurtanuki

PR #316, opened 2021-12-13
use KnowledgePanelElementType.UNKNOWN as default value when a value it's not present

jasmeet0817:


jasmeet0817:
Fixes https://github.com/openfoodfacts/smooth-app/issues/715

teolemon (approved):

Issue #306, opened 2021-11-29
Add Server-side photo unselection 


teolemon:
### What

Add Server-side photo unselection 
https://openfoodfacts.github.io/api-documentation/#jump-4WRITERequests-Deselectingaphoto

### Part of 
- #218

monsieurtanuki:
@teolemon Not very verbose, is it?
![Capture d’écran 2021-12-04 à 15 32 21](https://user-images.githubusercontent.com/11576431/144713265-426abfda-b4e1-4bec-b290-2af8dab87f30.png)

teolemon:
thanks @monsieurtanuki , filed a documentation task at https://github.com/openfoodfacts/api-documentation/issues/49 with the proper route

monsieurtanuki:
It looks like https://world.openfoodfacts.net/cgi/product_image_unselect.pl?code=4250752200784&id=ingredients_de from what I read in
* https://github.com/openfoodfacts/openfoodfacts-server/blob/main/cgi/product_image_unselect.pl
* https://github.com/openfoodfacts/openfoodfacts-server/blob/main/lib/ProductOpener/Images.pm

I'm about to PR.

PR #319, opened 2021-12-14
feat: #306 - unselect product image

monsieurtanuki:
Impacted files:
* `api_addProductImage_test.dart`: new test about "unselect product image"
* `openfoodfacts.dart`: new method `unselectProductImage`
* `ProductImage.dart`: minor unrelated refactoring

### DIFF
--- a/lib/openfoodfacts.dart
+++ b/lib/openfoodfacts.dart
@@ -1022,4 +1022,46 @@ class OpenFoodAPIClient {
         '/' +
         filename;
   }
+
+  /// Unselect a product image.
+  ///
+  /// Typically, after that the openfoodfacts web page will _not_ show
+  /// the image as selected for this product x imagefield x language anymore.
+  /// Throws an exception if not successful.
+  /// Will work OK even when there was no previous selected product image.
+  static Future<void> unselectProductImage({
+    required final String barcode,
+    required final ImageField imageField,
+    required final OpenFoodFactsLanguage language,
+    final QueryType? queryType,
+  }) async {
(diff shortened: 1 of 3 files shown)


lib/openfoodfacts.dart
```diff
@@ -1022,4 +1022,46 @@ class OpenFoodAPIClient {
...
+  }) async {
+    final String id = '${imageField.value}_${language.code}';
+    final Uri uri = UriHelper.getUri(
+      path: 'cgi/product_image_unselect.pl',
+      queryType: queryType,
+      queryParameters: <String, String>{'code': barcode, 'id': id},
+    );
+
+    final Response response = await HttpHelper()
+        .doGetRequest(uri, userAgent: OpenFoodAPIConfiguration.userAgent);
```
M123-dev, line 1046: With a comma after the user agent it is probably a bit more readable.

M123-dev (approved):
Thanks @monsieurtanuki, looks good

lib/openfoodfacts.dart
monsieurtanuki: Good idea. I've just committed a refactoring.
It made me realize that it would probably be better to systematically include a `User user` parameter to methods where we impact the server database (not just read it).

monsieurtanuki: @M123-dev I have added a mandatory `User user` parameter in my recent `setProductImageAngle` and `setProductImageCrop` methods: does that qualify for a breaking change and the corresponding version upgrade? That would be a bit too generous.

PR #323, opened 2021-12-15
Process map elements in knowledge panel

jasmeet0817:


monsieurtanuki (commented):
Hi @jasmeet0817!
Of course `Map` can have different meanings (hashmap or world map), but I imagine there are more relevant names than `KnowledgePanelMapElement`. Neither `Element` sounds nice: I assumed it was a hashmap item.
What about `KnowledgePanelGeoLocations`?
(my 2 cents...)

jasmeet0817:
> Hi @jasmeet0817! Of course `Map` can have different meanings (hashmap or world map), but I imagine there are more relevant names than `KnowledgePanelMapElement`. Neither `Element` sounds nice: I assumed it was a hashmap item. What about `KnowledgePanelGeoLocations`? (my 2 cents...)

That is indeed a good point, but the thing is the element type is `map` so we should keep `Map` somewhere in the name, and also `Element`. `Element` is present as the suffix of all such objects that are specific types of KnowledgePanel elements.

So I changed it to `WorldMap`, also made some other changes, hope they make you happy :)

lib/model/KnowledgePanelElement.dart
```diff
@@ -180,6 +180,57 @@ class KnowledgePanelTableColumn extends JsonObject {
...
+      _$KnowledgePanelWorldMapElementFromJson(json);
+
+  @override
+  Map<String, dynamic> toJson() => _$KnowledgePanelWorldMapElementToJson(this);
+}
+
+/// Element representing a geo location of a map pointer.
+@JsonSerializable()
+class KnowledgePanelGeoPointer extends JsonObject {
+  final KnowledgePanelLatLng? geo;
```
monsieurtanuki, line 202: Can it be `null`? What would be the point, then?

monsieurtanuki (approved):
@jasmeet0817 Feel free to ignore my comments.

lib/model/KnowledgePanelElement.dart
jasmeet0817: yeah Stephane said it could have other information (non-geo). I actually don't know exactly what yet

PR #327, opened 2021-12-17
Convert Country helper Enum to capitalized string

jasmeet0817:


monsieurtanuki (commented):
Hi @M123-dev!
I understand the code and its approximations, but what motivates it? Is there a real use case behind?
Besides, it's not a `name`, at best it's a name in English.
And we could also populate a `Map<OFFCountry, String>` with clean names (in English).

monsieurtanuki:
> Hi @M123-dev!

Oops, I meant @jasmeet0817...

lib/utils/CountryHelper.dart
```diff
@@ -1003,6 +1003,16 @@ extension OpenFoodFactsCoutryExtension on OpenFoodFactsCountry {
...
   String get iso2Code => _ISO_2_CODES[this]!;
+
+  String name() {
+    final String countryName = this
+        .toString()
+        .replaceAll('OpenFoodFactsCountry.', '')
+        .replaceAll('_', ' ');
+    return (countryName.isNotEmpty)
+        ? '${countryName[0].toUpperCase()}${countryName.substring(1).toLowerCase()}'
+        : countryName;
```
M123-dev, line 1014: If name is not empty: name else: name, this doesn't seem logical, does it

jasmeet0817: why not?

jasmeet0817:
> Hi @M123-dev! I understand the code and its approximations, but what motivates it? Is there a real use case behind? Besides, it's not a `name`, at best it's a name in English. And we could also populate a `Map<OFFCountry, String>` with clean names (in English).

renamed to `toSanitizedString`

use case: Onboarding flow: https://github.com/openfoodfacts/smooth-app/issues/669

I like the programatic way much more than manually creating a map. For the following reasons
1. Less tedious
2. Less error prone
3. Changes in enum value could lead to bugs if the map isn't updated.

lib/utils/CountryHelper.dart
```diff
@@ -1004,9 +1004,8 @@ extension OpenFoodFactsCoutryExtension on OpenFoodFactsCountry {
...
   String get iso2Code => _ISO_2_CODES[this]!;
 
-  String name() {
-    final String countryName = this
-        .toString()
+  String toSanitizedString() {
+    final String countryName = toString()
         .replaceAll('OpenFoodFactsCountry.', '')
         .replaceAll('_', ' ');
     return (countryName.isNotEmpty)
```
monsieurtanuki: How could this happen anyway? A country as `enum` with no letters at all?

lib/utils/CountryHelper.dart
jasmeet0817: here, made it easier to read

lib/utils/CountryHelper.dart
```diff
@@ -1005,12 +1005,11 @@ extension OpenFoodFactsCoutryExtension on OpenFoodFactsCountry {
...
   String toSanitizedString() {
-    final String countryName = toString()
-        .replaceAll('OpenFoodFactsCountry.', '')
-        .replaceAll('_', ' ');
-    return (countryName.isNotEmpty)
-        ? '${countryName[0].toUpperCase()}${countryName.substring(1).toLowerCase()}'
-        : countryName;
+    final String countryName =
+        toString().replaceAll('OpenFoodFactsCountry.', '').replaceAll('_', ' ');
+    return (countryName.isEmpty)
```
monsieurtanuki, line 1010: It's never going to happen, will it?

monsieurtanuki (commented):
Now I understand the purpose, but I think it's wrong to do it 1. in off-dart 2. in English.
It is smoothie's problem to select the right language and the right country, not off-dart's.
We use ISO codes for language and country, so that whoever using off-dart can translate any country in any language.
@jasmeet0817 I guess the onboarding in smoothie will be in the user's language, won't it?

jasmeet0817:
> Now I understand the purpose, but I think it's wrong to do it 1. in off-dart 2. in English. It is smoothie's problem to select the right language and the right country, not off-dart's. We use ISO codes for language and country, so that whoever using off-dart can translate any country in any language. @jasmeet0817 I guess the onboarding in smoothie will be in the user's language, won't it?

Actually you're right, I was thinking we can show the countries in English and it shouldn't make that much of a difference, but it will make a difference for other scripts (like Arabic).

How do you suggest we translate all of these? Should I create one variable per language in `app_en.arb`

M123-dev:
I found this; https://github.com/appsah/countries/blob/master/lib/src/data/countries.dart in a pckage called [countries ](https://pub.dev/packages/countries), this could help but also only in english

jasmeet0817:
> I found this; https://github.com/appsah/countries/blob/master/lib/src/data/countries.dart in a pckage called [countries ](https://pub.dev/packages/countries), this could help but also only in english

I can try [this one](https://pub.dev/packages/iso_countries) but have to make sure it's in sync with our list of countries, can be complicated 😅

monsieurtanuki:
> Should I create one variable per country in app_en.arb, sounds like an overkill but is there another way?

If you don't, someone will have to. You may find a package in pub.dev that already does that (as @M123-dev suggested) (what about https://pub.dev/packages/country_codes?), but anyway you'll probably have to create something like a `Map<OFFCountry, String>`, and to handle the critical cases (e.g. `'uk'` and `'gb'`). And other side effects like "hey there's no translation for this country in this language!".

jasmeet0817:
> > Should I create one variable per country in app_en.arb, sounds like an overkill but is there another way?
> 
> If you don't, someone will have to. You may find a package in pub.dev that already does that (as @M123-dev suggested) (what about https://pub.dev/packages/country_codes?), but anyway you'll probably have to create something like a `Map<OFFCountry, String>`, and to handle the critical cases (e.g. `'uk'` and `'gb'`). And other side effects like "hey there's no translation for this country in this language!".

[iso_countries](https://pub.dev/packages/iso_countries) does the trick, renders different countries for different languages. Of the top of both your heads: What other critical cases are there (except UK/GB) in our app that I should take care of ?

monsieurtanuki:
iso_countries does NOT do the trick at 100%:
* it forces us to upgrade the Android version to 21 (from 20)
* it works only on Android and ios (but it looks like I'm the only one who thinks that running on the other platforms is cool)

Beyond that, I guess you need to be flexible: just ignore a country where there's no localized name. Or use your initial conversion method as backup (from OpenFoodFactsCountry to sanitized name as String)
And from the comments I left in `CountryHelper.dart`, I assume that `gb`vs `uk` should be the only issue.

jasmeet0817:
> iso_countries does NOT do the trick at 100%:
> 
> * it forces us to upgrade the Android version to 21 (from 20)
This is true, is it a big problem?
> * it works only on Android and ios (but it looks like I'm the only one who thinks that running on the other platforms is cool)
This, I don't know, I know that the website is not going away anytime soon, so at least in the mid-term we shouldn't base any decisions for this. By the time we would like to deploy on more surfaces, a lot would have changed.

> 
> Beyond that, I guess you need to be flexible: just ignore a country where there's no localized name. Or use your initial conversion method as backup (from OpenFoodFactsCountry to sanitized name as String) And from the comments I left in `CountryHelper.dart`, I assume that `gb`vs `uk` should be the only issue.

Cool, thanks.

monsieurtanuki:
@jasmeet0817 I've just double-checked: in both cases (Android API 20 and 21) the target is the same, 98%. Therefore we can say it's not a problem.

My suggestions:
* use [`iso_countries`](https://pub.dev/packages/iso_countries) in smoothie
* careful with `gb` / `uk`.
* the list of countries being `OpenFoodFactsCountry`, you may not find all countries or all translations in `iso_countries` - the fallback translation should be your initial `toSanitizedString` method
* but this has nothing to do with openfoodfacts-dart, it's just about smoothie, therefore you should close this current PR and write all the code in smoothie

jasmeet0817:
> @jasmeet0817 I've just double-checked: in both cases (Android API 20 and 21) the target is the same, 98%. Therefore we can say it's not a problem.
> 
> My suggestions:
> 
> * use [`iso_countries`](https://pub.dev/packages/iso_countries) in smoothie
> * careful with `gb` / `uk`.
> * the list of countries being `OpenFoodFactsCountry`, you may not find all countries or all translations in `iso_countries` - the fallback translation should be your initial `toSanitizedString` method
> * but this has nothing to do with openfoodfacts-dart, it's just about smoothie, therefore you should close this current PR and write all the code in smoothie

Agreed on all points, thanks for the discussion.

PR #329, opened 2021-12-17
feat: Added globalUser to OpenFoodAPIConfiguration

M123-dev:


lib/utils/OpenFoodAPIConfiguration.dart
```diff
@@ -13,6 +14,9 @@ class OpenFoodAPIConfiguration {
   ///Defines a global userAgent to tell the backend the source of the request.
   static UserAgent? userAgent;
 
+  ///Defines a global user to avoid adding it to every request
+  static User? globalUser;
```
jasmeet0817, line 18: what is this initialized with ?

M123-dev: `Null`, we don't want to pass a user if there is non provided

jasmeet0817: but I don't see `globalUser` getting any value, will it always be `Null`? Sorry if I'm missing something.

M123-dev: The OpenFoodAPIConfiguration class is a kind of settings, that means the individual values can be set once at the start of the app and they are then automatically added to each requests. This includes things like the UserAgent or the individual url parts so that they don't have to be added to each request.

jasmeet0817: Ah I see. In that case, I would advice making these private members and setting them via a constructor since you have methods for accessing them anyways.

The issue that can happen in current code is someone accesses `globalUser` directly instead of calling `getUser()` method

M123-dev: Thats actually not needed here, they are allowed to be directly accessed for setting them and (maybe if someone wants to get them). We also allow to pass some of these configuration options via the method itself and we needed some kind of hirach. 

Most of my requests should go under the main account
             - But this special account should be send by a different user Account

And the method is only to simplify this process

lib/utils/HttpHelper.dart
```diff
@@ -91,7 +91,7 @@ class HttpHelper {
 
     request.headers.addAll(
       _buildHeaders(
-        user: user,
+        user: OpenFoodAPIConfiguration.getUser(user),
```
monsieurtanuki, line 94: You should simplify the code and call `OpenFoodAPIConfiguration.getUser(user)` only once: inside `_buildHeaders`. Does that make sense?

lib/utils/OpenFoodAPIConfiguration.dart
jasmeet0817: ok sounds good.

lib/utils/HttpHelper.dart
M123-dev: Me: Ignore the thing I am doing at the same time for the userAgent, 

Fixed

monsieurtanuki (approved):

Issue #326, opened 2021-12-16
There is a parameter `serving_quantity` with no corresponding `ProductField`


wu-lee:
**Is your feature request related to a problem? Please describe.**

I want to access the `serving_quantity` parameter, which I believe is a numeric value in grams. It is to `serving_size` (a String) what `product_quantity` is to `quantity`: a parsed and normalised numeric version of the string.

However, I can't without being able to request it via an appropriate `ProductField` set in the query, and there isn't one (I checked the source code too).

There also seems to be no way to select only products which have this value set, as there is for `product_quantity`.

**Describe the solution you'd like**

Please add this field so this value can be retrieved. Please also add a `states` tag which allows me to set a TagFilter to select only products which have this value defined.

**Describe alternatives you've considered**

I suppose I could abandon the Dart library and roll my own HTTP based client?  But that seems counter-productive, I'm sure OFF would prefer me to use this library.

Possibly I could try and parse and normalise the string `serving_size` myself, but this seems hazardous, and not in keeping with the intent of this library.

**Additional context**

I am using the library version 1.3.8 on Android. I've checked the source of 1.7.0 and these features appear to be missing there too.

wu-lee:
Just to add that I've tried setting `ProductField.SERVING_SIZE` - this results in `Product.servingSize` being set, but not `Product.servingQuantity`.

As an alternative to a new `ProductField.SERVING_QUANTITY` field as suggested above, possibly it could be that `ProductField.SERVING_SIZE` in the request indicates that both the `servingSize` and `servingQuantity` values should be supplied?

wu-lee:
A hacky workaround I'm employing, a delegate class, which at least demonstrates that the API supports the `serving_quantity` field, independently of `serving_size`:
```
/// This is a hack to work around lack of a [ProductField] value for `serving_quantity`
///
/// See https://github.com/openfoodfacts/openfoodfacts-dart/issues/326
class WorkaroundProductSearchQueryConfiguration implements ProductSearchQueryConfiguration {
  final ProductSearchQueryConfiguration delegate;

  WorkaroundProductSearchQueryConfiguration(this.delegate);


  Map<String, String> getParametersMap() {
    final params = delegate.getParametersMap();
    params["fields"] = params["fields"]! + ",serving_quantity";
    print(">>> $params");
    return params;
  }

  @override
  List<Parameter> get additionalParameters => delegate.additionalParameters;
  set additionalParameters(List<Parameter> val) {
[...]

Issue #291, opened 2021-11-13
Confusion with Product field priorities and read/write statuses


monsieurtanuki:
**Describe the bug**
When you save a product
* some product fields seem to have priority over others (e.g. `productNameInLanguages` being more important than `productName`)
* some product fields seem to be computed from other fields and cannot be set directly (e.g. `ingredientsTags` computed from `ingredientsText`)

The problem is not that such priority rules exist, the problem is that they are not explicitly available.

**To Reproduce**
I wrote a test code; as comments the values you get after the save+get operation:
```dart
const String barcode = '7340011364184';
final Product product = Product(
  barcode: barcode,
  // will be [en:product-categories-test-3, en:product-categories-test-4]
  categoriesTags: [
    'en:product-categories-test-1',
    'en:product-categories-test-2'
  ],
  // will be Product categories test 3,Product categories test 4
  categories: 'Product categories test 3,Product categories test 4',
  // will be null
  categoriesTagsInLanguages: {
    OpenFoodFactsLanguage.ENGLISH: [
      'Product categories test 5',
      'Product categories test 6',
    ],
  },
  // will be Product labels test 1,Product labels test 2
  labels: 'Product labels test 1,Product labels test 2',
  // will be [en:product-labels-test-1, en:product-labels-test-2]
  labelsTags: [
    'en:product-labels-test-3',
    'en:product-labels-test-4',
  ],
  // will be null
  labelsTagsInLanguages: {
    OpenFoodFactsLanguage.ENGLISH: [
      'Product labels test 5',
      'Product labels test 6',
    ],
  },
  // will be Italy,Japan
  countries: 'Italy,Japan',
  // will be [en:italy, en:japan]
  countriesTags: [
    'en:france',
    'en:portugal',
  ],
  // will be null
  countriesTagsInLanguages: {
    OpenFoodFactsLanguage.ENGLISH: [
      'Cyprus',
      'Hungary',
    ],
  },
  // will be Toto2
  productName: 'Toto1',
  // will be {OpenFoodFactsLanguage.ENGLISH: Toto2}
  productNameInLanguages: {OpenFoodFactsLanguage.ENGLISH: 'Toto2'},
  // will be beer,lemonade
  ingredientsText: 'sugar,milk',
  // will be {OpenFoodFactsLanguage.ENGLISH: beer,lemonade}
  ingredientsTextInLanguages: {
    OpenFoodFactsLanguage.ENGLISH: 'beer,lemonade',
[...]

PR #331, opened 2021-12-17
Fix: Added missing product fields to ProductField enum

M123-dev:
Fixes: #326

Transfered missing values of Product and ProductField enum


lib/model/Product.dart
```diff
@@ -102,6 +103,9 @@ class Product extends JsonObject {
       includeIfNull: false)
   Map<OpenFoodFactsLanguage, String>? productNameInLanguages;
 
+  @JsonKey(name: 'generic_name')
+  String? genericName;
+
```
monsieurtanuki, line 108: What's that?

lib/utils/ProductFields.dart
```diff
@@ -101,15 +108,20 @@ extension ProductFieldExtension on ProductField {
     ProductField.CATEGORIES: 'categories',
     ProductField.CATEGORIES_TAGS: 'categories_tags',
     ProductField.CATEGORIES_TAGS_IN_LANGUAGES: 'categories_tags_',
+    ProductField.LABELS: 'labels',
```
monsieurtanuki, line 111: Probably would be worth checking in a test that when we ask for those fields in a get product query, we actually get them.

monsieurtanuki (commented):
@M123-dev Please have a look at my comments.

lib/model/Product.dart
M123-dev: I checked the diffs and this was in the ProductFields but not in the Product, Its in the normal reponses and since its particially already in the code I just added it

monsieurtanuki: Fair enough!

Nothing personal, but there are already tons of fields in `Product` (50+) and none of them has comments.
Adding a `String?` called `genericName`, why not, next one is going to be `dynamic justSomething`?

More seriously: we should definitely add comments on the `Product` fields. In another PR.

monsieurtanuki (changes requested):
@M123-dev Please add a couple of tests around the initial issue #326:
* asking for field serving quantity in a getProduct, and getting serving quantity
* asking for all fields in a getProduct, and getting serving quantity

And we're done :)

lib/model/Product.dart
M123-dev: Yes there I agree with you, we should make in any case. In fact, I've been thinking about updating the entire documentation a bit for a while now. 

 I don't know how to approach this, but it would also be good to somehow manage to have the products in just one place and not in two files in two places each.

lib/utils/ProductFields.dart
M123-dev: On it

M123-dev:
@monsieurtanuki I added a test but this does not include serving_quantity as this can be retrieved (manually checked) individually and with everything, but not manually set as I found out after way too long research. I left it completely out so that the tests don't start to fail so often.

monsieurtanuki (approved):
Thank you @M123-dev!
Still, I would have liked an additional test for `serving_quantity` (read mode).
But that's OK for the moment I guess: I'm adding comments to #291.

lib/model/Product.dart
```diff
@@ -102,6 +105,11 @@ class Product extends JsonObject {
       includeIfNull: false)
   Map<OpenFoodFactsLanguage, String>? productNameInLanguages;
 
+  ///Common name
+  ///Example: Chocolate bar with milk and hazelnuts
+  @JsonKey(name: 'generic_name', includeIfNull: false)
+  String? genericName;
+
```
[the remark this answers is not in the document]
M123-dev: Just looked into a response and it has generic_name, generic_name_fr as well as a generic_name_fr_debug_tags list, interesting

PR #335, opened 2021-12-20
feat: #326 - refreshed Product.g.dart

monsieurtanuki:
@M123-dev It looks like you forgot to run locally `dart run build_runner build` (or you did run it, but too early).


monsieurtanuki:
@M123-dev Actually #336 does it even better.

PR #343, opened 2021-12-28
fix: Lowercased country code

monsieurtanuki:
Impacted file:
* `CountryHelper.dart`: lowercased the country code for conversions


peterwvj (approved):
Thanks!

Issue #350, opened 2021-12-31
Add support for product/score readiness support


teolemon:
### What
- Add support for score readiness support
### Why
- to support that kind of features, without building an ad-hoc backend.
![image](https://user-images.githubusercontent.com/1689815/147817132-2f25fe5c-cb26-46d9-a079-f8862e73eebf.png)

### How
- A simplistic way for apps to monitor whether the product is ready would be to check for states at every app opening/daily/whatever. 
- They would call a new method with the product barcode (and potentially scores): is_product_ready(ecoscore), is_product_ready(ecoscore,nutriscore), is_product_ready(ecoscore,nutriscore, ingredients)
- When Nutri-Score, Eco-Score and ingredients are computed, the method would return a specific value `product_ready=true`
- The app would then decide how to surface it to the user

### Note
- This is a bit close to the Improvements that was added to the app a while ago, but it is assuming more passive users, and the absence of a self-service mobile contribution pipeline in the app.

monsieurtanuki:
### Smoothie ###
That would mean doing that only for a implicitly restricted list of products, like the products I created on my smartphone, right? Or the products where I ticked the "alert me" checkbox (unticked by default). We can't do that on 1K+ products each time we're opening the app.

### off-dart ###
* The input would be a list of barcodes
* The output would be a `Map<barcode, ScoreReadiness>`, where `ScoreReadiness` would be a new simple class with three boolean fields like `bool ecoscoreReady; bool nutriscoreReady; bool ingredientsReady; `

monsieurtanuki:
Done in #351: now it's time for smoothie to work on it!

PR #351, opened 2021-12-31
feat: #350 - new method getProductFreshness

monsieurtanuki:
New file:
* `ProductFreshness.dart`: Freshness of a [Product].

Impacted files:
* `api_searchProduct_test.dart`: new test about "product freshness"
* `openfoodfacts.dart`: new method `getProductFreshness`

### DIFF
--- a/lib/model/ProductFreshness.dart
+++ b/lib/model/ProductFreshness.dart
@@ -1,0 +1,32 @@
+import 'package:openfoodfacts/model/Product.dart';
+
+/// Freshness of a [Product].
+class ProductFreshness {
+  ProductFreshness._({
+    required this.isEcoscoreReady,
+    required this.isNutriscoreReady,
+    required this.isIngredientsReady,
+  });
+
+  final bool isEcoscoreReady;
+  final bool isNutriscoreReady;
+  final bool isIngredientsReady;
+
+  factory ProductFreshness.fromProduct(final Product product) {
+    return ProductFreshness._(
+      isEcoscoreReady: product.ecoscoreScore != null,
+      isNutriscoreReady: product.nutriscore != null,
+      isIngredientsReady: product.ingredientsTags != null &&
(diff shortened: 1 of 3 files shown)


teolemon:
Can we reserve the freshness namespace for this ?
https://github.com/openfoodfacts/openfoodfacts-server/issues/5674
Score readiness or something instead ?

monsieurtanuki:
I'm not against changing the name.

My point was to create a class (`ProductFreshness`) where we could put several different freshness or readiness fields.
Currently:
* bool isEcoscoreReady
* bool isNutriscoreReady
* bool isIngredientsReady

And maybe later
* Etag
* int lastModified

The goal being to provide a method that helps to quickly monitor the state of a product: how complete it is, and how fresh the data is.
But that's perhaps a bit redundant with the "states".
`ProductCompletionState`?

teolemon:
- So first, Happy New Year :) 
- If we have all of that in mind, I guess it's good. In addition to product states, we also have the Product Improvements system (https://github.com/openfoodfacts/smooth-app/issues/811)

monsieurtanuki:
Happy New Year!
@teolemon You're right, that sounds a bit like `ProductImprovement`.
We'd be better off merging that in a single class:
* bool isEcoscoreReady
* bool isNutriscoreReady
* bool isIngredientsReady
* Etag
* int lastModified
* bool ORIGINS_TO_BE_COMPLETED
* bool CATEGORIES_BUT_NO_NUTRISCORE
* bool ADD_NUTRITION_FACTS
* bool ADD_CATEGORY
* bool ADD_NUTRITION_FACTS_AND_CATEGORY
* bool OBSOLETE_NUTRITION_IMAGE

Does that make sense?

teolemon:
it does

monsieurtanuki:
@teolemon ping

teolemon (approved):

PR #362, opened 2022-01-11
feat: new method "getOrderedNutrientsJsonString"

monsieurtanuki:
The point (for apps) is to retrieve that string, store it locally, and refresh it from the API when relevant (and not systematically).

Impacted files:
* `openfoodfacts.dart`: new method `getOrderedNutrientsJsonString`; refactored `getOrderedNutrients`
* `ordered_nutrient_test.dart`: added test for new method `getOrderedNutrientsJsonString`


M123-dev (approved):
Somehow missed this, but looks good, thanks @monsieurtanuki

monsieurtanuki:
Thank you @M123-dev!

PR #366, opened 2022-01-16
fix: Rename enum so it doesn't conflict with Size class from geometry.dart

jasmeet0817:


teolemon (approved):

monsieurtanuki (commented):
@M123-dev Still upgrading my computer, that's why you didn't hear from me these days.
Not convinced it's a good idea to rename an `enum` that is already used by other developers.
Unless you create a major version - and I guess this PR by itself does not deserve a switch to 2.0.0
Feel free to merge anyway.

jasmeet0817:
This is unused, I just added this enum 2 days ago and it broke Smoothie.

M123-dev:
@monsieurtanuki you probably didn't mean me but anyways have fun with your new pc

M123-dev:
But regarding the major release, maybe we don't have any new features here until we are done with the refactoring, so we can add it to v2 as well. I'd still put my focus on smoothie

jasmeet0817:
Not sure what you guys  are talking about, I just created 1.11.1 :smile:

monsieurtanuki:
@M123-dev It's not a new pc, it's an old mac that I painfully upgrade. In the meanwhile I cannot work on smoothie or off, which explains my comment - which was not specifically directed at you but was in my first comment in days.

Regarding 2.0.0, I don't think we need a major feature - what would that be anyway?
The breaking change would be new files (e.g. one for robotoff) and parameter refactoring (e.g. double-check that we use similar parameters for similar methods).

PR #371, opened 2022-01-23
feat: new method getProductString, api V0 and V2, new product and ingredient fields

monsieurtanuki:
Impacted files:
* `api_getProduct_test.dart`: refactored the ecoscore html test; added a knowledge panels test; added a group in order to test ingredients in API version V0 and V2
* `Ingredient.dart`: added optional fields `percentEstimate` and `ingredients`
* `Ingredient.g.dart`: generated
* `KnowledgePanels.dart`: added static methods in order to be considered as a `Product` field
* `openfoodfacts.dart`: new method `getProductString`, used from methods `getProductRaw`, `getProduct` and now deprecated methods  `getEcoscoreHtmlDescription`, `getKnowledgePanels`
* `Product.dart`: added optional fields `knowledgePanels` and `environmentInfoCard`
* `Product.g.dart`: generated
* `ProductFields.dart`: added fields about `knowledgePanels` and `environmentInfoCard`
* `ProductQueryConfiguration.dart`: added an API version parameter as enum `ProductQueryVersion` and a `getUri` helper method

### DIFF
--- a/lib/utils/ProductQueryConfigurations.dart
+++ b/lib/utils/ProductQueryConfigurations.dart
@@ -2,15 +2,24 @@ import 'package:openfoodfacts/utils/AbstractQueryConfiguration.dart';
 import 'package:openfoodfacts/utils/CountryHelper.dart';
 import 'package:openfoodfacts/utils/LanguageHelper.dart';
 import 'package:openfoodfacts/utils/ProductFields.dart';
+import 'package:openfoodfacts/utils/QueryType.dart';
+import 'package:openfoodfacts/utils/UriHelper.dart';
+
+enum ProductQueryVersion {
+  V0,
+  V2,
+}
 
 /// Query Configuration for single barcode
 class ProductQueryConfiguration extends AbstractQueryConfiguration {
   final String barcode;
+  final ProductQueryVersion version;
 
   /// See [AbstractQueryConfiguration.languages] for
   /// parameter's description.
   ProductQueryConfiguration(
     this.barcode, {
+    this.version = ProductQueryVersion.V0,
     final OpenFoodFactsLanguage? language,
     final List<OpenFoodFactsLanguage> languages = const [],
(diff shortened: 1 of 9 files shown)


lib/utils/ProductQueryConfigurations.dart
```diff
@@ -2,15 +2,24 @@ import 'package:openfoodfacts/utils/AbstractQueryConfiguration.dart';
 import 'package:openfoodfacts/utils/CountryHelper.dart';
 import 'package:openfoodfacts/utils/LanguageHelper.dart';
 import 'package:openfoodfacts/utils/ProductFields.dart';
+import 'package:openfoodfacts/utils/QueryType.dart';
+import 'package:openfoodfacts/utils/UriHelper.dart';
+
+enum ProductQueryVersion {
+  V0,
+  V2,
```
M123-dev, line 10: We should somewhere document what the difference between those two are. 

I don't even know what separates them

test/api_getProduct_test.dart
```diff
@@ -1534,20 +1534,124 @@ void main() {
     });
 
     test('get ecoscore html description', () async {
```
M123-dev, line 1536: This needs to be renamed

M123-dev (approved):

lib/utils/ProductQueryConfigurations.dart
monsieurtanuki: Good point.
I don't know either the difference - except that ingredients look better on V2 (tree instead of flat structure), and that knowledge panels used V2.

monsieurtanuki:

lib/utils/ProductQueryConfigurations.dart
teolemon: Actually v2048 will work 🙃
There is some documentation in the official reference, but it might not be complete. I remember that the products route is a major one, and possibly ingredients nesting

Issue #348, opened 2021-12-30
Add a method to return a localized taxonomy translation url


teolemon:
### What
- [ ] Add a method to get localized taxonomy translation urls
  - Structure: https://world-fr.openfoodfacts.org/categories?translate=1
  - https://world-LC.openfoodfacts.org/TAXONOMY_NAME?translate=1
- [ ] Add a method to get localized Crowdin URL for Open Food Facts
  - https://crowdin.com/project/openfoodfacts/fr
### Why
- We have a section in the app on how to contribute to the project, but there's no way to help with taxonomy translation, which is important.

monsieurtanuki:
Done in #372.

PR #372, opened 2022-01-23
feat: #348 - new methods getTaxonomyTranslationUri and getCrowdinUri

monsieurtanuki:
@teolemon the URI does not work for EMB_CODES

Impacted files:
* `api_getProduct_test.dart`: tests for new methods `getTaxonomyTranslationUri` and `getCrowdinUri`
* `openfoodfacts.dart`: new methods `getTaxonomyTranslationUri` and `getCrowdinUri`

### DIFF
--- a/lib/openfoodfacts.dart
+++ b/lib/openfoodfacts.dart
@@ -244,6 +244,37 @@ class OpenFoodAPIClient {
     );
   }
 
+  /// Returns the URI to the translation page for a taxonomy.
+  ///
+  /// If the target website supports different domains for country + language,
+  /// [replaceSubdomain] should be set to true.
+  static Uri getTaxonomyTranslationUri(
+    final TagType taxonomyTagType, {
+    required final OpenFoodFactsLanguage language,
+    final OpenFoodFactsCountry? country,
+    final QueryType? queryType,
+    required final bool replaceSubdomain,
+  }) {
+    // TODO(teolemon): does not work for emb_codes - what is the URI for them?
+    final Uri uri = UriHelper.getUri(
+      path: taxonomyTagType.key,
+      queryType: queryType,
+      queryParameters: {'translate': '1'},
+    );
+    if (!replaceSubdomain) {
+      return uri;
+    }
+    return UriHelper.replaceSubdomain(
+      uri,
+      language: language,
+      country: country,
+    );
+  }
+
+  /// Returns the URI to the crowdin page for a [language].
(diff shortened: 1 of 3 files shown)


teolemon:
Traceability codes (how I renamed the superset of EMB_CODES) is normalized by replacing everything that looks like a point or a space by -
https://world.openfoodfacts.org/packager-codes
Not sure if the API returns a normalized version.
https://world.openfoodfacts.org/packager-code/fr-72-264-002-ce
https://world.openfoodfacts.org/packager-code/fr-72-264-002-ce.json

teolemon:
- Already the Taxonomy & Crowdin are great. 
- Eventually, being able to show facets in the apps (like emb_codes) would be nice, but it might be another PR

lib/openfoodfacts.dart
```diff
@@ -244,6 +244,37 @@ class OpenFoodAPIClient {
...
+    return UriHelper.replaceSubdomain(
+      uri,
+      language: language,
+      country: country,
+    );
+  }
+
+  /// Returns the URI to the crowdin page for a [language].
+  static Uri getCrowdinUri(final OpenFoodFactsLanguage language) =>
+      Uri.parse('https://crowdin.com/project/openfoodfacts/${language.code}');
```
M123-dev, line 276: Does this also work for GB and EN

lib/openfoodfacts.dart
```diff
@@ -244,6 +244,37 @@ class OpenFoodAPIClient {
...
+  /// Returns the URI to the translation page for a taxonomy.
+  ///
+  /// If the target website supports different domains for country + language,
+  /// [replaceSubdomain] should be set to true.
+  static Uri getTaxonomyTranslationUri(
+    final TagType taxonomyTagType, {
+    required final OpenFoodFactsLanguage language,
+    final OpenFoodFactsCountry? country,
+    final QueryType? queryType,
+    required final bool replaceSubdomain,
```
M123-dev, line 256: What about giving it an initial value and not requiring it

M123-dev (approved):

monsieurtanuki:
Thank you @M123-dev and @teolemon for your review!
* Regarding EMB_CODES, now I throw an exception in `getTaxonomyTranslation` because there's no needed translation for them (but as it's a `TagType` "taxonomy" it can be an input parameter)
* Regarding setting the country for `getTaxonomyTranslation`, it does not seem to make much sense - I removed that parameter (and therefore it will always be `world-$countryCode`)

Issue #370, opened 2022-01-22
Implement the new App uuid system


teolemon:
### What
- [x] Implement the new App uuid system
  - [x] app_name
  - [x] app_version
  - [x] app_uuid
- [ ] Deprecate the equivalent using comment=

For further details and potential questions, see: https://github.com/openfoodfacts/openfoodfacts-server/pull/6319

### Part of 
- #218

monsieurtanuki:
@teolemon Basically it's about adding 3 optional `String` parameters to API queries, but will that concern _all_ queries, or just the writes or just the reads?

M123-dev:
I think we can edit the UserAgent to be used for that

teolemon:
- It's just the `WRITE` queries at the moment, and you can pass it in the URL, especially for apps that can't send a UserAgent. 
- It's also probably optional, especially since compulsory would create backwards compatibility issues. @stephanegigandet might have a stronger opinion.

M123-dev:
What apps can't send a UserAgent, the current UserAgent in this sdk is pretty similar, it's only lacking the uuid and is probably send different

https://github.com/openfoodfacts/openfoodfacts-dart/blob/master/lib/model/UserAgent.dart



  `final String? name;
  final String? version;
  final String? system;
  final String? url;
  final String? comment;`

stephanegigandet:
I agree with @M123-dev , the info is the same as what we have in the UserAgent class, we can reuse it, so we don't need a separate SourceConfiguration class as proposed by @monsieurtanuki in his PR.

A few points:

1. Apps should send both the User-Agent header (as we do today) and the extra app_name and app_version parameters (as GET or POST parameters, but they should not be mixed: either send all parameters as GET query string parameters, or all parameters as POST).

The reason we want both is that we can see them in different places. The User-Agent header is easy to find in nginx and Apache logs, while the more structured app_name and app_version are better for the OFF API internals.

2. The UUID is kind of separate, we can deal with it somewhere else than in the User-Agent class (e.g. when we build the query parameters and add the User info). Maybe AbstractQueryConfiguration

3. This user-agent and the app_name + app_version should be sent on every request: not only the save product, but also image uploads and crops/selection, and queries to read and search products as well.

4. In addition, the user specific username, or UUID should be sent for save product, uploading image and image selection (anything that changes a product).

stephanegigandet:
[...]

PR #375, opened 2022-01-23
feat: #370 - new source configuration for saveProduct

monsieurtanuki:
New file:
* `SourceConfiguration.dart`: Source Configuration for detection of apps (populates data sources on BE).

Impacted file:
* `openfoodfacts.dart`: added source configuration parameter to `saveProduct`


M123-dev:
Some open questions:

Do we also need this for other, for example, edit requests? If so, let's add it to the configuration 

What is the difference to the user agent @teolemon 


Also, it's probably good to add some comments on what the uuid is.

monsieurtanuki:
@M123-dev I agree with you, but my point here is just to provide the tools to add source parameters to the save product query.
That's a start. Other methods will follow, and examples would be nice indeed.

PR #378, opened 2022-02-02
feat: #370 - added "uuid" to configuration

monsieurtanuki:
Impacted files:
* `configuration_test.dart`: added tests with "uuid"
* `HttpHelper.dart`: included new static `uuid` to parameters
* `OpenFoodAPIConfiguration.dart`: added `static String? uuid`
* `pubspec.yaml`: unrelated upgrade

### DIFF
--- a/lib/utils/HttpHelper.dart
+++ b/lib/utils/HttpHelper.dart
@@ -33,16 +33,17 @@ class HttpHelper {
   static Map<String, String>? addUserAgentParameters(
     Map<String, String>? map,
   ) {
-    if (OpenFoodAPIConfiguration.userAgent == null) {
-      return map;
-    }
-    if (OpenFoodAPIConfiguration.userAgent!.name != null) {
+    if (OpenFoodAPIConfiguration.userAgent?.name != null) {
       map ??= <String, String>{};
       map['app_name'] = OpenFoodAPIConfiguration.userAgent!.name!;
     }
-    if (OpenFoodAPIConfiguration.userAgent!.version != null) {
+    if (OpenFoodAPIConfiguration.userAgent?.version != null) {
       map ??= <String, String>{};
       map['app_version'] = OpenFoodAPIConfiguration.userAgent!.version!;
+    }
+    if (OpenFoodAPIConfiguration.uuid != null) {
+      map ??= <String, String>{};
+      map['app_uuid'] = OpenFoodAPIConfiguration.uuid!;
     }
     return map;
   }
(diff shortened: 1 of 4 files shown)


M123-dev (approved):
Looks good thanks

monsieurtanuki:
Now the trick is to tell developers to use user agent and uuid, but that's beyond my coding abilities ;)

MohamedFBoussaid:
@monsieurtanuki thank you for the changes. Maybe updating the readMe and the example file can help ? 
So by default when someone will start coding with OFF package will coy/paste the example with the uuid

M123-dev:
That's something we can add to https://github.com/openfoodfacts/openfoodfacts-dart/projects/2

monsieurtanuki:
Fair enough @MohamedFBoussaid, but I don't know
1. which values can be set - [strict UUID format](https://en.wikipedia.org/wiki/Universally_unique_identifier#Format) or just a String?
=> probably just a String
2. how the developers get a UUID - they just import whatever uuid package or they call an OFF dedicated API, as @jasmeet0817 feared that UUID may not be unique if generated locally (if I remember well)
=> probably locally with a package - cf. https://github.com/openfoodfacts/openfoodfacts-dart/issues/370#issuecomment-1021659401
3. when the UUID must be set
=> probably at the very first init of the app, and saved in preferences

It's looked so far like uuid should be populated by a device id, but that's not supported by (agnostic) dart. Only by flutter.

MohamedFBoussaid:
@monsieurtanuki I can give it a try with my Flutter app, and see how it will work. And then I will be back with examples 

Future<String?> _getId() async {
  var deviceInfo = DeviceInfoPlugin();
  if (Platform.isIOS) { // import 'dart:io'
    var iosDeviceInfo = await deviceInfo.iosInfo;
    return iosDeviceInfo.identifierForVendor; // unique ID on iOS
  } else {
    var androidDeviceInfo = await deviceInfo.androidInfo;
    return androidDeviceInfo.androidId; // unique ID on Android
  }
}

MohamedFBoussaid:
Thanks @monsieurtanuki I will use that package.

MohamedFBoussaid:
@monsieurtanuki , I have made a quick test using that package. And normally I made a fake update for a product 7613035989535 with the following UUID : 1723BE2C-C0EB-4055-8960-5534AFBFB63F. 
@stephanegigandet @teolemon is there a way to check if the UUID was received on the backend ?

monsieurtanuki:
I've just added uuid to smoothie too (https://github.com/openfoodfacts/smooth-app/pull/1064)

PR #384, opened 2022-02-12
fix: successful unit tests

monsieurtanuki:
Impacted files:
* `api_getProduct_test.dart`: fixed a value because of new packaging algo
* `api_getProductRaw_test.dart`: fixed a value that slightly changed
* `api_saveProduct_test.dart`: fixed a value because of new packaging algo
* `api_searchProducts_test.dart`: specific test for packaging filter; refactored
* `Product.dart`: fixed an ambiguous way of coding (that used to work, strangely)

### What
The idea is to have successful (and reliable) unit tests.

### DIFF
--- a/lib/model/Product.dart
+++ b/lib/model/Product.dart
@@ -526,7 +526,7 @@ class Product extends JsonObject {
     final dynamic langs,
   ) {
     Map<OpenFoodFactsLanguage, String>? result;
-    if (!langs is Map<String, dynamic>) {
+    if (langs is! Map<String, dynamic>) {
       return result;
     }
     for (final langValuePair in langs.entries) {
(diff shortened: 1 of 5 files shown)


M123-dev (approved):
👍🏼 

Lets open a issue on the server and ask for a fully filled example product with barcode: test or something like that

M123-dev:
@monsieurtanuki  https://github.com/openfoodfacts/openfoodfacts-server/issues/6411

monsieurtanuki:
Beyond the recurring problems of fluctuating data, the current PR was about the new packaging algorithm on the server side, and a mistake of mine 7 days ago - that cause the tests to crash but that I probably dismissed because we're too used to failed tests.

PR #385, opened 2022-02-14
feat: added the set of nutrient ids supported by Nutriments.

monsieurtanuki:
Impacted file:
* `Nutriments.dart`: added the set of nutrient ids supported by `Nutriments`.

### What
To be used when another source of nutrients (e.g. `OrderedNutrient`) has a broader list of nutrients - that we simply could not handle with `Nutriments`.


peterwvj (approved):
Thanks!

PR #386, opened 2022-02-15
feat: #349 - added support for Events API

### LINKED ISSUE #349: Add support for the openfoodfacts-events API (leaderboard, badge and gamification system)
### What
- [ ] Add support for the badge and event system
  - Source: https://github.com/openfoodfacts/openfoodfacts-events
  - API documentation: https://events.openfoodfacts.org/docs / https://events.openfoodfacts.net/docs
### Why
 - Will unblock openfoodfacts/smooth-app/issues/20

monsieurtanuki:
Still missing: couldn't test how to create an event. Are we talking about the same users as in the "official OpenFoodFacts" API? (aka "off off"). Good news is that the TEST env works - which is not the case for Folksonomy.

New files:
* `api_events_test.dart`: Tests around Events API.
* `BadgeBase.dart`: Events API: badge.
* `BadgeBase.g.dart`: generated
* `EventCreate.dart`: Events API: event create.
* `EventCreate.g.dart`: generated
* `events.dart`: Client calls of the Events API (Open Food Facts).
* `EventsBase.dart`: Events API: event.
* `EventsBase.g.dart`: generated
* `LeaderboardEntry.dart`: Events API: leaderboard entry.
* `LeaderboardEntry.g.dart`: generated

Impacted files:
* `JsonHelper.dart`: added a nullable timestamp conversion method.
* `OpenFoodAPIConfiguration.dart`: added host entries for Events API.
* `openfoodfacts.dart`: exported new Events related files.
* `UriHelper.dart`: added Events related method `getEventsUri`.

- Fixes #349

### DIFF
--- a/lib/events.dart
+++ b/lib/events.dart
@@ -1,0 +1,214 @@
+import 'dart:async';
+import 'dart:convert';
+
+import 'package:http/http.dart';
+import 'package:openfoodfacts/model/BadgeBase.dart';
+import 'package:openfoodfacts/model/EventCreate.dart';
+import 'package:openfoodfacts/model/EventsBase.dart';
+import 'package:openfoodfacts/model/LeaderboardEntry.dart';
+import 'package:openfoodfacts/model/User.dart';
+
+import 'utils/HttpHelper.dart';
+import 'utils/QueryType.dart';
+import 'utils/UriHelper.dart';
+
+/// Client calls of the Events API (Open Food Facts).
+///
+/// cf. https://events.openfoodfacts.net/docs
+class EventsAPIClient {
+  EventsAPIClient._();
+
+  /// Returns all the [EventsBase], with optional filters and paging.
+  static Future<List<EventsBase>> getEvents({
+    final String? userId,
+    final String? deviceId,
+    final int? skip,
+    final int? limit,
+    final QueryType? queryType,
+  }) async {
+    final Map<String, String> parameters = <String, String>{};
+    if (userId != null) {
+      parameters['user_id'] = userId;
+    }
+    if (deviceId != null) {
+      parameters['device_id'] = deviceId;
+    }
+    if (skip != null) {
+      parameters['skip'] = skip.toString();
+    }
+    if (limit != null) {
+      parameters['limit'] = limit.toString();
+    }
+    final Response response = await HttpHelper().doGetRequest(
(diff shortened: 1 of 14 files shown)


teolemon:
Oh great :-) gamification in Smoothie 🎉

teolemon (approved):

M123-dev (approved):
Looks good, thanks @monsieurtanuki good start that it is in it's own class

monsieurtanuki:
Thank you @teolemon @M123-dev for your review!

As mentioned earlier, there's one method that I could not test, the only one that writes: "create events", because I get the following error: `Incorrect username or password`.
@ocervell in my tests I saw that you were the only user with Events data in TEST and PROD environments: where do the Events users come from? Are they supposed to be "standard" open food facts users or are they specific to Events - if so how are they created?

monsieurtanuki:
@ocervell ping

ocervell:
@monsieurtanuki events are currently created by the admin user, set on the API side by the env variables `ADMIN_USERNAME` and `ADMIN_PASSWORD`. Initially we were not supposed to expose that API, so the OFF API should contact it, not the client app.

The request flow should be:
- User in app responds to questions / adds a product / etc...
- API call made to off-server backend
- off-server backend authentifies to Events API (not exposed externally) using admin credentials and adds the event

The problem with client app authenticating directly to the Events API is that anyone could add as many events as they want by simply using the API, thus falsifying the data (imagine if I just trace the API calls to events API I could add myself 20 "product_add" events and thus increase my score artifically.

teolemon:
Oh, I really goofed on that one. I'm so sorry.

ocervell:
Note that only the CreateEvent method is obsolete, the rest is still valid (querying badges for a user, querying leaderboard, querying events ...) so the PR is still mostly valid.

monsieurtanuki:
Thank you @ocervell for your answers.

If I understand correctly, that means that we need the off backend to directly call CreateEvent each time an "event" is detected, like "product_add". And there's nothing we have to do on the app side, except read/count the events, right?

ocervell:
Yes, that's correct, `robotoff `and the OFF backend will have code to contact the Events API CreateEvent endpoint internally. There might be an exception for the 'invite_shared' event as this is an in-app event but we could add an API call in off backend for that one (or just remove that event).

PR #398, opened 2022-03-01
test: added disabled test about nutrition

monsieurtanuki:
Just to keep in mind it should be fixed - one day.

Impacted file:
* `api_saveProduct_test.dart`

### What
cf. https://github.com/openfoodfacts/smooth-app/pull/1142


M123-dev (approved):
An additional todo would still be good

monsieurtanuki:
Thank you @M123-dev for your review; I've added the `TODO`.

Issue #401, opened 2022-03-02
Download of the integrality of a taxonomy


monsieurtanuki:
**Is your feature request related to a problem? Please describe.**
Typical use-case: in my app I want the end-user to change taxonomy values for a product, therefore I need to need which are the possible values. Currently the only possibility is to get the root values and to climb the value tree. Each time needing internet access and `await`ing.

**Describe the solution you'd like**
A single API call would download the whole taxonomy tree, to be used by the app developer without additional API calls.

**Describe alternatives you've considered**
The alternative solution is running recursive API calls.

stephanegigandet:
Note: Some taxonomies like categories are huge and may take quite a long time to download. Then there's the issue that many taxonomies get updated frequently. Depending on the use case and the actual taxonomy, there might be other options (e.g. suggest as you type as we have on the OFF web site).

monsieurtanuki:
@stephanegigandet That makes sense, but that also means we cannot change taxonomy values for a product when offline, right? Feel free to close this issue if it's not relevant.

PR #402, opened 2022-03-03
feat: #401 - added packaging taxonomy

monsieurtanuki:
New files:
* `api_getTaxonomyPackagingsServer_test.dart`: Integration test about packagings.
* `TaxonomyPackaging.dart`: A JSON-serializable version of a Packaging taxonomy result.
* `TaxonomyPackaging.g.dart`: generated

Impacted file:
* `openfoodfacts.dart`: new method `getTaxonomyPackagings`; exported new class `TaxonomyPackaging`

### What
- first step towards a potential "download the whole tree" method for selected taxonomies
- in the meanwhile, `TaxonomyPackaging` did not exist before and deserved to be created

### DIFF
--- a/lib/model/TaxonomyPackaging.dart
+++ b/lib/model/TaxonomyPackaging.dart
@@ -1,0 +1,147 @@
+import 'package:json_annotation/json_annotation.dart';
+import 'package:openfoodfacts/interface/JsonObject.dart';
+import 'package:openfoodfacts/openfoodfacts.dart';
+import 'package:openfoodfacts/utils/CountryHelper.dart';
+import 'package:openfoodfacts/utils/TaxonomyQueryConfiguration.dart';
+import 'package:openfoodfacts/utils/TagType.dart';
+
+part 'TaxonomyPackaging.g.dart';
+
+/// Fields of an [TaxonomyPackaging]
+enum TaxonomyPackagingField {
+  ALL,
+  NAME,
+  SYNONYMS,
+  WIKIDATA,
+  CHILDREN,
+  PARENTS,
+  //PACKAGING_MATERIALS, only 4/122 found for roots
+  //NON_RECYCLABLE_AND_NOT_BIODEGRADABLE, only 4/122 found for roots
+  //WEIGHT, only 4/122 found for roots
+  //PACKAGING_SHAPE, only 2/122 found for roots
+}
+
+extension TaxonomyPackagingFieldExtension on TaxonomyPackagingField {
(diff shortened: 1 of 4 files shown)


lib/model/TaxonomyPackaging.dart
```diff
@@ -0,0 +1,147 @@
...
+    TaxonomyPackagingField.ALL: 'all',
+    TaxonomyPackagingField.NAME: 'name',
+    TaxonomyPackagingField.SYNONYMS: 'synonyms',
+    TaxonomyPackagingField.WIKIDATA: 'wikidata',
+    TaxonomyPackagingField.CHILDREN: 'children',
+    TaxonomyPackagingField.PARENTS: 'parents',
+    //TaxonomyPackagingField.PACKAGING_MATERIALS: 'packaging_materials',
+    //TaxonomyPackagingField.NON_RECYCLABLE_AND_NOT_BIODEGRADABLE:'non_recyclable_and_non_biodegradable',
+    //TaxonomyPackagingField.WEIGHT: 'weight',
+    //TaxonomyPackagingField.PACKAGING_SHAPE:'packaging_shapes',
```
M123-dev, line 36: Any reason why they are commented out besides that they are very rare

lib/openfoodfacts.dart
```diff
@@ -499,6 +501,17 @@ class OpenFoodAPIClient {
...
   }
 
+  static Future<Map<String, TaxonomyPackaging>?> getTaxonomyPackagings(
+    final TaxonomyPackagingQueryConfiguration configuration, {
+    final User? user,
+    final QueryType? queryType,
+  }) =>
+      getTaxonomy<TaxonomyPackaging, TaxonomyPackagingField>(
+        configuration,
+        user: user,
```
M123-dev, line 511: Do we even need to put a user here

M123-dev (approved):

lib/model/TaxonomyPackaging.dart
monsieurtanuki: The fact that they are very rare means that they are not reliable.
As I started this PR from json results, I'm just guessing the fields we should use. Maybe there are even other rarer fields.
In first approach, let's not get distracted by useless fields.

lib/openfoodfacts.dart
monsieurtanuki: I don't know, I've just copied the code from other taxonomies.

monsieurtanuki:

Issue #408, opened 2022-03-13
Map the source_text, source_url, ... fields for the Knowledge panel API


teolemon:
### What
- Map the `source_text, source_url,...` field for the Knowledge panel API

### Why
- https://github.com/openfoodfacts/smooth-app/issues/1026
### Code pointer
- https://world.openfoodfacts.org/api/v0/product/5425015401027.json?fields=knowledge_panels
### Screenshots
![image](https://user-images.githubusercontent.com/1689815/151361069-b1981f3f-670a-4148-b226-f4c1de54674d.png)
![image](https://user-images.githubusercontent.com/1689815/151361126-fcf8792b-4c25-48d6-a542-817dd4a853b2.png)

PR #409, opened 2022-03-13
feat: Support for source fields in the Knowledge panel API

g123k:
PR related to issue #408

I have included the different fields within the `KnowledgePanelTextElement` class.

I was wondering if a dedicated class like `KnowledgePanelTextElementSource` wouldn't be better in that case.
However, it would force us to create a dedicated parser/serializer.
What would you think about this?

### DIFF
--- a/lib/model/KnowledgePanelElement.dart
+++ b/lib/model/KnowledgePanelElement.dart
@@ -1,5 +1,6 @@
 import 'package:json_annotation/json_annotation.dart';
 import 'package:openfoodfacts/model/KnowledgePanel.dart';
+
 import '../interface/JsonObject.dart';
 
 part 'KnowledgePanelElement.g.dart';
@@ -9,7 +10,6 @@ enum KnowledgePanelTextElementType {
   /// The description summarizes the knowledge panel.
   @JsonValue('summary')
   SUMMARY,
-
   @JsonValue('warning')
   WARNING,
 
@@ -33,6 +33,10 @@ enum KnowledgePanelColumnType {
   PERCENT,
 }
 
+Object? test(Map map, String str) {
+  return null;
+}
+
 /// Description element of the Knowledge panel.
 @JsonSerializable()
 class KnowledgePanelTextElement extends JsonObject {
@@ -46,7 +50,29 @@ class KnowledgePanelTextElement extends JsonObject {
       unknownEnumValue: KnowledgePanelTextElementType.DEFAULT)
   final KnowledgePanelTextElementType? type;
 
-  const KnowledgePanelTextElement({required this.html, this.type});
+  /// Human readable source language (eg: "English")
+  @JsonKey(name: 'source_language')
+  final String? sourceLanguage;
+
+  /// Source locale name (eg: "en")
+  @JsonKey(name: 'source_lc')
+  final String? sourceLocale;
+
+  /// Name of the source (eg: "Wikipedia")
+  @JsonKey(name: 'source_text')
+  final String? sourceText;
+
+  /// Link to the source (eg: "https://en.wikipedia.org/wiki/Sodium acetate")
+  @JsonKey(name: 'source_url')
+  final String? sourceUrl;
+
+  const KnowledgePanelTextElement(
+      {required this.html,
+      this.type,
+      this.sourceLanguage,
+      this.sourceLocale,
+      this.sourceText,
+      this.sourceUrl});
 
   factory KnowledgePanelTextElement.fromJson(Map<String, dynamic> json) =>
       _$KnowledgePanelTextElementFromJson(json);
@@ -279,19 +305,14 @@ enum KnowledgePanelElementType {
   /// The description summarizes the knowledge panel.
   @JsonValue('text')
   TEXT,
-
   @JsonValue('image')
   IMAGE,
-
   @JsonValue('panel')
   PANEL,
-
   @JsonValue('panel_group')
   PANEL_GROUP,
-
   @JsonValue('table')
   TABLE,
-
   @JsonValue('map')
   MAP,
   UNKNOWN,
--- a/lib/model/KnowledgePanelElement.dart
+++ b/lib/model/KnowledgePanelElement.dart
@@ -1,6 +1,5 @@
 import 'package:json_annotation/json_annotation.dart';
 import 'package:openfoodfacts/model/KnowledgePanel.dart';
-
 import '../interface/JsonObject.dart';
 
 part 'KnowledgePanelElement.g.dart';
@@ -10,6 +9,7 @@ enum KnowledgePanelTextElementType {
   /// The description summarizes the knowledge panel.
   @JsonValue('summary')
   SUMMARY,
+
   @JsonValue('warning')
   WARNING,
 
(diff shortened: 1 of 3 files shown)


lib/model/KnowledgePanelElement.dart
```diff
@@ -279,19 +305,14 @@ enum KnowledgePanelElementType {
   /// The description summarizes the knowledge panel.
   @JsonValue('text')
   TEXT,
```
monsieurtanuki, line 307: Please avoid to reformat your way by adding or removing empty lines.

lib/model/KnowledgePanelElement.dart
```diff
@@ -46,7 +50,29 @@ class KnowledgePanelTextElement extends JsonObject {
...
+  @JsonKey(name: 'source_url')
+  final String? sourceUrl;
+
+  const KnowledgePanelTextElement(
+      {required this.html,
+      this.type,
+      this.sourceLanguage,
+      this.sourceLocale,
+      this.sourceText,
+      this.sourceUrl});
```
monsieurtanuki, line 75: A trailing comma + reformat would be nice.

lib/model/KnowledgePanelElement.dart
```diff
@@ -33,6 +33,10 @@ enum KnowledgePanelColumnType {
   PERCENT,
 }
 
+Object? test(Map map, String str) {
+  return null;
+}
+
```
monsieurtanuki, line 39: Please remove that.

monsieurtanuki (commented):
Hi @g123k!
Please read my minor comments.
In addition to that, I'd like to see those new fields in unit tests.

lib/model/KnowledgePanelElement.dart
g123k: Ooops sorry for this one

lib/model/KnowledgePanelElement.dart
g123k: This is not the case for all constructors in this file (eg: KnowledgePanelTableCell, KnowledgePanelPanelIdElement…).
Should I add only to this particular class or to the whole file?

monsieurtanuki: Just this one. The less I need to read, the better.

lib/model/KnowledgePanelElement.dart
g123k: Actually that's what the dart formatter (dart format) generates.
This repo doesn't follow the dart rules?

monsieurtanuki: Oh keep it that way then. The rules change sometimes.

g123k:
Hi @monsieurtanuki and thanks for your super quick review.
For the unit test part, I have simply added some data to an existing test.
Do you prefer a dedicated one instead?

monsieurtanuki:
> I have simply added some data to an existing test.

Not really: you've added data to some expected values. You didn't actually test the values.
I guess you could find some place in `api_getProduct_test.dart` to add some test specifically for the fields you added.

test/knowledge_panels_from_json_test.dart
```diff
@@ -422,4 +427,67 @@ void main() {
     KnowledgePanels kp = KnowledgePanels.fromJson(panels);
     expect(kp.panelIdToPanelMap.length, equals(8));
   });
+
+  // Ensure that a "text_element" with some sources data is well parsed
+  test('Load knowledge panel JSON with a text element', () {
+    final String id = 'additive_en:e262';
+
+    final Map<String, dynamic> jsonData = <String, dynamic>{
```
monsieurtanuki, line 435: Given that `'additive_en:e262'` is not in your json string I would be surprised if the test actually passed.

test/knowledge_panels_from_json_test.dart
```diff
@@ -422,4 +427,67 @@ void main() {
...
+
+    KnowledgePanels kp = KnowledgePanels.fromJson(jsonData);
+    expect(kp.panelIdToPanelMap.length, equals(1));
+
+    expect(
+      kp.panelIdToPanelMap[id]!.elements!.length,
+      equals(1),
+    );
+    expect(
+      kp.panelIdToPanelMap[id]!.elements!.first.elementType,
```
monsieurtanuki, line 465: I'm sure you could put `kp.panelIdToPanelMap[id]!.elements!.first` in a variable instead of repeating it 6 times.

monsieurtanuki (commented):
@g123k As you can see [in the logs the test failed](https://github.com/openfoodfacts/openfoodfacts-dart/runs/5529351775?check_suite_focus=true), for dart formatting reasons: please fix that.
In addition to that I'd be very surprised if it passed once the format is fixed - please have a look at my comments.
Beyond those consideration there's my natural preference for integration tests rather than unit tests, but that's another story. 
Just fine-tune the test and we're good!

monsieurtanuki (approved):
Now it's better @g123k!

teolemon:
@monsieurtanuki the failing test seem to be the taxo ones fixed your separate PR ?
this thus seems good to merge @g123k  ?
```
00:07 +10 ~1: test/api_getTaxonomyCountriesServer_test.dart: OpenFoodAPIClient getTaxonomyCountries get all countries                                                                                  
00:07 +10 ~1: test/api_getTaxonomyCountriesServer_test.dart: OpenFoodAPIClient getTaxonomyCountries get a country                                                                                      
00:08 +10 ~1: test/api_getTaxonomyCountriesServer_test.dart: OpenFoodAPIClient getTaxonomyCountries get a country                                                                                      
00:08 +10 ~1 -1: test/api_getTaxonomyCountriesServer_test.dart: OpenFoodAPIClient getTaxonomyCountries get a country [E]
                                                      
  Null check operator used on a null value
  test/api_getTaxonomyCountriesServer_test.dart [45](https://github.com/openfoodfacts/openfoodfacts-dart/runs/5529800584?check_suite_focus=true#step:7:45):31  main.<fn>.<fn>


00:08 +10 ~1 -1: test/api_getTaxonomyCountriesServer_test.dart: OpenFoodAPIClient getTaxonomyCountries get a country that doesn't exist                                                                
00:09 +10 ~1 -1: test/api_getTaxonomyCountriesServer_test.dart: OpenFoodAPIClient getTaxonomyCountries get a country that doesn't exist                                                                
00:09 +11 ~1 -1: test/api_getTaxonomyCountriesServer_test.dart: OpenFoodAPIClient getTaxonomyCountries get a country that doesn't exist                                                                
00:09 +11 ~1 -1: test/api_getTaxonomyCountriesServer_test.dart: OpenFoodAPIClient getTaxonomyCountries get a country that doesn't exist with one that does                                             
00:09 +11 ~1 -2: test/api_getTaxonomyCountriesServer_test.dart: OpenFoodAPIClient getTaxonomyCountries get a country that doesn't exist with one that does [E]                                         
  Null check operator used on a null value
  test/api_getTaxonomyCountriesServer_test.dart 68:31  main.<fn>.<fn>
```

monsieurtanuki:
@teolemon That's correct: this PR should be merged, and #411 should be approved :)

teolemon:
@g123k congratulations on your first PR to openfoodfacts-dart 🎉

Issue #407, opened 2022-03-10
Limit for number of products in "getProductList"?


monsieurtanuki:
**Describe the potential bug**
`OpenFoodAPIClient.getProductList` takes a list of barcodes as an input parameter, and concatenates them in a GET url that retrieves the product data.
But there may be limits in the url length.

**Preliminary checks**
From what I found on the internet the url max length is around 2000 characters.
Let's say that we need to provide enough barcodes to go to 3000 characters, and see what happens (does it crash? Do we get all the products that match the barcodes, or only the one that match the first 300 barcodes?)

**Actions**
A solution might be to change the request as POST, or to ask the same thing to the server side but with another syntax.

monsieurtanuki:
That was a good idea:
```html
<html>
<head><title>414 Request-URI Too Large</title></head>
<body bgcolor="white">
<center><h1>414 Request-URI Too Large</h1></center>
<hr><center>nginx/1.10.3</center>
</body>
</html>
```

In my tests it looks like 8000 characters is the limit.

cf. #105: @teolemon @stephanegigandet Is there another way (like POST) to get a list of products than https://world.openfoodfacts.dev/products/8024884500403,3263855093192,3045320001570,3021762383344,4008400402222,3330720237255,3608580823513,3700278403936,3302747010029,3608580823490,3250391660995,3760020506605,8722700202387,3330720237330,3535800940005,20000691,3270190127512.json?fields=code,product_name?

AshAman999:
I am new to this repo and eager to contribute, 
As it seems backend also needs to be ready to accept the changes if we are making a post request to get the data .

```
// thought of something like this instead of a get request
List<String> foodList = ['13434','343131','343131'];
  http.Response response = await http.post(
    Uri.parse('http://some_random_url'),
    headers: <String, String>{
      'Content-Type': 'application/json; charset=UTF-8',
    },
    body: jsonEncode(
      {
        'foodList': foodList,
      },
    ),
  );
```

if possible can i work on the front end part of this issue, a little bit of help in the context where to make the changes would help me a lot .
Thanks :)

monsieurtanuki:
Hi @AshAman999! First we definitely need some changes on the server side, or another existing URL.
ping @stephanegigandet  @teolemon

stephanegigandet:
@monsieurtanuki the OFF API accepts all parameters as POST parameters, but don't mix and match them: either send only query parameters, or only POST parameters.

[...]

PR #413, opened 2022-03-15
fix: #407 - getProductList now works with unlimited number of barcodes

monsieurtanuki:
Impacted files:
* `api_searchProducts_test.dart`: additional test; refactored
* `openfoodfacts.dart`: changed method `getProductList` from GET to POST
* `ProductListQueryConfiguration.dart`: overrode `getParametersMap` in order to add barcodes as POST parameters

### What
- GET has limits (number of characters), POST doesn't.
- For getProductList, there are potentially tons of parameters (the barcodes), therefore we had to play it safe and to switch from GET to POST

### Fixes bug(s)
- #407

### DIFF
--- a/lib/openfoodfacts.dart
+++ b/lib/openfoodfacts.dart
@@ -324,14 +324,14 @@ class OpenFoodAPIClient {
     QueryType? queryType,
   }) async {
     final Uri uri = UriHelper.getUri(
-      path: 'products/${configuration.barcodes.join(',')}.json',
-      queryParameters: configuration.getParametersMap(),
+      path: 'api/v2/search/',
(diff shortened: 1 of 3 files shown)


M123-dev (approved):

monsieurtanuki:
The first use-case is for a product history refresh. Regarding ETag that would make sense too, though I think we already saw that ETag was not precise enough as core data may not change while computed data may (e.g. ecoscore), with an identical ETag. But maybe we were talking about the timestamp instead.
Anyway, with this PR we can now get tons of products in the same query. Maybe we should test the page parameter too, but that's another story.

Issue #232, opened 2021-09-16
Update tests so that they do not expect specific values from existing products on openfoodfacts.net, unless they write them first


stephanegigandet:
We changed the infrastructure that serves the openfoodfacts.net preprod / test server yesterday afternoon, and 13 tests are failing (67 pass).

Initially many more tests were failing because of a slightly different nginx configuration that resulted in URLs like https://world.openfoodfacts.net/api/v0/product/3274080005003.json with ".json" at the end not being served (we could remove the .json from the call, it's not needed).

The failing tests are because a lot of values for a lot of products have changed: the openfoodfacts.net server now has a full copy of the production database (including images), and that copy is updated every night.

That means that the tests can't rely on a specific value for an existing product: the tests need to first write those values, and then after that read them.

gspencergoog:
Sounds like what we need is a test database with values that don't change, or some standard setup that writes a bunch of test data that can be used in tests.

Issue #318, opened 2021-12-13
api_addProductImage_test.dart failing


teolemon:
02:58 +63 -4: test/api_addProductImage_test.dart: OpenFoodAPIClient modify product image image crop [E]                                                                                                
  Expected: <50>
    Actual: <400>
  
  package:test_api                          expect
  test/api_addProductImage_test.dart 195:9  main.<fn>.<fn>
  

02:58 +63 -4: loading test/api_getTaxonomyLabels_test.dart

monsieurtanuki:
@teolemon I've just tested again locally, it works.
In that test I crop an image into a [50,300] rectangle. It looks like my crop instruction has priority over the "400 max" rule.
Probably it's just a "bad luck" collision. There would be less collisions if distinct barcodes were used for "image rotation" and "image crop" tests.

monsieurtanuki:
It looks ok now.

PR #419, opened 2022-03-21
fix: actually upload an image on a random product

stephanegigandet:
Related to #232 

The current image upload test does not update an image, because the product already has it. This PR uses a random barcode so that we can actually upload the image.

### DIFF
--- a/lib/utils/OpenFoodAPIConfiguration.dart
+++ b/lib/utils/OpenFoodAPIConfiguration.dart
@@ -29,6 +29,14 @@ class OpenFoodAPIConfiguration {
   ///Uri host of the test requests to the backend
   static String uriTestHost = 'world.openfoodfacts.net';
 
+  ///Url base for images in prod: needs to match the domain of uriProdHost
+  static String ImageProdUrlBase =
+      'https://static.openfoodfacts.org/images/products/';
+
+  ///Url base for images in test: needs to match the domain of uriTestHost
+  static String ImageTestUrlBase =
+      'https://off:off@static.openfoodfacts.net/images/products/';
+
   ///Uri host of the Robotoff requests to the backend, modify this to direct the request to a self-hosted instance.
   static String uriProdHostRobotoff = 'robotoff.openfoodfacts.org';
 
--- a/lib/utils/OpenFoodAPIConfiguration.dart
+++ b/lib/utils/OpenFoodAPIConfiguration.dart
@@ -27,7 +27,7 @@ class OpenFoodAPIConfiguration {
   static String uriProdHost = 'world.openfoodfacts.org';
 
   ///Uri host of the test requests to the backend
-  static String uriTestHost = 'world.openfoodfacts.dev';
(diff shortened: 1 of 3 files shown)


lib/utils/OpenFoodAPIConfiguration.dart
```diff
@@ -29,6 +29,14 @@ class OpenFoodAPIConfiguration {
   ///Uri host of the test requests to the backend
   static String uriTestHost = 'world.openfoodfacts.net';
 
+  ///Url base for images in prod: needs to match the domain of uriProdHost
+  static String ImageProdUrlBase =
```
g123k, line 33: It would be better here to add the `const` keyword.
The same goes for line 37

test/api_addProductImage_test.dart
```diff
@@ -10,10 +12,14 @@ void main() {
   const User user = TestConstants.TEST_USER;
 
   /// Common constants for several image operations
-  const String barcode = '4250752200784';
   const OpenFoodFactsLanguage language = OpenFoodFactsLanguage.GERMAN;
   const ImageField imageField = ImageField.FRONT;
 
+  /// Use a random barcode so that we can create a new product
+  /// and really upload images
+  var r = Random();
```
g123k, line 20: Could you specify the variable type, instead of the `var` keyword please?

g123k (changes requested):
A few minor things to fix

lib/utils/OpenFoodAPIConfiguration.dart
stephanegigandet: The other strings in this file don't have const, I'm guessing it's because we want to be able to update them at runtime. (e.g. in smoothie we can set any API server url through the advanced settings mode)

test/api_addProductImage_test.dart
stephanegigandet: is String barcode = (50000000000000 + Random().nextInt(100000000)).toString(); ok ?

stephanegigandet:
Should also solve #318

lib/utils/OpenFoodAPIConfiguration.dart
g123k: Mmmm good point, that's not how I manage this kind of config in my other projects, but why not.
Forget this comment in that case

test/api_addProductImage_test.dart
```diff
@@ -1,3 +1,5 @@
+import 'dart:math';
```
g123k, line 1: In general, we import the "math" library with a prefix: `import 'dart:math' as math;`
and all related calls will now look like this: `math.xxx`

test/api_addProductImage_test.dart
g123k: My point here was just to change the var keyword with the type of Variable

`var r = Random();`
⬇️
`Random r = Random();`

g123k (changes requested):
A few comments added ;)

Issue #422, opened 2022-03-22
Add support for 'no_nutrition_data'


teolemon:
### What
- Add support for 'no_nutrition_data'
- Proto documentation available at: https://github.com/openfoodfacts/api-documentation/issues/70

### Part of
- #218

PR #423, opened 2022-03-22
feat: Support for Product - No nutrition data field

g123k:
Support for the "no_nutrition_data" field in the Product object.

I have modified the toJson() method to ensure that when "no_nutrition_data" is true, no nutriment should be passed (an empty object is permitted

Please let me know if additional tests are required


lib/model/Nutriments.dart
```diff
@@ -1,5 +1,6 @@
 import 'package:json_annotation/json_annotation.dart';
 import 'package:openfoodfacts/utils/UnitHelper.dart';
+
```
g123k, line 3: That lineis created by dartfmt ;)

teolemon:
I'll defer to @stephanegigandet on this, but If I am right, the no_nutrition_facts boolean (True or False) has to be sent server side, and just not sending the nutrients is not the right way to do it.
Otherwise, we cannot reenable nutrition facts, just like on Android.

g123k:
Ok thanks for the comment @teolemon.
Right now, a nutriment object can still be passed, but with all values set to null

lib/utils/JsonHelper.dart
```diff
@@ -201,4 +201,17 @@ class JsonHelper {
...
+  static bool checkboxFromJSON(dynamic jsonValue) {
+    return jsonValue is String && jsonValue.trim().toLowerCase() == 'on';
+  }
+
+  static String checkboxToJSON(dynamic value) {
+    if (value == true ||
+        (value is String && value.trim().toLowerCase() == 'on')) {
+      return 'on';
+    } else {
+      return 'off';
```
stephanegigandet, line 214: The possible values for the checkbox are in fact either "on" when it is checked, or "" (empty string) or not existing field when it is unchecked.

test/api_getProduct_test.dart
```diff
@@ -1858,4 +1858,57 @@ void main() {
...
+      final ProductResult result = await OpenFoodAPIClient.getProduct(
+        configurations,
+        user: TestConstants.TEST_USER,
+      );
+
+      expect(result.product?.noNutritionData, isFalse);
+    });
+
+    test('Empty value', () async {
+      const String barcode = '5449000000996';
```
stephanegigandet, line 1881: Those products are likely not to exist when the .net database gets refreshed.

I recommend that any test that expects a specific value first writes that value.

lib/utils/JsonHelper.dart
g123k: Ok, should now be fixed with the new commit

test/api_getProduct_test.dart
g123k: Ok @stephanegigandet 

Just to confirm if I have well understood: 
- First, I create a product locally with the expected value
- Upload this product
- Fetch this product for the test?

stephanegigandet: Yes, I think we have to do that so that tests can pass regardless of the initial state of the OFF test database.

g123k:
This should now be OK :)

lib/model/Nutriments.dart
```diff
@@ -1513,4 +1514,179 @@ class Nutriments extends JsonObject {
     'omega-6-fat',
     'trans-fat',
   };
+
+  bool get hasData =>
```
monsieurtanuki, line 1518: I'm not convinced by that bug-proned method. I'll detail my thoughts a bit later.

lib/model/Product.dart
```diff
@@ -592,6 +600,12 @@ class Product extends JsonObject {
       }
     }
 
+    if (noNutritionData == true && nutriments?.hasData == true) {
+      throw StateError(
+        'If \'noNutritionData\' is true, no \'nutriments\' should be provided!',
+      );
+    }
+
```
monsieurtanuki, line 608: I think we would be better off
* making `nutriments` private (therefore `_nutriments`)
* giving "public" access to `nutriments` with getter/setter
* in those getter/setter, include the `noNutritionData` flag, e.g. something like that
```dart
Nutriments? get nutriments => noNutritionData == true ? null : _nutriments
```

test/api_getProduct_test.dart
```diff
@@ -1858,4 +1859,77 @@ void main() {
     expect(result.product?.packagingTags, ['de:in-einer-plastikflasche']);
     expect(result.product?.quantity, '5.5 Liter');
   });
+
+  group('no nutrition data', () {
+    const String barcode = '111111555555';
```
monsieurtanuki, line 1864: Please add a comment stating that it's an unused fake barcode.

test/api_saveProduct_test.dart
```diff
@@ -487,4 +488,19 @@ void main() {
         // some tests can be slow here
         Duration(seconds: 90),
       ));
+
+  test('No nutrition data with nutriments', () async {
```
monsieurtanuki, line 492: As said earlier, I don't believe in maintaining `hasData`, therefore I don't think we should send an exception in that case - we should just ignore the nutriment parameter over which noNutritionData has priority.

What could be tested instead is:
* create a product with nutriments
* get the product: nutriments, and noNutritionData not true
* save same product with noNutritionData = true
* get the product: null nutriments, and noNutritionData true

monsieurtanuki (commented):
Hi @g123k!
Please have a look at my comments from a place where we used getters and setters, and tell me what you think.

lib/model/Product.dart
g123k: Your solution is indeed the one I should implement.
However if `nutriments` and/or `nutritionData` is private, there is no way to have a correct mapping with json_serializable… or please correct me on this.

Do you have any idea how to bypass this problem?

g123k:
@monsieurtanuki I've made a few changes to have private values for `no_nutrition_data` and `nutriments`.
I think I should be better now.

lib/model/Product.dart
```diff
@@ -655,4 +666,28 @@ class Product extends JsonObject {
...
+  set noNutritionData(bool? noNutritionData) {
+    if (_noNutritionData == true) {
+      _nutriments = null;
+    }
+    _noNutritionData = noNutritionData;
+  }
+
+  @JsonKey(
+      name: 'nutriments', includeIfNull: false, toJson: Nutriments.toJsonHelper)
```
monsieurtanuki, line 684: Please put a `,` before the `)`.

lib/model/Product.dart
```diff
@@ -655,4 +666,28 @@ class Product extends JsonObject {
...
+  Nutriments? get nutriments => _noNutritionData == true ? null : _nutriments;
+
+  set nutriments(Nutriments? nutriments) {
+    if (nutriments != null) {
+      _noNutritionData = false;
+    }
```
monsieurtanuki, line 690: What about `_noNutritionData = nutriments == null;`?

lib/model/Product.dart
```diff
@@ -231,9 +231,17 @@ class Product extends JsonObject {
       toJson: IngredientsAnalysisTags.toJson)
   IngredientsAnalysisTags? ingredientsAnalysisTags;
 
-  @JsonKey(
-      name: 'nutriments', includeIfNull: false, toJson: Nutriments.toJsonHelper)
-  Nutriments? nutriments;
+  /// When no nutrition data is true, nutriments are always null
+  /// This logic is handled by the getters/setters of [noNutritionData] and
+  /// [nutriments]
+  @JsonKey(ignore: true)
```
monsieurtanuki, line 237: We don't need `@JsonKey` there, do we?

lib/model/Product.dart
```diff
@@ -231,9 +231,17 @@ class Product extends JsonObject {
...
+  bool? _noNutritionData;
+
+  /// When nutriments are not null, no nutrition data can't be true
+  /// This logic is handled by the getters/setters of [noNutritionData] and
+  /// [nutriments]
+  @JsonKey(ignore: true)
```
monsieurtanuki, line 243: We don't need `@JsonKey` there, do we?

monsieurtanuki (approved):
Hi @g123k!
Thank you for your refactoring, the code is easier to read now. Minor fixes suggested here and there, please have a look at them.

I still have a strange feeling that we miss something here, but as I cannot put words on that feeling let's say that the code is right :)

lib/model/Product.dart
g123k: Ok, on other fields like `productNameInLanguages` the comma is also missing
Should I also add it to them?

lib/model/Product.dart
g123k: Json Serialize doesn't work with private fields

lib/model/Product.dart
g123k: Json Serialize doesn't work with private fields, it will just ignore them

g123k:
@monsieurtanuki It should be OK now

monsieurtanuki:
@g123k Beyond unrelated tests, your own new tests didn't pass. Please have a look at the test logs.

monsieurtanuki:
Capture d’écran 2022-04-01 à 10 49 08

monsieurtanuki:
@g123k I guess now we're all set, and you can merge this PR and close #422, right?

monsieurtanuki:
@g123k Oh I didn't notice that you changed the code again :(

lib/utils/JsonHelper.dart
```diff
@@ -201,4 +201,23 @@ class JsonHelper {
   /// Returns a [DateTime] from a JSON-encoded String (e.g. '2021-10-29T11:00:56.177379')
   static DateTime? nullableStringTimestampToDate(dynamic json) =>
       json == null ? null : stringTimestampToDate(json);
+
+  static const String checkboxOnValue = 'on';
+  static const String checkboxOffValue = '';
```
monsieurtanuki, line 206: Should be private.

test/api_getProduct_test.dart
```diff
@@ -1858,4 +1859,60 @@ void main() {
     expect(result.product?.packagingTags, ['de:in-einer-plastikflasche']);
     expect(result.product?.quantity, '5.5 Liter');
   });
+
+  group('no nutrition data', () {
+    // This is barcode refers to a test product
+    const String barcode = '111111555555';
+
+    uploadProduct({required bool? noNutritionData}) =>
```
monsieurtanuki, line 1867: * Explicit `Future<void>` would make it easier to read.
* `bool` instead of `bool?` would make it easier to read.

test/api_getProduct_test.dart
```diff
@@ -1858,4 +1859,60 @@ void main() {
...
+
+    test('Without nutriments', () async {
+      await uploadProduct(noNutritionData: true);
+
+      final ProductQueryConfiguration configurations =
+          ProductQueryConfiguration(
+        barcode,
+        fields: [
+          ProductField.NO_NUTRITION_DATA,
+        ],
```
monsieurtanuki, line 1885: You should add `ProductField.NUTRIMENTS,`

test/api_saveProduct_test.dart
```diff
@@ -487,4 +488,39 @@ void main() {
...
+
+      expect(product.noNutritionData, isFalse);
+      expect(product.nutriments, isNotNull);
+
+      var serverData = product.toServerData();
+      expect(
+        serverData[ProductField.NO_NUTRITION_DATA.key],
+        equals(JsonHelper.checkboxOffValue),
+      );
+      expect(serverData[ProductField.NUTRIMENTS.key], isNotEmpty);
```
monsieurtanuki, line 523: It's a bit too much: all we need are the `noNutritionData` and `nutriments` fields.

test/api_saveProduct_test.dart
```diff
@@ -487,4 +488,39 @@ void main() {
...
+            salt: 1.0,
+          ));
+
+      expect(product.noNutritionData, isTrue);
+      expect(product.nutriments, isNull);
+
+      var serverData = product.toServerData();
+      expect(serverData[ProductField.NO_NUTRITION_DATA.key],
+          equals(JsonHelper.checkboxOnValue));
+      expect(serverData[ProductField.NUTRIMENTS.key], equals('{}'));
```
monsieurtanuki, line 506: It's a bit too much: all we need are the `noNutritionData` and `nutriments` fields.

test/api_getProduct_test.dart
```diff
@@ -1858,4 +1859,60 @@ void main() {
...
+      );
+
+      final ProductResult result = await OpenFoodAPIClient.getProduct(
+        configurations,
+        user: TestConstants.TEST_USER,
+      );
+
+      expect(result.product?.noNutritionData, isTrue);
+      expect(result.product?.nutriments, isNull);
```
monsieurtanuki, line 1894: `product!` instead of `product?`.

test/api_getProduct_test.dart
```diff
@@ -1858,4 +1859,60 @@ void main() {
...
+        ],
+      );
+
+      final ProductResult result = await OpenFoodAPIClient.getProduct(
+        configurations,
+        user: TestConstants.TEST_USER,
+      );
+
+      expect(result.product?.noNutritionData, isFalse);
+      expect(result.product?.nutriments, isNotNull);
```
monsieurtanuki, line 1915: `product!` instead of `product?`.

monsieurtanuki (commented):
@g123k Please have a look at my latest comments.
The added value of adding commas in `Product` is a questionable as it's destroyed the history delta check possibility.

test/api_saveProduct_test.dart
g123k: So you mean I should remove this part of the test?

monsieurtanuki (approved):
Thank you @g123k!

g123k:
@monsieurtanuki I have just this question: https://github.com/openfoodfacts/openfoodfacts-dart/pull/423#discussion_r846654121

Then after, I think everything will be resolved

test/api_saveProduct_test.dart
monsieurtanuki: I do.

g123k: Ok, I remove them 👍

M123-dev:
Ready to merge ?

monsieurtanuki:
Refactoring all the `Product` file and making it impossible to check for differences was not an initiative I agree with.

g123k:
If that's a showstopper, I can revert the "formatting" part

monsieurtanuki:
@g123k Please do.

g123k:
It should be OK now 🤓

monsieurtanuki (approved):

PR #424, opened 2022-03-30
fix: #349 - removed irrelevant "create event" classes and methods

monsieurtanuki:
Deleted files:
* `EventCreate.dart`
* `EventCreate.g.dart`

Impacted files:
* `api_events_test.dart`: removed commented code around `createEvent`
* `events.dart`: removed method `createEvent`
* `openfoodfacts.dart`: removed export of `EventCreate`

### What
- The "create event" action should be performed on the API side, not on the client side, cf. https://github.com/openfoodfacts/openfoodfacts-dart/pull/386#issuecomment-1073763346
- This PR is the removal of related irrelevant classes and methods


M123-dev (approved):
👍🏼

PR #428, opened 2022-04-01
fix: getProductRaw tests

monsieurtanuki:
Impacted file:
* `api_getProductRaw_test.dart`: getting cleaner data from PROD; more lenient tests

### What
- PR tests were failing because of data change on one product
- This PR slightly changed the tests, in order to pass


M123-dev (approved):
Thanks @monsieurtanuki

Issue #132, opened 2021-03-14
Add support for non-detructive value addition


teolemon:
add_categories
add_labels
add_brands
…

This adds a new value but doesn't delete the previous ones. 
 Examples:

https://world.openfoodfacts.net/cgi/product_jqm2.pl?code=0048151621226&user_id=username&password=*****&add_brands=Brand%203

MohamedFBoussaid:
@teolemon I guess once this implemented I should rollback my changes here ? https://github.com/openfoodfacts/openfoodfacts-dart/pull/128

Otherwise writing labels, categories will saveProduct() will delete the previous values and keep only the last input

teolemon:
@MohamedFBoussaid What's the current behavior ? We could have both ways, a full replace or a patch solution ?
add_ has some limitations as well (it does not allow for precision editing)

monsieurtanuki:
I suggest a new method like this one:
```dart
static Future<Status> OpenFoodAPIClient.addToProduct(
    User user,
    Product product,
    Map<AddableField, String> addedValues, {
    QueryType queryType = QueryType.PROD,
  })
```
with `AddableField` being an `enum` (`BRANDS`, `LABELS`, `CATEGORIES`, ...)
That would have no impact on @MohamedFBoussaid's code.
@teolemon Would that suit your needs?

VaiTon:
> @teolemon Would that suit your needs?

I think it would. I think editing the product is the main behavior we want to keep as a default here. @teolemon any thoughts?

M123-dev:
Yeah lets do it like that

monsieurtanuki:
@VaiTon Any idea about the list of addable fields? Will they all be accepting `String` as value?

VaiTon:
> @VaiTon Any idea about the list of addable fields? Will they all be accepting `String` as value?

I think every field has an 'add' version except from code and lang. (From here https://wiki.openfoodfacts.org/API/Write)

Yes, everything can be sent as a string.

On android we have atm:
```kotlin
const val ADD_BRANDS = "add_brands"
const val ADD_PURCHASE = "add_purchase_places"
const val ADD_STORES = "add_stores"
const val ADD_COUNTRIES = "add_countries"
const val ADD_TRACES = "add_traces"
```

teolemon:
[...]

PR #436, opened 2022-04-09
feat: #132 - new method OpenFoodAPIClient.addProductFields

monsieurtanuki:
New files:
* `AddableProductFields.dart`: Fields of a `Product` that can be simply added.
* `api_addProductFields_test.dart`: Unit/Integration tests around `OpenFoodAPIClient.addProductFields`

Impacted file:
* `openfoodfacts.dart`: new method `addProductFields`

### What
- new method `OpenFoodAPIClient.addProductFields`

### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/1515

### DIFF
--- a/lib/openfoodfacts.dart
+++ b/lib/openfoodfacts.dart
@@ -141,7 +141,7 @@ class OpenFoodAPIClient {
     QueryType? queryType,
   }) async {
     if (map.isEmpty) {
-      throw Exception('Empty map');
+      throw Exception('Please add at least one AddableProductField');
     }
 
     final Map<String, String> parameterMap = <String, String>{};
--- a/lib/openfoodfacts.dart
+++ b/lib/openfoodfacts.dart
@@ -20,6 +20,7 @@ import 'package:openfoodfacts/model/TaxonomyLabel.dart';
 import 'package:openfoodfacts/model/TaxonomyLanguage.dart';
 import 'package:openfoodfacts/model/TaxonomyPackaging.dart';
 import 'package:openfoodfacts/utils/AbstractQueryConfiguration.dart';
+import 'package:openfoodfacts/utils/AddableProductFields.dart';
 import 'package:openfoodfacts/utils/CountryHelper.dart';
 import 'package:openfoodfacts/utils/ImageHelper.dart';
 import 'package:openfoodfacts/utils/OcrField.dart';
@@ -124,6 +125,37 @@ class OpenFoodAPIClient {
     parameterMap.remove('nutriments');
     final Response response = await HttpHelper().doPostRequest(
       productUri,
+      parameterMap,
+      user,
+      queryType: queryType,
+    );
+    return Status.fromApiResponse(response.body);
+  }
+
+  /// Add the given product to the database.
+  /// Returns a Status object as result.
+  static Future<Status> addProductFields(
+    final User user,
+    final String barcode,
+    final Map<AddableProductField, String> map, {
+    QueryType? queryType,
+  }) async {
+    if (map.isEmpty) {
+      throw Exception('Empty map');
+    }
+
+    final Map<String, String> parameterMap = <String, String>{};
+    parameterMap.addAll(user.toData());
+    parameterMap['code'] = barcode;
+    for (final MapEntry<AddableProductField, String> entry in map.entries) {
+      parameterMap[entry.key.addableKey] = entry.value;
+    }
+
+    final Response response = await HttpHelper().doPostRequest(
+      UriHelper.getPostUri(
+        path: '/cgi/product_jqm2.pl',
+        queryType: queryType,
+      ),
       parameterMap,
       user,
       queryType: queryType,
--- a/lib/utils/AddableProductFields.dart
+++ b/lib/utils/AddableProductFields.dart
@@ -1,0 +1,17 @@
+/// Fields of a [Product] that can be simply added.
+enum AddableProductField {
+  BRANDS,
+  STORES,
+  COUNTRIES,
+}
+
+extension AddableProductFieldExtension on AddableProductField {
+  static const Map<AddableProductField, String> _KEYS = {
+    AddableProductField.BRANDS: 'add_brands',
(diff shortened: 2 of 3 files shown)


lib/openfoodfacts.dart
```diff
@@ -131,6 +132,37 @@ class OpenFoodAPIClient {
...
+  /// Add the given product to the database.
+  /// Returns a Status object as result.
+  static Future<Status> addProductFields(
+    final User user,
+    final String barcode,
+    final Map<AddableProductField, String> map, {
+    QueryType? queryType,
+  }) async {
+    if (map.isEmpty) {
+      throw Exception('Empty map');
```
g123k, line 144: You could, maybe, have a more relevant message like "Please add at least one AddableProductField"

lib/utils/AddableProductFields.dart
```diff
@@ -0,0 +1,17 @@
...
+
+extension AddableProductFieldExtension on AddableProductField {
+  static const Map<AddableProductField, String> _KEYS = {
+    AddableProductField.BRANDS: 'add_brands',
+    AddableProductField.STORES: 'add_stores',
+    AddableProductField.COUNTRIES: 'add_countries',
+  };
+
+  /// Returns the key of the product field
+  String get addableKey => _KEYS[this] ?? '';
```
g123k, line 16: Instead of an empty string, an Exception would be better, to let us know a case is missing

test/api_addProductFields_test.dart
```diff
@@ -0,0 +1,78 @@
...
+          user: TestConstants.TEST_USER,
+        );
+        expect(productResult.product, isNotNull);
+        expect(productResult.product!.brands, expectedBrands);
+      }
+    });
+  },
+      timeout: Timeout(
+        // some tests can be slow here
+        Duration(seconds: 90),
```
g123k, line 76: You can use const here (not that important)

g123k (changes requested):

monsieurtanuki:
@g123k ping

lib/openfoodfacts.dart
```diff
@@ -131,6 +132,37 @@ class OpenFoodAPIClient {
...
+    final String barcode,
+    final Map<AddableProductField, String> map, {
+    QueryType? queryType,
+  }) async {
+    if (map.isEmpty) {
+      throw Exception('Please add at least one AddableProductField');
+    }
+
+    final Map<String, String> parameterMap = <String, String>{};
+    parameterMap.addAll(user.toData());
```
g123k, line 148: By curiosity, why don't you call `Map.from(user.toData)`?

test/api_addProductFields_test.dart
```diff
@@ -0,0 +1,114 @@
...
+      late Status status;
+      late Product product;
+
+      // from scratch
+      product = initialProduct;
+      status = await OpenFoodAPIClient.saveProduct(user, product);
+      _checkStatus(status);
+
+      // cumulative list
+      String expectedValue = _getValue(field, product)!;
```
g123k, line 70: A `StringBuffer` would be better, since you concatenate multiple times

g123k (changes requested):

monsieurtanuki:
@g123k "Requested changes" don't look like an appropriate term to me in that case: "By curiosity", "would be better". We're talking about mere comments.
At least the way I see things, "Requested changes" is a bazooka when something problematic happens in the code, like obvious use-cases not taken into account or daft copy/pastes.

g123k:
> @g123k "Requested changes" don't look like an appropriate term to me in that case: "By curiosity", "would be better". We're talking about mere comments. At least the way I see things, "Requested changes" is a bazooka when something problematic happens in the code, like obvious use-cases not taken into account or daft copy/pastes.

True, but [this one](https://github.com/openfoodfacts/openfoodfacts-dart/pull/436#discussion_r846769473) requires a change

test/api_addProductFields_test.dart
monsieurtanuki: @g123k `StringBuffer` is never used in the whole project. Here we're talking about adding a String to a String twice, and checking the resulting value 3 times (that would mean 3 additional `toString()` calls with `StringBuffer`). In a test. That makes async calls to the back-end that take longer time than a String allocation.

M123-dev:
Ready to merge?

monsieurtanuki:
I'm afraid not.

test/api_addProductFields_test.dart
monsieurtanuki: @g123k I had a hunch about `StringBuffer` not being significantly faster than `String` when you only concatenate two `String`s and convert each time the result as a `String`.
I've coded a test with the 2 solutions: with `String` and with `StringBuffer`.
<details>
<summary>test code</summary>

```dart
      const String initialString = "init";
      const List<String> additionalValues = <String>['djobi', 'djoba'];
      final List<String> result = <String>[];

      late int start;
      late int end;

      result.clear();
      start = DateTime.now().microsecondsSinceEpoch;
      final StringBuffer asStringBuffer = StringBuffer(initialString);
      result.add(asStringBuffer.toString());
      for (final String additionalValue in additionalValues) {
        asStringBuffer.write(', $additionalValue');
        result.add(asStringBuffer.toString());
      }
      end = DateTime.now().microsecondsSinceEpoch;
      print('${end - start}: $result');

      result.clear();
      start = DateTime.now().microsecondsSinceEpoch;
      String asString = initialString;
      result.add(asString);
      for (final String additionalValue in additionalValues) {
        asString += ', $additionalValue';
        result.add(asString);
      }
      end = DateTime.now().microsecondsSinceEpoch;
      print('${end - start}: $result');
```
</details>

After several runs, I've noticed that the first block run is a bit slower than the other - regardless of `StringBuffer` and `String`. By approximately 1ms.
Can you afford that 1ms uncertainty for _tests_? I think so. If that's OK with you, please approve my PR. If you think there is a hidden performance potential, don't hesitate to keep this PR rejected and to run performance tests.

Issue #441, opened 2022-04-16
Sign up: limit usernames to 20 chars


g123k:
@see https://github.com/openfoodfacts/smooth-app/issues/1585

PR #442, opened 2022-04-16
feat: Username max length

g123k:
A username may not exceed 20 characters
#441


M123-dev (approved):
Looks good, thanks @g123k

Issue #440, opened 2022-04-16
Unhandled exception using OpenFoodAPIClient.searchProducts() with specific ProductSearchQueryConfiguration


christianfl:
**Describe the bug**
OpenFoodAPIClient.searchProducts() throws Unhandled Exception with a given ProductSearchQueryConfiguration. Occured searching for "**vitamin**".

**To Reproduce**

```
import 'package:openfoodfacts/model/UserAgent.dart';
import 'package:openfoodfacts/model/parameter/SearchTerms.dart';
import 'package:openfoodfacts/openfoodfacts.dart';
import 'package:openfoodfacts/utils/OpenFoodAPIConfiguration.dart';

var parameters = <Parameter>[
      const Page(page: 1),
      const PageSize(size: 10),
      const SortBy(option: SortOption.POPULARITY),
      SearchTerms(terms: ['vitamin']),
    ];

    ProductSearchQueryConfiguration configuration =
        ProductSearchQueryConfiguration(
            parametersList: parameters, language: OpenFoodFactsLanguage.GERMAN);

    SearchResult result =
        await OpenFoodAPIClient.searchProducts(null, configuration);
```

**Expected behavior**
result variable is filled with SearchResult without any unhandled exceptions.

**Screenshots**
If applicable, add screenshots to help explain your problem. Otherwise provide the error stacktraces:

```
E/flutter ( 7340): [ERROR:flutter/lib/ui/ui_dart_state.cc(209)] Unhandled Exception: type 'int' is not a subtype of type 'String?'
E/flutter ( 7340): #0      JsonHelper.imagesFromJson
package:openfoodfacts/utils/JsonHelper.dart:86
E/flutter ( 7340): #1      _$ProductFromJson
package:openfoodfacts/model/Product.g.dart:41
E/flutter ( 7340): #2      new Product.fromJson
package:openfoodfacts/model/Product.dart:417
E/flutter ( 7340): #3      _$SearchResultFromJson.<anonymous closure>
package:openfoodfacts/model/SearchResult.g.dart:15
E/flutter ( 7340): #4      MappedListIterable.elementAt (dart:_internal/iterable.dart:413:31)
E/flutter ( 7340): #5      ListIterator.moveNext (dart:_internal/iterable.dart:342:26)
E/flutter ( 7340): #6      new _GrowableList._ofEfficientLengthIterable (dart:core-patch/growable_array.dart:206:27)
[...]

PR #443, opened 2022-04-16
fix: coordinates_image_size is no subtype of String

M123-dev:
### What
- fixes: #440 

`JsonHelper.imagesFromJson` failed when the coordinates are `400` (as int) and not `full
`

``` 
E/flutter ( 7340): [ERROR:flutter/lib/ui/ui_dart_state.cc(209)] Unhandled Exception: type 'int' is not a subtype of type 'String?'
E/flutter ( 7340): #0      JsonHelper.imagesFromJson
package:openfoodfacts/utils/JsonHelper.dart:86
```

### DIFF
--- a/lib/utils/JsonHelper.dart
+++ b/lib/utils/JsonHelper.dart
@@ -84,7 +84,7 @@ class JsonHelper {
           JsonObject.parseInt(fieldObject['angle']),
         );
         final String? coordinatesImageSize =
-            fieldObject['coordinates_image_size'].toString();
+            fieldObject['coordinates_image_size']?.toString();
         final int? x1 = JsonObject.parseInt(fieldObject['x1']);
         final int? y1 = JsonObject.parseInt(fieldObject['y1']);
         final int? x2 = JsonObject.parseInt(fieldObject['x2']);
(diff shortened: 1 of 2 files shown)


lib/utils/JsonHelper.dart
```diff
@@ -84,7 +84,7 @@ class JsonHelper {
           JsonObject.parseInt(fieldObject['angle']),
         );
         final String? coordinatesImageSize =
-            fieldObject['coordinates_image_size'];
+            fieldObject['coordinates_image_size'].toString();
```
monsieurtanuki, line 87: Should be `fieldObject['coordinates_image_size']?.toString();`
For some reason that I don't understand, when `fieldObject['coordinates_image_size'] == null`, `fieldObject['coordinates_image_size'].toString()` returns the `String` `'null'` instead of crashing - and we expect the `null` value instead.

test/api_searchProducts_test.dart
```diff
@@ -142,6 +142,29 @@ void main() {
       expect(result.count, greaterThan(900));
     });
 
+    test('search products by keywords 2', () async {
```
monsieurtanuki, line 145: I would mention the issue: if you don't know the context you may think that this test is not very interesting.

monsieurtanuki (changes requested):
Thank you @M123-dev for the fix!
2 very important points though: please have a look at my comments (it will be quick).

M123-dev:
Thanks for your review @monsieurtanuki, applied your comments

monsieurtanuki (approved):
Thank you @M123-dev!

PR #444, opened 2022-04-18
feat: Barcodes validation for EAN8 & EAN13

### LINKED ISSUE #439: Validity of a barcode
On the Android app, [a barcode is valid if](https://github.com/openfoodfacts/openfoodfacts-androidapp/blob/develop/app/src/main/java/openfoodfacts/github/scrachx/openfood/utils/Utils.kt#L244):
- Its length is > 3
- Is an EAN 13
- Doesn't start with 977, 978 or 979

If I'm not wrong, we don't have these checks on Smoothie.
Should we implement them?

g123k:
OFF only accepts EAN8 and EAN13 barcodes.
This PR allows to check if a barcode is valid.

My implementation is freely based on https://commons.apache.org/proper/commons-validator/, to allow potential evolutions.

To check if a barcode is valid, just call `BarcodeValidator.isValid("1234")`.

Will fix #439

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,8 +101,8 @@ dart test
 Here is a list of known apps using this SDK:
 
 - **Smoothie**. [GitHub](https://github.com/openfoodfacts/smooth-app) / Will soon be released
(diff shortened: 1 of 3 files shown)


M123-dev (approved):
LGTM

Issue #447, opened 2022-04-19
Page(page: 10) not importing in openfoodfacts: ^1.14.0


Vishnu-Jmatics:
**Describe the bug**
`Page(page: 10)` not importing in openfoodfacts: ^1.14.0

**To Reproduce**
I'm developing a flutter application and it need the support of open-foods api for getting items by keyword.  For that I have set parameters like
 `PageSize(size: 1)`, 
`SearchTerms(terms: [searchText])`,
`SortBy(option: SortOption.PRODUCT_NAME)`,
**`Page(page: 10)`** 
etc. 

But **`Page(page: 10)`** shows syntax error always after I import 
import:

 `'package:openfoodfactsopenfoodfacts.dart';`

I don't know why this. I need to do this for pagination. Kindly let me know why this error so that I can forward with this feature. Thanks. 
      

**Screenshots**
![Screenshot 2022-04-19 120346](https://user-images.githubusercontent.com/45096096/163940610-d59131e0-d08c-4835-a2d6-cc6346dd1e16.png)

**package information:**
Openfoodfacts package version: `openfoodfacts: ^1.14.0`

g123k:
Hi @Vishnu-Jmatics,

Thanks for your report.
I have tested on a new project and I don't have any issues with this class.

Are you certain that in your file, you don't import another file with also exports a `Page`?

Vishnu-Jmatics:
@g123k  Thanks for the info. I 've checked it and shows this.

The name 'Page' is defined in the libraries 'package:flutter/src/widgets/navigator.dart' and 'package:openfoodfacts/model/parameter/Page.dart (via package:openfoodfacts/openfoodfacts.dart)'._ 

How can rectify this. Please help..

g123k:
In your imports just say:

```
import 'package:flutter/src/widgets/navigator.dart' hide Page;
```

M123-dev:
Otherwise if you at some point need both `Page` classes in a single file you can replace `hide Page;` with for example `as off;` then all imports from openfoodfacts would start with off.

`off.SearchTerms
off.Page
...
`

Vishnu-Jmatics:
@g123k Yey great. _import 'package:flutter/material.dart'_ is not directly imported in file but I supposed it resides in _'package:flutter/material.dart'_. I 've replaced it it by **_import  'package:flutter/material.dart' hide Page;_**. Thanks for ur valuable response, I can forward with this.

[...]

PR #449, opened 2022-04-19
fix: #447 - new class "PageNumber" replacing  ambiguous "Page" parameter

monsieurtanuki:
New file:
* `PageNumber.dart`: "Page number" search API parameter.

Impacted files:
* `api_searchProducts_test.dart`: now using `PageNumber` instead of deprecated `Page`
* `openfoodfacts.dart`: exports new class `PageNumber`
* `Page.dart`: deprecated in favor of less ambiguous new class `PageNumber`
* `ProductListQueryConfiguration.dart`: now using `PageNumber` instead of deprecated `Page`

### What
- in #447 a developer had issues with a class named "Page" being in both openfoodfacts AND in `material.dart`
- there are ways to deal with that kind of issues, but it's fair to prevent bad interactions with capital packages like `material.dart`
- the solution was to create a new class with a less common name (`PageNumber`) and to deprecate the old one (`Page`)

### Fixes bug(s)
- #447

### DIFF
--- a/lib/model/parameter/Page.dart
+++ b/lib/model/parameter/Page.dart
@@ -1,14 +1,9 @@
-import 'package:openfoodfacts/interface/Parameter.dart';
+import 'package:openfoodfacts/model/parameter/PageNumber.dart';
 
-/// "Page number" search API parameter
-class Page extends Parameter {
-  @override
-  String getName() => 'page';
-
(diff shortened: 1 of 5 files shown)


M123-dev (approved):
Looks good

PR #456, opened 2022-04-23
feat: Expose to Attribute related panelID

M123-dev:
### What
- Exposing to Attribute related panelID

### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/1029

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,9 @@ ios/
 *.ipr
 .idea/
 
+# VSCode
+.vscode/
+
 # eclipse project file
 .settings/
 .classpath
--- a/lib/model/Attribute.dart
+++ b/lib/model/Attribute.dart
@@ -13,6 +13,7 @@ class Attribute extends JsonObject {
     this.descriptionShort,
     this.match,
     this.status,
+    this.panelID,
   });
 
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -24,6 +24,9 @@ ios/
 *.ipr
 .idea/
 
+# VSCode
+.vscode/
+
```
monsieurtanuki, line 29: ???

lib/model/Attribute.dart
```diff
@@ -107,6 +111,7 @@ class Attribute extends JsonObject {
   final String? descriptionShort;
   final double? match;
   final String? status;
+  final String? panelID;
```
monsieurtanuki, line 114: I would say `panelId` instead.

monsieurtanuki (approved):
@M123-dev Looks good to me! I have minor comments: read them but feel free to ignore them.

lib/model/Attribute.dart
M123-dev: 👍🏼

.gitignore
M123-dev: Tried dart in vscode today and yeah thils folder was missing in the gitignore

Issue #466, opened 2022-05-19
Implement new algorithm to compute how products match users food preferences


monsieurtanuki:
cf. https://github.com/openfoodfacts/smooth-app/issues/1894
cf. https://github.com/openfoodfacts/openfoodfacts-server/blob/main/html/js/product-search.js

monsieurtanuki:
Fixed by #467.

PR #467, opened 2022-05-20
feat: #466 - matched product v2

monsieurtanuki:
New files:
* `api_matchedProductV1_test.dart`: Tests around Matched Product v1.
* `api_matchedProductV2_test.dart`: Tests around Matched Product v2.
* `api_productPreferences_test.dart`: Tests around Product Preferences. (unrelated)
* `matched_product_v2.dart`: Match and score of a `Product` vs. Preferences, v2.

Impacted files:
* `api_getProduct_test.dart`: refactored moving code to new file `api_matchedProductV1_test.dart`
* `api_searchProducts_test.dart`: unrelated minor refactoring
* `product_preferences_manager.dart`: unrelated new method `getAttributeIdsWithImportance`

### What
- Implementation of the V2 of `MatchedProduct`

### Fixes bug(s)
- #466

### DIFF
--- a/lib/personalized_search/matched_product_v2.dart
+++ b/lib/personalized_search/matched_product_v2.dart
@@ -12,22 +12,22 @@ import 'package:openfoodfacts/personalized_search/product_preferences_manager.da
 
 enum MatchedProductStatusV2 {
   /// Very good match (very good score)
-  VERY_GOOD,
+  VERY_GOOD_MATCH,
 
   /// Good match (good score)
-  GOOD,
+  GOOD_MATCH,
 
   /// Poor match (poor score)
-  POOR,
+  POOR_MATCH,
 
   /// Unknown match (at least 50% unknown attributes or one mandatory unknown)
-  UNKNOWN,
+  UNKNOWN_MATCH,
 
   /// May not match (at least one mandatory attribute that may not match)
-  MAY,
+  MAY_NOT_MATCH,
 
   /// Does not match (at least one mandatory attribute that does not match)
-  NO,
+  DOES_NOT_MATCH,
 }
 
 class MatchedProductV2 {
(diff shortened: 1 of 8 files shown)


monsieurtanuki:
@stephanegigandet or whoever is going to review the code, the main part is in `matched_product_v2.dart` - to be compared with the `js` version (https://github.com/openfoodfacts/openfoodfacts-server/blob/main/html/js/product-search.js or https://openfoodfacts.org/js/product-search.js, whatever).

lib/personalized_search/matched_product_v2.dart
```diff
@@ -0,0 +1,180 @@
...
+  GOOD,
+
+  /// Poor match (poor score)
+  POOR,
+
+  /// Unknown match (at least 50% unknown attributes or one mandatory unknown)
+  UNKNOWN,
+
+  /// May not match (at least one mandatory attribute that may not match)
+  MAY,
```
stephanegigandet, line 27: Can we keep the full names? VERY_GOOD_MATCH, MAY_NOT_MATCH etc.? I think it would be clearer. And it's easier if all SDKs use the same names.

stephanegigandet (approved):
Looks good to me, thank you!

stephanegigandet:
Something we may want to do in the future is to keep a list of the attributes that make the status does_not_match or may_not_match (e.g. to tell users "this product is not a match for you because it is not vegetarian and it contains gluten"), that's why in the JS code we used a list of attributes for each status, and not booleans. But we can always do it later if needed.

monsieurtanuki:
Thank you @stephanegigandet for your code review!
I've switched to "full" names like "DOES_NOT_MATCH" instead of "NO".

> that's why in the JS code we used a list of attributes for each status, and not booleans

Damn it, I was so proud of my code optimization :)
I understand the purpose and as you said there won't be any problem if we want to add those detailed lists on top of the current implementation.
Actually, after re-reading the `js` code it's probable that I dismissed some aspects of the code, as `js` is rather relaxed about declaring fields and my `java`/`flutter` instinct just ignores not-declared fields.
First, let's merge that `dart` code, let Smoothie use it with colors, and then we'll upgrade the `dart` code if needed.

PR #471, opened 2022-05-26
Fix new lints

M123-dev:
### What
- Updated lints


peterwvj (approved):

M123-dev:
Thanks for the quick review @peterwvj

PR #472, opened 2022-05-26
fix: Just send user id in from header

M123-dev:
### What
- The author is currently not honored for fixes since we send the whole object not just the user id


peterwvj (approved):

M123-dev:
Thanks @peterwvj

PR #483, opened 2022-06-15
fix: Make barcode not nullable

M123-dev:
### What
- Make product barcodes not nullable
- Forced barcode as product field as package would crash if we force a barcode but don't get it
- For: #2261


lib/model/ProductResult.dart
```diff
@@ -8,13 +9,17 @@ part 'ProductResult.g.dart';
 class ProductResult extends JsonObject {
   final int? status;
   @JsonKey(name: 'code')
-  final String? barcode;
+  final String barcode;
```
monsieurtanuki, line 12: What if no result?

monsieurtanuki (commented):
Hi @M123-dev!
Not convinced it's something we can do casually, IMHO that's typically something that would trigger a new major version. That's why I would have waited until we do some other major refactoring, like moving/renaming the files in order to get a max score of 130/130.
And when I look at your changes, I even wonder if it's a good idea, after all, to make barcode non-nullable, given the possible use-cases, not for smoothie but in general.

M123-dev:
I agree @monsieurtanuki thanks for the review. Lets close for now

Issue #482, opened 2022-06-15
Add "app_platform" in user agent, along with app_name and app_version


monsieurtanuki:
**Additional context**
cf. https://github.com/openfoodfacts/smooth-app/issues/2262 https://github.com/openfoodfacts/smooth-app/pull/2274

PR #484, opened 2022-06-16
fix: #482 add user agent parameters

cli1005:
### What
add user agent parameters:
app_platform = userAgent.system
comment = combined string 

### Screenshot
Capture d’écran 2022-06-16 à 14 47 45

### Fixes bug(s)
- #482


monsieurtanuki (approved):
Hi @cli1005!
As you said, assuming that the server understands both `app_platform` and `comment`, we're good!

PR #489, opened 2022-06-21
feat: Check every hour that the API works welll

teolemon:
feat: Check every hour that the server is up, and that some change hasn't broken the SDK

### What
- feat: Check every hour that the server is up, and that some change hasn't broken the SDK

### DIFF
--- a/.github/workflows/heartbeat.yml
+++ b/.github/workflows/heartbeat.yml
@@ -1,0 +1,39 @@
+name: test-sdk every hour
+
+on:
+  schedule:
+    - cron: "*/60 * * * *"
+concurrency:
+  group: tests
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Chekout code
+        uses: actions/checkout@v3
+
+      - name: Setup Dart
+        uses: dart-lang/setup-dart@v1
+
+      - name: Install dependencies
+        run: dart pub get
+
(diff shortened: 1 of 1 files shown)


monsieurtanuki (commented):
I cannot understand fully this `.yml`, but my question is: what happens if it fails? Mail sent?
In addition to that, there are tests that work randomly, and if they are to be run every hour they should be fixed, skipped or rewritten. Unless we're happy with 24 new mails every day (for the record I have currently 29 mails in my inbox and I would like to keep it that way).

g123k:
Shouldn't we use dedicated monitoring tools instead of this GA?
Or maybe a service like Zapier

VaiTon:
@teolemon can we close this? I suggest using another repo if we want to use Github Workflows for server monitoring, although I don't think it's the best way to do this

M123-dev:
Actually this is already in place https://github.com/openfoodfacts/openfoodfacts-dart/blob/6394d094b3822e799abc844e827f484bf7390368/.github/workflows/test-sdk.yml#L7

I'm getting a gh notification every 12h for some tests failing but I agree there are better ways to do this

monsieurtanuki:
> I'm getting a gh notification every 12h for some tests failing but I agree there are better ways to do this

Btw for some reason the tests are also executed on my fork and they always fail. I get two failed mails every day.
I know there are worst things in life, especially these days, but if I could get rid of them I would appreciate that - any idea?

Issue #481, opened 2022-06-08
The following fields do not exist in off-dart products AFAIK


teolemon:
@teolemon I can work on "Packaging", but the following fields do not exist in off-dart products AFAIK:

- [ ] Origins of ingredients
- [x] Manufacturing places
- [x] EMB Codes
- [ ] Countries

_Originally posted by @monsieurtanuki in https://github.com/openfoodfacts/smooth-app/issues/2169#issuecomment-1149481786_

M123-dev:
Can this be closed @monsieurtanuki

monsieurtanuki:
* "origins" added in #509.
* "countries" were already in off-dart but were added later to Smoothie (in the default list of fields).

PR #490, opened 2022-06-21
feat: #481 - added product fields embCodes and manufacturingPlaces

monsieurtanuki:
### What
- added product fields embCodes and manufacturingPlaces

### Part of 
- #481


M123-dev (approved):
Thought I've already done it. Read it already that's certain.

And no I don't dissaprove anything here

PR #491, opened 2022-06-22
feat: saveProduct with language and country

monsieurtanuki:
Impacted file:
* `openfoodfacts.dart`: added language and country parameters for method `saveProduct`

### What
- That's a ["Aha!" moment](https://www.youtube.com/watch?v=djV11Xbc914): it looks like on the server side "tags'ed" data like countries or categories take the language/country in order to compute the corresponding "localized tag", from whatever the user typed in.
- So it sounds fair to add country and language to method `saveProduct`, doesn't it?
- That will help tremendously Smoothie regarding saving tags'ed data.


M123-dev (approved):
LGTM, thanks @monsieurtanuki 

Can't say if the server allows it but if so it's a great thing

monsieurtanuki:
@M123-dev Yes the server does accept those parameters, and even uses them!
When you say that category is 'pain de mie', for the moment in Smoothie we don't pass the language in `saveProduct`, therefore the server assumes it's English, and we get a painful "en:pain-de-mie" from the server as a tag and 'en:pain de mie' as a value (if we get the product in French).

PR #498, opened 2022-06-26
feat: #465 - new "limit" parameter for autocomplete suggestions

### LINKED ISSUE #465: Autocomplete suggestions limited to 25: not enough!
**Describe the bug**
* Method `getAutocompletedSuggestions` takes a taxonomy type and an input arguments (and a language).
* It returns all the taxonomy values from this type, and that match the input.
* The problem is that only 25 items max are returned, which can be problematic with common terms.

**To Reproduce**
1. Look for "`soja`" in French - go to https://world.openfoodfacts.org/cgi/suggest.pl?tagtype=categories&term=soja&lc=fr
2. You'll see only 3 results that match "`Boissons végétales de soja`"
1. Look for "`Boissons végétales de soja`" in French - go to https://world.openfoodfacts.org/cgi/suggest.pl?tagtype=categories&term=Boissons%20végétales%20de%20soja&lc=fr
2. You'll see 6 results that match "`Boissons végétales de soja`"

**Expected behavior**
* We would expect the 6 "`Boissons végétales de soja`" to be returned with the mere "`soja`" search.
* We could expect an additional parameter to go beyond 25 returned items.
* We could expect an additional returned value that gives the total number of items that match.
* That makes the method a bit difficult to trust, because it also expects a perfect match, and unless you're a genius you cannot imagine it should be "`Boissons végétales de soja`", "`Boissons à base de soja`" or "`Boissons végétales au soja`"
* Therefore we cannot realistically use that method IRW - I found the bug during unrelated tests.

**Additional context**
The problem obviously comes from the server, where `$limit = 25` forces the max result length.

monsieurtanuki:
Impacted files:
* `api_get_automcompleted_suggestions.dart`: additional test about the new `limit` parameter.
* `openfoodfacts.dart`: new `limit` parameter for autocomplete suggestions.

### What
- New `limit` parameter for autocomplete suggestions.
- Before that, the limit was always 25 suggestions max.
- Now the limit is a parameter. There's a max limit of 400 set on the server side.

### Fixes bug(s)
- Closes #465

### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/2380


M123-dev (approved):
Thanks @monsieurtanuki looks good, a simple assert for values over 400 could be added but that's not a priority

monsieurtanuki:
IMHO adding an `assert` would not be a great idea: we don't want the app to crash because we're not able to deliver on the server side. A warning would make sense - I've put one in the comment, and I don't know how we could add one in the code, given that we don't know which tools the developers use.

M123-dev:
> Flutter enables assertions in debug mode.

monsieurtanuki:
@M123-dev To me `assert` means something is really wrong, like "dividing by zero", and that the algorithm upstream is flawed ("this value should never be 0, or even then, we should have handle the 0 case before").
Here it's just that if ever there are more than 400 items on the server side that match the search, we'll return only (!) 400 items.
Following this logic we should rather assert that the value is positive. Anyway, that's the beauty of open source: if the developers find it strange that a negative value or a value greater than 400 does not fully work, they can have a look at the code.

PR #506, opened 2022-07-06
fix: toBeCompleted - fixed uri and parameters for to be completed and user queries

monsieurtanuki:
### What
- Fixed uri and parameters for "to be completed" and user queries.
- Before, the language parameter caused problems for "to be completed" queries

### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/2315


monsieurtanuki:
Thank you @M123-dev!

Issue #500, opened 2022-06-26
❌ test/api_saveProduct_test.dart: OpenFoodAPIClient add new products add new product test 5 (failed)


teolemon:
### What
❌ test/api_saveProduct_test.dart: OpenFoodAPIClient add new products add new product test 5 (failed)
  Expected: <1>
    Actual: <400>
  
  package:test_api                      expect
  test/api_saveProduct_test.dart 264:7  main.<fn>.<fn>

monsieurtanuki:
@teolemon The problem is probably on the server side.
Like "it's such a lousy product (very few data) that when we try to compute the scores, we fail (with a time-out)".

Even the website itself crashes when we just add a nutrient value: https://world.openfoodfacts.net/product/7340011364184/chili-beans-null (off/off)

When I try, I get a time-out:
```
504 Gateway Time-out
nginx/1.21.6
```

And if I go back to the product page, I see that my change has been made.

For the record, those were the nutrients currently with data:
* Energy (kJ)
* Fat
* Saturated fat
* Carbohydrates
* Sugars
* Proteins
* Alcohol
* Vitamin B12 (cobalamin)

And the completion states are:
* Done
  * Nutrition facts completed
  * Categories completed
  * Brands completed
  * Packaging completed
  * Product name completed
* To do
  * To be completed
  * Ingredients to be completed
  * Expiration date to be completed
  * Packaging code to be completed
  * Characteristics to be completed
  * Origins to be completed
  * Quantity to be completed
  * Photos to be uploaded

teolemon:
ok, closing

monsieurtanuki:
Now it's skipped.

PR #509, opened 2022-07-10
feat: added "origins" field to product

monsieurtanuki:
Impacted files:
* `Product.dart`: added "origins" field
* `Product.g.dart`: generated
* `ProductFields.dart`: added "origins" field

### What
- added "origins" field to product

### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/2501


M123-dev (approved):

M123-dev:
We should probably add simple tests for that though

monsieurtanuki:
Thank you @M123-dev!
Actually we should refactor the whole `getProduct` tests here, probably on PROD data.

M123-dev:
The next thing on the todo list

monsieurtanuki:
The next thing on the todo list should rather be tests that work, which is not the case anymore, cf. #500.

PR #511, opened 2022-07-11
feat: Improvements for Sign up errors

g123k:
The API only returns a `Status` object for the user registration. It contains a sentence in English to explain potential issues.

To improve this behavior, here is a proposal:
- Create a class extending `Status` to avoid breaking current clients
- Support all signup error cases
- Allow checking errors based on an `enum`

Will help to fix https://github.com/openfoodfacts/smooth-app/issues/2504


lib/model/SignUpStatus.dart
```diff
@@ -0,0 +1,85 @@
...
+      );
+    } else if (status.body!
+            .contains('The password needs to be a least 6 characters long.') ||
+        status.body!.contains(
+            'The password and confirmation password are different.')) {
+      return SignUpStatus._(
+        status: 400,
+        statusError: SignUpStatusError.INVALID_PASSWORD,
+        error:
+            'The e-mail address is already used by another user. Maybe you already have an account? You can reset the password of your other account.',
```
monsieurtanuki, line 57: In general, just put `status.body!`. Here in particular the String is not adequate, and we don't even really care about the String as the added value of your PR is the `enum`.

lib/openfoodfacts.dart
```diff
@@ -805,7 +806,7 @@ class OpenFoodAPIClient {
   /// Returns [Status.status] 201 = complete; 400 = wrong inputs + [Status.error]; 500 = server error;
   ///
   /// When creating a [producer account](https://world.pro.openfoodfacts.org/) use [orgName] (former requested_org) to name the Producer or brand
-  static Future<Status> register({
+  static Future<SignUpStatus> register({
```
monsieurtanuki, line 809: That's a breaking change, isn't it?
I guess you should deprecate this method and create another one. With additional comments about the new feature.

lib/model/SignUpStatus.dart
```diff
@@ -0,0 +1,85 @@
...
+        status: 400,
+        statusError: SignUpStatusError.UNKNOWN,
+        error: 'Unrecognized request error',
+      );
+    }
+  }
+}
+
+enum SignUpStatusError {
+  EMPTY_EMAIL,
```
monsieurtanuki, line 78: Never used, is it?

lib/model/SignUpStatus.dart
```diff
@@ -0,0 +1,85 @@
...
+  EMAIL_ALREADY_USED,
+  EMPTY_USERNAME,
```
monsieurtanuki, line 80: Never used, is it?

lib/model/SignUpStatus.dart
```diff
@@ -0,0 +1,85 @@
...
+        status: 500,
+        statusError: SignUpStatusError.UNKNOWN,
+        error:
+            'No response, open an issue here: https://github.com/openfoodfacts/openfoodfacts-dart/issues/new',
+      );
+    } else if (status.body!.contains('loggedin')) {
+      return SignUpStatus._(
+        status: 201,
+      );
+    } else if (status.body!
```
monsieurtanuki, line 35: Instead of this long string of `if`s I would recommend something that starts like this:
```dart
final Map<String, SignUpStatusError> _errorTexts = <String, SignUpStatusError>{
  'This username already exists, please choose another.':
  	SignUpStatusError.USERNAME_ALREADY_USED,
  'The e-mail address is already used.':
  	SignUpStatusError.EMAIL_ALREADY_USED,
  'The password needs to be a least 6 characters long.':
  	SignUpStatusError.INVALID_PASSWORD
  'The password and confirmation password are different.':
  	SignUpStatusError.INVALID_PASSWORD,
  'The user name must contain only unaccented letters, digits and dashes.':
  	SignUpStatusError.INVALID_USERNAME,
};
```

lib/openfoodfacts.dart
g123k: Actually the current one will still work, as the SignUpStatus does exactly was Status did before.
So I'm not sure a breaking change is required here

monsieurtanuki: Fair enough @g123k: the previous version of the method returned a `Status`, now it returns a `SignUpStatus` which is also a `Status`. For the record it has nothing to do with "does exactly was Status did before".

lib/model/SignUpStatus.dart
g123k: Actually it's possible to raise this kind of error.
But I don't know how to fix it without breaking everything.

Let's take the empty email address.
- If I check for this before sending the request -> not retro compatible
- If I check for this in the response, I will only have "invalid email"
In that case, I have to check for the error in the body + check with the data sent to the server

For you, what's the right way to implement this?

monsieurtanuki: I understand your questioning.
To me the method is supposed to return what the server said. The server never says that the email is empty: that's a check you have to do before.

This is what I imagine, assuming that you do this for good localizations:
* in the app
  * check if the email is empty, and localize "hey, the email is empty"
  * check if the username is empty, and localize "hey, the username is empty"
  * check if the password is empty, and localize "hey, the password is empty"
  * check if both passwords are equal, and localize "hey, both passwords must be identical"
* so now we're at the maximum of what we could do in the app; let's call the server
* the server now returns a `SignUpStatus` and we should localize each `enum` values

I don't see why this would not be retro-compatible - developers that used `Status` still get a `Status` and we just add data that they won't use.

You might have coded that directly in Smoothie in a first approach, just checking that everything was fine, with the `enum` and a `static` method that returns the `enum` value from the `Status`.
And once it's stabilized, move it to off-dart.

g123k:
Thanks a lot @monsieurtanuki for all your feedbacks.
I have made some changes, could you tell me if it's better now?

lib/model/SignUpStatus.dart
```diff
@@ -0,0 +1,121 @@
...
+        '<!-- end templates/web/common/includes/error_list.tt.html -->';
+
+    final int startIndex = body.indexOf(startLine);
+    final int endIndex = body.indexOf(endLine);
+
+    if (startIndex != -1 && endIndex != 1) {
+      return body.substring(startIndex + startLine.length, endIndex);
+    }
+
+    return null;
```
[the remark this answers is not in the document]
g123k: Should I return the whole HTML here instead?

lib/openfoodfacts.dart
```diff
@@ -846,31 +847,10 @@ class OpenFoodAPIClient {
...
-        status: 400,
-        error:
-            'The e-mail address is already used by another user. Maybe you already have an account? You can reset the password of your other account.',
-      );
-    } else {
-      return Status(status: 400, error: 'Unrecognized request error');
-    }
+    status.body?.indexOf(
+        '<!-- start templates/web/common/includes/error_list.tt.html -->');
+
```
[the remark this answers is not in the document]
g123k: Ooops, you're right, really sorry for this one

lib/model/SignUpStatus.dart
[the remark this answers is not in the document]
monsieurtanuki: It's better to return something we can run investigations on than a `null`. In the rare cases when the HTML is not at all what we expected.

g123k:
I have removed the extension (I tend to think we have to be consistent with the rest of the API) + passed the UNKNOWN value directly the `statusError`.

Please let me know if it's better 👍

lib/model/SignUpStatus.dart
```diff
@@ -0,0 +1,118 @@
...
+  /// [https://github.com/openfoodfacts/openfoodfacts-translations/blob/main/openfoodfacts-web/openfoodfacts-web-fr.po]
+  static const Map<String, SignUpStatusError> _errorTexts =
+      <String, SignUpStatusError>{
+    'This username already exists, please choose another.':
+        SignUpStatusError.USERNAME_ALREADY_USED,
+    'The e-mail address is already used.': SignUpStatusError.EMAIL_ALREADY_USED,
+    'The password needs to be a least 6 characters long.':
+        SignUpStatusError.INVALID_PASSWORD,
+    'The password and confirmation password are different.':
+        SignUpStatusError.INVALID_PASSWORD,
```
monsieurtanuki, line 95: Btw we could have 2 different `SignUpStatusError`s as those are 2 different cases.

lib/model/SignUpStatus.dart
```diff
@@ -0,0 +1,118 @@
...
+      return SignUpStatus._(
+        status: 400,
+        statusError: _findErrorInMap(status.body) ?? SignUpStatusError.UNKNOWN,
+        error: _extractErrorFromHTMLBody(status.body),
+      );
+    }
+  }
+
+  /// [_errorTexts] have values where have to call the [contains] method
+  static _findErrorInMap(String? sentence) {
```
monsieurtanuki, line 48:
```suggestion
  static SignUpStatusError _findErrorInMap(String? sentence) {
```

lib/model/SignUpStatus.dart
```diff
@@ -0,0 +1,118 @@
...
+      return null;
+    }
+
+    for (String errorLabel in _errorTexts.keys) {
+      if (sentence.contains(errorLabel)) {
+        return _errorTexts[errorLabel];
+      }
+    }
+
+    return null;
```
monsieurtanuki, line 59:
```suggestion
    return SignUpStatusError.UNKNOWN;
```
This is important.

lib/model/SignUpStatus.dart
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  /// Try to extract the error from the HTML… or will return the [body] otherwise
+  static String? _extractErrorFromHTMLBody(String? body) {
```
monsieurtanuki, line 63: Now that I'm thinking we would probably be better off just `return`ing the body: same behavior as before, no change.
Not very important though.

lib/model/SignUpStatus.dart
g123k: We only have one field for the password.
On the client side, it will be the same

lib/model/SignUpStatus.dart
g123k: You're right, we should also add the global rules for Lint

lib/model/SignUpStatus.dart
```diff
@@ -0,0 +1,118 @@
...
+            'No response, open an issue here: https://github.com/openfoodfacts/openfoodfacts-dart/issues/new',
+      );
+    } else if (status.body!.contains('loggedin')) {
+      return SignUpStatus._(
+        status: 201,
+      );
+    } else {
+      return SignUpStatus._(
+        status: 400,
+        statusError: _findErrorInMap(status.body) ?? SignUpStatusError.UNKNOWN,
```
monsieurtanuki, line 41:
```suggestion
        statusError: _findErrorInMap(status.body),
```

monsieurtanuki (approved):
That's good @g123k!
Sorry for my last "this is important" comment, I hadn't noticed your `?? UNKNOWN`.

lib/model/SignUpStatus.dart
```diff
@@ -0,0 +1,118 @@
...
+      return SignUpStatus._(
+        status: 400,
+        statusError: _findErrorInMap(status.body),
+        error: _extractErrorFromHTMLBody(status.body),
+      );
+    }
+  }
+
+  /// [_errorTexts] have values where have to call the [contains] method
+  static SignUpStatusError _findErrorInMap(String? sentence) {
```
monsieurtanuki, line 48:
```suggestion
  static SignUpStatusError? _findErrorInMap(String? sentence) {
```

lib/model/SignUpStatus.dart
```diff
@@ -0,0 +1,118 @@
...
+      return null;
+    }
+
+    for (String errorLabel in _errorTexts.keys) {
+      if (sentence.contains(errorLabel)) {
+        return _errorTexts[errorLabel];
+      }
+    }
+
+    return SignUpStatusError.UNKNOWN;
```
monsieurtanuki, line 59:
```suggestion
    return null;
```

lib/model/SignUpStatus.dart
```diff
@@ -0,0 +1,118 @@
...
+            'No response, open an issue here: https://github.com/openfoodfacts/openfoodfacts-dart/issues/new',
+      );
+    } else if (status.body!.contains('loggedin')) {
+      return SignUpStatus._(
+        status: 201,
+      );
+    } else {
+      return SignUpStatus._(
+        status: 400,
+        statusError: _findErrorInMap(status.body),
```
monsieurtanuki, line 41:
```suggestion
        statusError: _findErrorInMap(status.body) ?? SignUpStatusError.UNKNOWN,
```

monsieurtanuki (approved):
Dancing tango with null parameters...

g123k:
I'm not sure where we are now ^^
Is-it ok or do still have some comments?

monsieurtanuki:
@g123k Actually I was right about `return`ing the `body` without changing it: our off-dart tests fail.

g123k:
I will change a little bit by implementation, because multiple can be trigerred (e.g.: invalid email + username)

g123k:
Are my changes OK in this PR?

monsieurtanuki:
@g123k Your changes are OK. The tests fail, but not your tests.

g123k:
So possible to merge it?
It will allow me to work on The Smoothie ´s ticket

monsieurtanuki:
> So possible to merge it?

Yes it's possible. I thought you had the privileges for that.

> It will allow me to work on The Smoothie ´s ticket

No it won't, because for that you would need a new _release_. Do you need a new release of off-dart too?
For the record you can also - temporarily - work with a local version of off-dart, or embed your new off-dart code temporarily in smoothie. Much faster and safer, in case you have something to fix in your new off-dart code.

That said, please confirm that you need a new release of off-dart.

Issue #450, opened 2022-04-20
pub.dev score from 130 to 120 :(


monsieurtanuki:
**Describe the bug**
The pub.dev score went down from 130 to 120.

**To Reproduce**
Go to https://pub.dev/packages/openfoodfacts/score
Alternate solution:
1. edit `analysis_options.yaml`
3. comment all the errors and all the rules, in order to more or less match the pub.dev parameters
5. go to the package root folder, and run `dart analyze | grep lowercase_with_underscores | grep -v 'test/'`
```
   info - lib/interface/JsonObject.dart:1:1 - Name source files using `lowercase_with_underscores`. - file_names
   info - lib/interface/Parameter.dart:1:1 - Name source files using `lowercase_with_underscores`. - file_names
   info - lib/model/Additives.dart:1:1 - Name source files using `lowercase_with_underscores`. - file_names
   info - lib/model/Agribalyse.dart:1:1 - Name source files using `lowercase_with_underscores`. - file_names
   info - lib/model/Allergens.dart:1:1 - Name source files using `lowercase_with_underscores`. - file_names
   info - lib/model/Attribute.dart:1:1 - Name source files using `lowercase_with_underscores`. - file_names
   info - lib/model/AttributeGroup.dart:1:1 - Name source files using `lowercase_with_underscores`. - file_names
   info - lib/model/BadgeBase.dart:1:1 - Name source files using `lowercase_with_underscores`. - file_names
   info - lib/model/EcoscoreAdjustments.dart:1:1 - Name source files using `lowercase_with_underscores`. - file_names
   info - lib/model/EcoscoreData.dart:1:1 - Name source files using `lowercase_with_underscores`. - file_names
   info - lib/model/EnvironmentImpactLevels.dart:1:1 - Name source files using `lowercase_with_underscores`. - file_names
   info - lib/model/EventsBase.dart:1:1 - Name source files using `lowercase_with_underscores`. - file_names
   info - lib/model/Ingredient.dart:1:1 - Name source files using `lowercase_with_underscores`. - file_names
   info - lib/model/IngredientsAnalysisTags.dart:1:1 - Name source files using `lowercase_with_underscores`. - file_names
   info - lib/model/Insight.dart:1:1 - Name source files using `lowercase_with_underscores`. - file_names
   info - lib/model/KeyStats.dart:1:1 - Name source files using `lowercase_with_underscores`. - file_names
[...]

PR #512, opened 2022-07-12
fix: summer cleaning

monsieurtanuki:
Deleted files:
* `ContainsAdditives.dart`: deprecated from 2021-07-21
* `OutputFormat.dart`: deprecated from 2021-07-13
* `PnnsGroupQueryConfiguration.dart`: deprecated from 2021-07-13
* `SearchSimple.dart`: deprecated from 2021-07-13

Impacted files:
* `AbstractQueryConfiguration.dart`: removed parameter `lc` (deprecated from 2021-07-20)
* `api_getProduct_test.dart`: added a timeout
* `api_ocrIngredients_test.dart`: added a timeout; changed the source barcodes
* `api_saveProduct_test.dart`: changed a test
* `api_searchProducts_test.dart`: added a timeout
* `InvalidBarcodes.dart`: refreshed the date (same data)
* `openfoodfacts.dart`: refactored around removed classes and parameters
* `ProductListQueryConfiguration.dart`: removed parameter `lc` (deprecated from 2021-07-20)
* `ProductQueryConfigurations.dart`: removed parameter `lc` (deprecated from 2021-07-20)
* `ProductSearchQueryConfiguration.dart`: removed parameter `lc` (deprecated from 2021-07-20)

### What
- Summer cleaning
  - deleted deprecated classes and methods
  - fixed some tests


lib/utils/ProductListQueryConfiguration.dart
```diff
@@ -12,8 +12,6 @@ class ProductListQueryConfiguration extends AbstractQueryConfiguration {
     this.barcodes, {
     final OpenFoodFactsLanguage? language,
     final List<OpenFoodFactsLanguage> languages = const [],
-    @Deprecated('Use parameters language or languages instead')
-        final String? lc,
     @Deprecated('Use parameter country instead') final String? cc,
```
M123-dev, line 15: cc

lib/utils/ProductSearchQueryConfiguration.dart
```diff
@@ -11,16 +11,13 @@ class ProductSearchQueryConfiguration extends AbstractQueryConfiguration {
   ProductSearchQueryConfiguration({
     final OpenFoodFactsLanguage? language,
     final List<OpenFoodFactsLanguage> languages = const [],
-    @Deprecated('Use parameters language or languages instead')
-        final String? lc,
     @Deprecated('Use parameter country instead') final String? cc,
```
M123-dev, line 14: cc

M123-dev (commented):
Thanks @monsieurtanuki looks good.

Missing some cc's but still approving for github.

Also this is a breaking change, so it has to be `fix!:` and we should probably combine it with some other things

M123-dev (approved):

lib/utils/ProductListQueryConfiguration.dart
monsieurtanuki: `cc` is much younger, it's from 2021-11-15 (2021-07-20 for `lc`).

monsieurtanuki:
Technically speaking, you're right, it's a "breaking change".
That said, whatever was removed or changed in this PR was marked as deprecated one year ago. We don't have an explicit policy about that, do we?
I would keep the "breaking change" / "new major version number" for PR with more impacts, like #450.

M123-dev:
Agree with that 👍🏻

Issue #518, opened 2022-07-18
Format Exception When searching


Zubi1001:
**Describe the bug**

`My parameters are ["Olive","Oil"]`
```

  Future<SearchResult?> searchForProduct(List<String> searchTerms) async {
    try {
      ProductSearchQueryConfiguration configuration =
          ProductSearchQueryConfiguration(parametersList: [
        SearchTerms(terms: searchTerms),
        const SortBy(option: SortOption.POPULARITY)
      ], fields: [
        ProductField.ALL,
      ], language: OpenFoodFactsLanguage.ENGLISH);

      SearchResult result = await OpenFoodAPIClient.searchProducts(
          uc.openFoodUser, configuration,
          queryType: QueryType.TEST);

      return result;
    } catch (e, stacktrace) {
      log(e.toString());
      print(stacktrace);
      return null;
    }
  }
```

**Expected behavior**

```
**```
[log] FormatException: Unexpected character (at character 1)
      <!-- start templates/web/common/site_layout.tt.html -->
      ^
flutter: #0      _ChunkedJsonParser.fail (dart:convert-patch/convert_patch.dart:1383:5)
#1      _ChunkedJsonParser.parseNumber (dart:convert-patch/convert_patch.dart:1250:9)
#2      _ChunkedJsonParser.parse (dart:convert-patch/convert_patch.dart:915:22)
#3      _parseJson (dart:convert-patch/convert_patch.dart:35:10)
#4      JsonDecoder.convert (dart:convert/json.dart:612:36)
#5      JsonCodec.decode (dart:convert/json.dart:216:41)
#6      OpenFoodAPIClient.getProducts
package:openfoodfacts/openfoodfacts.dart:312
<asynchronous suspension>
#7      FoodService.searchForProduct
package:affili_fit/…/foodService/food_service.dart:36
<asynchronous suspension>
#8      _TestGroundState.build.<anonymous closure>
package:affili_fit/main.dart:61
<asynchronous suspension>
```**
```
**package information:**
Openfoodfacts package version:

**Additional context**

monsieurtanuki:
Hi @Zubi1001!

I've just tested and that's OK for me:
* it's maybe a "bad luck" error, as the TEST env is notoriously unstable
* I would suggest you to work on PROD instead, unless you have good reasons to work on TEST

[...]

Issue #420, opened 2022-03-21
Selection / cropping of photos is not attributed to users (user_id and password fields not sent)


stephanegigandet:
When images are selected / cropped, the packaged does not pass the user_id and password field, so the selection is not attributed to the user.

Related to https://github.com/openfoodfacts/openfoodfacts-server/issues/6496

Fix: send the user credentials in the body of POST requests.

Issue #414, opened 2022-03-15
It's probably best to use POST for all requests


monsieurtanuki:
**Is your feature request related to a problem? Please describe.**
cf. @stephanegigandet's comment in #407.

**Describe the solution you'd like**
The impact would be all GET queries in `openfoodfacts.dart`, that should be changed into POST queries.

**Describe alternatives you've considered**
In most cases we don't _need_ it because there are only rare cases where we can reach the 8000 character limit for a URL.
But that would look less artisanal without all the parameters in the GET URL.

**Additional context**
All parameters must be in the same side; in our case, in the POST parameter side.
Could solve #354 too.
It would also be probably a good idea to add `Uri getUri({final QueryType? queryType});` to `AbstractQueryConfiguration`.
For test purposes, a new method in `HttpHelper` that creates a GET URL with the same parameters as method `doPostRequest` could be helpful.

monsieurtanuki:
ping @AshAman999

AshAman999:
@monsieurtanuki  I looked into the file where this code was and can probably come up with the solution, will raise a pr when done, though i am not sure about unit testing of the code as i have not done before, should i go ahead with my work?

monsieurtanuki:
@AshAman999 I'm a bit worried when you say "come up with the solution": in my mind you just have to change each `doPostRequest` with a `doPostRequest`, and there should be no impact at all on unit tests.
I suggest that you just do one change, e.g. for method `getProductString`, create a PR and we'll see if it's the good direction.

AshAman999:
Okay, 
I will raise a pr soon

PR #519, opened 2022-07-18
fix: Better error message when searching with wrong credentials

M123-dev:
### What
- Fixes: #518
- I analysed the response html and checked, the only thing I could find which is reliable for checking if its a credentials problem even for different languages was the amount of links to `reset_password.pl`


teolemon:
😅 👌

lib/utils/AbstractQueryConfiguration.dart
```diff
@@ -118,7 +117,6 @@ abstract class AbstractQueryConfiguration {
       // ignore: deprecated_member_use_from_same_package
       OpenFoodAPIConfiguration.computeCountryCode(country, cc);
 
-  @protected
```
monsieurtanuki, line 121: Please put it back.

test/api_wrong_credentials_search_test.dart
```diff
@@ -0,0 +1,28 @@
+import 'package:openfoodfacts/model/parameter/SearchTerms.dart';
+import 'package:openfoodfacts/openfoodfacts.dart';
+import 'package:openfoodfacts/utils/QueryType.dart';
+import 'package:test/test.dart';
+
+void main() async {
+  test('Test for FormatException when searching with wrong credentials',
```
monsieurtanuki, line 7: You should also run the same test
* with the TEST user, in order to check that things will work fine when they are supposed to.
* with a decent user but a wrong password (I assume that the user `sdfsdf` does not exists)

lib/openfoodfacts.dart
```diff
@@ -307,7 +307,20 @@ class OpenFoodAPIClient {
   }) async {
     final Response response = await configuration.getResponse(user, queryType);
     final String jsonStr = _replaceQuotes(response.body);
-    final SearchResult result = SearchResult.fromJson(json.decode(jsonStr));
+    final SearchResult result;
+    try {
+      result = SearchResult.fromJson(json.decode(jsonStr));
+    } on FormatException {
+      if (jsonStr.allMatches('/cgi/reset_password.pl').length == 3) {
```
monsieurtanuki, line 314: I don't speak regexp and I don't know what is expected in the `String`: you're looking for `'/cgi/reset_password.pl'`, right?
Maybe if you're looking for `'\/cgi\/reset_password.pl'` and don't use your `== 3` test it will be easier to read.

lib/openfoodfacts.dart
```diff
@@ -307,7 +307,20 @@ class OpenFoodAPIClient {
...
+        throw FormatException(
+          'Format error, likely a wrong username or password',
+        );
+      }
+      throw FormatException(
```
monsieurtanuki, line 319: Open question: maybe we should find inspiration in @g123k's recent `SignUpStatus` (#511), that extends `Status`, in order to be able to say in all languages "your credentials are not ok".

monsieurtanuki (commented):
Hi @M123-dev!

Beyond the actual coding, maybe that's not fair from the server side to crash a search when the user credentials are wrong, given that the user credentials are optional. But that's another story.

teolemon:
@M123-dev @monsieurtanuki Why are we sending credentials on a read query in the first place ?

monsieurtanuki:
> @M123-dev @monsieurtanuki Why are we sending credentials on a read query in the first place ?

I guess that's a combination of adding user credentials on POST for good reasons (#420), and switching most queries from GET to POST for good reasons (#414).
Capture d’écran 2022-09-18 à 18 16 09

The solution would be to add an optional parameter to the `doPostRequest` method, stating if we want to add user credentials (defaults to `false`, would be `true` at least for photos) (today implicitly it's `true`, which causes the bug).
@teolemon Ok with that fix?

teolemon:
Looks reasonable on my side. 
Regular credentials are for : edits, photo upload
Robotoff has specific credentials and does not expect a user password

adilwahla (commented):
looks good

monsieurtanuki:
@M123-dev We changed a lot the rest of the code since when you initially created this PR.
Closing; please create a new PR if needed.

M123-dev:
No problem

Maybe I'll find time this weekend to follow up with the changes from the last couple of weeks

PR #528, opened 2022-07-23
chore: upgrade dependencies

lsaudon:
### What
- upgrade dependencies

There are 3 tests that fail, but they also fail on “main”.

### DIFF
--- a/example/pubspec.yaml
+++ b/example/pubspec.yaml
@@ -13,4 +13,4 @@ dependencies:
 
 dev_dependencies:
   test: any
-  lints: ^1.0.1
+  lints: ^2.0.0
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -8,16 +8,16 @@ environment:
   sdk: '>=2.12.0 <3.0.0'
 
 dependencies:
-  json_annotation: ^4.5.0
+  json_annotation: ^4.6.0
   http: ^0.13.4
   path: ^1.8.2
-  image: ^3.1.3
+  image: ^3.2.0
   meta: ^1.8.0
 
 dev_dependencies:
   analyzer: ">=4.0.0 <5.0.0"
-  build_runner: ^2.1.11
-  json_serializable: ^6.2.0
+  build_runner: ^2.2.0
+  json_serializable: ^6.3.1
   lints: ^2.0.0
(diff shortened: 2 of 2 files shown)


monsieurtanuki (commented):
Hi @lsaudon!
The changes are OK; there's just the name of the PR itself I'm not sure about.
If I remember well, we're supposed to prefix our PR names with either `fix: ` or `feat: `.
Not sure what `chore: ` would produce.
@M123-dev Your views on that?

lsaudon:
[@commitlint/config-conventional](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional#type-enum)
```
[
  'build',
  'chore',
  'ci',
  'docs',
  'feat',
  'fix',
  'perf',
  'refactor',
  'revert',
  'style',
  'test'
];
```

monsieurtanuki:
@lsaudon Indeed, but that doesn't mean that we here at off-dart handle all those prefixes.
For instance, here `fix: ` will trigger a minor version number jump, and `feat: ` a major breaking change version number jump.
What happens with `chore: ` here? I don't know.

M123-dev:
chore isn't supported in release-please which we use for changelog generation. I wouldn't say this is a problem thought.

I don't have a strong opinion if we should mention it in the changelog but we can always manually change the changelog before merging the pending release PR.


The only problem could be if we directly want to release this PR without any other changes as no PR would be created.
But looking at your #529 this isn't a problem


TL;DR; this should be fine

example/pubspec.yaml
```diff
@@ -13,4 +13,4 @@ dependencies:
 
 dev_dependencies:
   test: any
-  lints: ^1.0.1
+  lints: ^2.0.0
```
M123-dev, line 16: Also interesting that this doesn't has any effect on us. Looks like we have some great self made rules .

M123-dev (approved):

monsieurtanuki:
Thank you @M123-dev for your comment: in short the _only_ explicit added value of `fix: ` and `feat: ` are the automatic release logs (that we can manually edit) and version number jump?
So typically, either in the current PR by @lsaudon or in my #529, we probably shouldn't create a new version or add it to the release logs.

Issue #531, opened 2022-07-26
API v2: states_tags=en:origins-completed,en:packaging-to-be-completed


monsieurtanuki:
https://world.openfoodfacts.org/api/v2/search?states_tags=en:origins-completed,en:packaging-to-be-completed&fields=code,states_tags

PR #532, opened 2022-07-26
feat: #531 - new StatesTagsParameter search filter

monsieurtanuki:
New files:
* `State.dart`: States of a `Product`. To be used in search API, with `StatesTagsParameter`.
* `StatesTagsParameter.dart`: States Tags as completed or to-be-completed. Filter combo mode is 'AND'.

Impacted file:
* `api_searchProducts_test.dart`: added a group of tests around the new parameter `StatesTagsParameter`

### What
- New `StatesTagsParameter` search filter.
- Based on new `State` `enum`, that lists the possible states. For each state, you'll have to say if it's completed or to-be-completed.

### Closes issue
- #531

### DIFF
--- a/lib/model/parameter/StatesTagsParameter.dart
+++ b/lib/model/parameter/StatesTagsParameter.dart
@@ -1,7 +1,7 @@
 import 'package:openfoodfacts/interface/Parameter.dart';
 import 'package:openfoodfacts/model/State.dart';
 
-/// States Tags as completed or to-be-completed. Filter combo mode is 'AND'.
+/// States Tags as completed or to-be-completed.
 class StatesTagsParameter extends Parameter {
   @override
   String getName() => 'states_tags';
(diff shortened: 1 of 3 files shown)


lib/model/parameter/StatesTagsParameter.dart
```diff
@@ -0,0 +1,24 @@
+import 'package:openfoodfacts/interface/Parameter.dart';
+import 'package:openfoodfacts/model/State.dart';
+
+/// States Tags as completed or to-be-completed. Filter combo mode is 'AND'.
```
M123-dev, line 4: Again, can't tell what a "Filter combo mode" is

M123-dev (approved):

lib/model/parameter/StatesTagsParameter.dart
monsieurtanuki: @M123-dev Fair enough; I've changed the comment.

PR #534, opened 2022-07-26
fix: deprecated search methods in favor of searchProducts

monsieurtanuki:
Impacted files:
* `api_getProduct_test.dart`: added a timeout to a unit test
* `api_getToBeCompletedProducts_test.dart`: no longer using a now deprecated method
* `api_getUserProducts_test.dart`: no longer using a now deprecated method
* `api_matchedProductV2_test.dart`: no longer using a now deprecated method
* `api_saveProduct_test.dart`: flagged as "skip" a problematic unit test
* `api_searchProducts_test.dart`: no longer using a now deprecated method
* `openfoodfacts.dart`: refactored method `searchProducts` to the detriment of now deprecated methods `getProducts` and `getProductList`

### What
- While working on API v2, I noticed similarities in code that could be solved through refactoring.
- I also "fixed" a test (longer timeout duration)

### DIFF
--- a/lib/openfoodfacts.dart
+++ b/lib/openfoodfacts.dart
@@ -45,7 +45,6 @@ import 'utils/HttpHelper.dart';
 import 'utils/LanguageHelper.dart';
 import 'utils/ProductHelper.dart';
 import 'utils/ProductQueryConfigurations.dart';
-import 'utils/ProductSearchQueryConfiguration.dart';
 
 export 'events.dart';
 export 'folksonomy.dart';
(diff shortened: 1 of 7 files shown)


M123-dev (approved):
Heyy @monsieurtanuki thanks for all the PR's here, all of them look very good.

This one won't be included in the changelog automatically, we should remember to mention the deprecation, especially from getProduct.

monsieurtanuki:
I've just renamed the issue before the merge, in order to be included in the change-log, but that's too late?

monsieurtanuki:
By the way when I "update" the other PRs, there are no test run and I receive a "test-sdk: No jobs were run" email.

M123-dev:
Thats interesting, good to have in mind, I'll have a look

M123-dev:
```
test-sdk: .github#L1
Canceling since a higher priority waiting request for 'tests' exists
```

We have a check in the test-sdk so that not two test instanced can run at the same time to avoid conflicts

monsieurtanuki:
That's problematic, because I've just merged conflicts and there was not even a check that my merge was correct, syntax-wise.

PR #538, opened 2022-08-03
feat: new "Barcodes" parameter; deprecated configurations

monsieurtanuki:
New files:
* `Barcodes.dart`: "Barcodes" search API parameter.
* `BoolMapParameter.dart`: Abstract map of `bool` as `Parameter`.

Impacted files:
* `AllergensParameter.dart`: refactored as `BoolMapParameter<AllergensTag>`
* `api_getToBeCompletedProducts_test.dart`: minor refactoring around a deprecated class
* `api_matchedProductV2_test.dart`: minor refactoring around a deprecated class
* `api_searchProducts_test.dart`: fixed test duration; refactoring
* `openfoodfacts.dart`: minor refactoring around a deprecated class
* `ProductListQueryConfiguration.dart`: deprecated
* `StatesTagsParameter.dart`: refactored as `BoolMapParameter<State>`
* `ToBeCompletedConfiguration.dart`: deprecated

### What
- New `Barcodes` search parameter
- Deprecated both `ToBeCompletedConfiguration` and `ProductListQueryConfiguration` in favor of `ProductSearchQueryConfiguration` with adapted parameters.
- Refactored `AllergensParameter` and `StatesTagsParameter` around the new `BoolMapParameter`

### DIFF
--- a/lib/model/parameter/Barcodes.dart
+++ b/lib/model/parameter/Barcodes.dart
@@ -1,0 +1,16 @@
+import 'package:openfoodfacts/interface/Parameter.dart';
+
+/// "Barcodes" search API parameter.
+class Barcodes extends Parameter {
+  @override
+  String getName() => 'code';
+
+  @override
+  String getValue() => codes.join(',');
+
+  final List<String> codes;
+
+  Barcodes(final String code) : this.list([code]);
+
+  const Barcodes.list(this.codes);
+}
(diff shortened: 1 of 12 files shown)


lib/model/parameter/Barcodes.dart
```diff
@@ -0,0 +1,16 @@
+import 'package:openfoodfacts/interface/Parameter.dart';
+
+/// "Barcodes" search API parameter.
+class Barcodes extends Parameter {
```
M123-dev, line 4: Thats a bit generic, what about BarcodeParameter

M123-dev (approved):

lib/model/parameter/Barcodes.dart
monsieurtanuki: @M123-dev Good point. "Filter" would even be better, but the rest is rather named "Parameter".
Renamed.

Issue #541, opened 2022-08-07
Add tests for login + register


M123-dev:
**Describe the bug**
More info in https://github.com/openfoodfacts/smooth-app/issues/1239

We already have tests for failing registers (https://github.com/openfoodfacts/openfoodfacts-dart/blob/master/test/user_management_test.dart)

But
> It's ok to create users on .dev and .net

So we should test if creation really works

M123-dev:
p1 for smooth_app

PR #542, opened 2022-08-07
test: Signup tests

M123-dev:
### What
- Fixes: #541
- For: https://github.com/openfoodfacts/smooth-app/issues/1239


M123-dev:
Heyy @g123k @monsieurtanuki for some reason I can't get the signup to work. I always recieve a 400 code with and Unkown error, on the website the register works and the multipart parameters seem right (used the same generated credentials so the register process really did not work). Do you have any idea

monsieurtanuki:
@M123-dev No idea where the problem comes from. Just a general idea: in TEST env things don't work smoothly.

M123-dev:
So after way to long debugging of the authentication method we use as it was a 401 status code I found the problem. We didn't pass the queryType in the login method. Thats now fixed 🥳

M123-dev:
Thanks for your review @monsieurtanuki

PR #544, opened 2022-08-15
fix: login

monsieurtanuki:
### What
- The `login` method didn't return a correct result.
- A test was added to make it "more correct".
- It's better but not very safe - other implementations should be used instead (on the server side too?)

### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/2784


AshAman999 (approved):
This will surely work for now :)
A quick fix that needs to be merged ASAP

PR #545, opened 2022-08-18
fix: imageHelper - minor fix about Windows; deprecated methods

monsieurtanuki:
Impacted file:
* `ImageHelper.dart`: deprecated methods that we have no added-value about; fixed a method that wouldn't work on Windows

### What
- It's been a long time since I noticed that we used package `image` for nothing.
- I've deprecated its usage, as it's not what the package is all about.
- I've also fixed an incorrect usage of `path`.


M123-dev (approved):

monsieurtanuki:
Thank you @M123-dev!

PR #549, opened 2022-09-02
ci: improve bug template

teolemon:
### What
- ci: improve bug template

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,24 +7,24 @@ assignees: ''
 
 ---
 
-**Describe the bug**
+### Description
 A clear and concise description of what the bug is.
 
-**To Reproduce**
+### Steps to Reproduce
 Steps to reproduce the behavior:
 1. Go to '...'
 2. Click on '....'
 3. Scroll down to '....'
 4. See error
 
-**Expected behavior**
+### Expected behavior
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -7,24 +7,24 @@ assignees: ''
...
 2. Click on '....'
 3. Scroll down to '....'
 4. See error
 
-**Expected behavior**
+### Expected behavior
 A clear and concise description of what you expected to happen.
 
-**Screenshots**
+### Screenshots
```
monsieurtanuki, line 23: There are no screenshots for a `dart` package, right? Or in very rare cases, I can't think of one of them now.

M123-dev (approved):

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -7,24 +7,24 @@ assignees: ''
...
-**Describe the bug**
+### Description
 A clear and concise description of what the bug is.
 
-**To Reproduce**
+### Steps to Reproduce
 Steps to reproduce the behavior:
 1. Go to '...'
 2. Click on '....'
 3. Scroll down to '....'
```
monsieurtanuki, line 17: That would rarely be the case for a `dart` package, more something like a called method and its parameters.

monsieurtanuki (approved):

PR #554, opened 2022-09-12
ci: Fix tests

M123-dev:
### What
- FIxed not working tests

### DIFF
--- a/lib/utils/HttpHelper.dart
+++ b/lib/utils/HttpHelper.dart
@@ -32,26 +32,27 @@ class HttpHelper {
   static Map<String, String>? addUserAgentParameters(
     Map<String, String>? map,
   ) {
+    map ??= <String, String>{};
     if (OpenFoodAPIConfiguration.userAgent?.name != null) {
-      map ??= <String, String>{};
       map['app_name'] = OpenFoodAPIConfiguration.userAgent!.name!;
     }
     if (OpenFoodAPIConfiguration.userAgent?.version != null) {
-      map ??= <String, String>{};
       map['app_version'] = OpenFoodAPIConfiguration.userAgent!.version!;
     }
     if (OpenFoodAPIConfiguration.uuid != null) {
-      map ??= <String, String>{};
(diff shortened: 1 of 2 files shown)


monsieurtanuki (approved):
You re-coded the same method differently in both PRs. I prefer this PR's version.

M123-dev:
Thanks for your review @monsieurtanuki! I noticed this is failing as well:
```
❌ test/api_searchProducts_test.dart: OpenFoodAPIClient search products with completed/to-be-completed states check products with 2 random states tags (failed)
  Expected: a value less than or equal to <640854>
    Actual: <640947>
     Which: is not a value less than or equal to <640854>
  
  package:test_api                          expect
  test/api_searchProducts_test.dart 1020:7  main.<fn>.<fn>
```



``` dart
    test('check products with 2 random states tags', () async {
      final List<int> indices = _getRandomInts(
        count: 2,
        max: State.values.length,
      );
      final State state1 = State.values[indices[0]];
      final State state2 = State.values[indices[1]];
      final Random random = Random();
      final bool completed1 = random.nextBool();
      final bool completed2 = random.nextBool();
      final int count1 = await _checkStatesTags(
        StatesTagsParameter(map: {
          state1: completed1,
        }),
      );
      final int count2 = await _checkStatesTags(
        StatesTagsParameter(map: {
          state2: completed2,
        }),
      );
      final int countBoth = await _checkStatesTags(
        StatesTagsParameter(map: {
          state1: completed1,
          state2: completed2,
        }),
      );
      // it's an AND: with both conditions we always get less results.
      expect(countBoth, lessThanOrEqualTo(count1));
      expect(countBoth, lessThanOrEqualTo(count2));
    });
```

Do you know if it's fine to raise this to `640947`

monsieurtanuki:
We don't put fixed values here. We just assume that the number of results of "A && B" are less than just "A" or just "B".
I suggest you add the "map" as the comment parameter of method "expect" (more verbose) and run the test again.
I think this was just bad luck.

M123-dev:
I added the <State, complete> map as reason to the expect. I that what you suggested

PR #556, opened 2022-09-15
ci: Don't run ci daily

M123-dev:
### What

- https://github.com/openfoodfacts/openfoodfacts-dart/pull/489#issuecomment-124835497

There are multiple ways to just run on the root repo but I see no real added value in running that often


monsieurtanuki (commented):
That's not exactly what I meant: if one day we put that twice daily process, I guess it was meaningful then. If it's not meaningful anymore, sure we can get rid of it.
But if it's still meaningful, it's just that I don't want it on my fork. Should I just apply the same change to my own fork, and that's it?

M123-dev:
@monsieurtanuki sorry for this late response. I think I understood what you mean. You just don't wan't it daily on your fork and I understand that. But while thinking about it again I think we don't need it daily here either. (Or in other words two failed runs in my inbox daily)

We have some ways to fix that:

1. What I did here, remove the sceduled task alltogether

2. Make the tests more stable so that it doesn't fail all the time due to timeouts or something else, always getting a failing message isn't helpful at all.
In Germany there is a saying about a child playing in the forest and screaming that a wolf is attacking him to fool the villagers, who then come to help him. 
But one day, once the people were fooled enough times and just ignored this kid, the wolf really came, I think you can guess the rest. Same goes for the test. I don't check if 3 are failing due to  timeouts or 8 due to something is broken serverside.

3. The correct answer to your request, just don't running on your fork. The right way for this would be a     `if: github.owner == openfoodfacts` as I added some in smoothie [here](https://github.com/openfoodfacts/smooth-app/pull/3094/files)

M123-dev:
The squirrel really made me laugh, thanks for that @monsieurtanuki 😆 

> The main problem we experience with tests is with write operations, because of the lousy TEST env: sometimes they fail, sometimes they time out.

> 
> I don't know what we could do about the write operations / TEST env.
> We should definitely run all READ tests directly on the PROD env.

I understand the point for using prod for read request, but there we still have the problem of chaning data, to solve that we always have to write our data before but that on prod. I don't know of a good solution for it though. 



> Maybe we should write less tests and focus on what really matters for off-dart (e.g. not testing numerous crazy combinations of query parameters of the OFF server) (this is not our purpose).
> At least, get rid of redundant tests that last long.

Thats the best thing we can do right now. At some point we can split tests into pre and post submit tests but thats really not to be discussed here. So yeah lets continue what you started and fix our testing strategy by cleaning our tests.
But lets until then merge this pr, stop the daily runs and revert then.



> Today I've spent half my day trying to replay the tests, each time after minor unrelated fixes (spoiler alert: still not solved). Even github is fed up with it and finally queued my tests 2 hours ago (still not started):

I feel with you, I spend countless hours commiting and waiting for github to spit out if they are kind enouph to let my workflow work or not. But I've never come as far as seeing something beeing queued for thats long. But I'm wondering why did you have to run it on github actions and not locally.

PR #562, opened 2022-09-19
Updated Readme.md

yashsharma81:
### What
- 

### Fixes bug(s)
- 

### Part of 
-


yashsharma81:
This is temporary solutions for showing scores. so that UX doesn't compromise I'm working on API

M123-dev:
Heyy @yashsharma81 I think you meight missunderstood the problem. The score batches are working perfectly. The problem is with the license. I guess we edited the license so that it isn't automatically detected anymore. This should be fixed if we get the correct Apache License 2.0 text any only change the to be changed text fields

yashjani01:
@M123-dev okay i understood now

M123-dev:
I think we can close this for now, can't we @yashjani01 ?

If not feel free to reopen or create a new PR

Issue #566, opened 2022-09-29
Add nutrient input for "product as prepared"


monsieurtanuki:
### Why - Problem description
Same as #565, but for "product as prepared", we need to add those product fields:
* "nutrition_data_prepared" (either "on" or "")
* "nutrition_data_prepared_per" (either "100g" or "serving")

In addition to those 2 fields, we need to create ALL the corresponding nutrient fields (e.g. `energy-kcal_prepared_100g` and `energy-kcal_prepared_serving`).
We can either do it very stupidly (copy paste the "product as sold" fields), or very elegantly (e.g. instead of one packed `Nutriments` class, a `Product` would have 4 "light nutrient" objects - for "sold/100g", "sold/serving", "prepared/100g", "prepared/serving").
The way the JSON tool we use currently works and the fact that it would probably be a breaking change seem to discard the "elegant" approach. Too bad because later we should also deal with "imported" and "imported_prepared" values...

PR #570, opened 2022-10-04
feat: #566 - refactored Nutriments as Map with new Nutrient class

monsieurtanuki:
Deleted file:
* `Nutriments.g.dart`

New file:
* `Nutrient.dart`: Nutrient
* `PerSize.dart`: Used for nutrient values: for which size of the product?

Impacted files:
* `api_getProduct_test.dart`: refactored
* `api_getProductRaw_test.dart`: refactored
* `api_getRobotoff_test.dart`: unrelated test fix
* `api_saveProduct_test.dart`: refactored
* `api_searchProducts_test.dart`: unrelated test fix; refactored
* `Nutriments.dart`: heavy refactoring - now we use a simple Map, and most of the fields are deprecated
* `NutrimentsHelper.dart`: refactored
* `Product.dart`: removed `const`s that are now in `PerSize`

### What
- The whole idea is to refactor `Nutriments` as a `Map`, and not anymore as individual fields.
- This is the first step of a process; a next step would include nutrients for "as prepared" products.
- The code is hugely simplified, except for the _numerous_ deprecated field getters and setters.

### Part of 
- #566


lib/model/Nutrient.dart
```diff
@@ -0,0 +1,309 @@
...
+    Nutrient.erucicAcid: 'erucic-acid',
+    Nutrient.monounsaturatedFat: 'monounsaturated',
+    Nutrient.polyunsaturatedFat: 'polyunsaturated',
+    Nutrient.pantothenicAcid: 'pantothenic-acid',
+    Nutrient.omega3: 'omega-3-fat',
+    Nutrient.omega6: 'omega-6-fat',
+    Nutrient.transFat: 'trans-fat',
+  };
+
+  /// Default units for nutrients.
```
stephanegigandet, line 222: On the website, default units depends on countries. e.g. sodium is in grams in EU, and mg in US.

lib/model/PerSize.dart
```diff
@@ -0,0 +1,21 @@
+/// Used for nutrient values: for which size of the product?
+enum PerSize {
+  /// Per serving of product
+  serving,
+
+  /// Per 100 grams of product
+  oneHundredGrams,
```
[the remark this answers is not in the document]
monsieurtanuki: No we can't, `100g` is not a valid variable or enum name.

lib/model/Nutrient.dart
```diff
@@ -0,0 +1,309 @@
...
+  /// Omega 6
+  omega6,
+
+  /// Trans Fats
+  transFat,
+}
+
+/// Helper class around [Nutrient].
+extension NutrientExtension on Nutrient {
+  /// Tags for nutrients, in the cases when the labels are different.
```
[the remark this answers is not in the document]
monsieurtanuki: Oh the tags are the same; it's just that from the enum value `calcium` I can compute the tag `'calcium'`, but it's not that systematic for `vitaminB12` and `'vitamin-b12'`.
Maybe the confusion is that `/// Tags for nutrients, in the cases when the labels are different.` should rather be `/// Tags for nutrients, in the cases when the tag is not identical to the enum value.`
Not sure what you meant precisely here, but the thing is that we cannot name variables 100% like we want.

lib/model/Nutrient.dart
```diff
@@ -0,0 +1,309 @@
...
+    },
+  };
+
+  /// Returns the tag used in [Nutriments]'s map.
+  String get tag =>
+      _specialNutrientTags[this] ?? toString().replaceFirst('Nutrient.', '');
+
+  /// Returns the default unit of values.
+  Unit? get defaultUnit {
+    for (final MapEntry<Unit, Set<Nutrient>> entry in _defaultUnits.entries) {
```
stephanegigandet, line 302: This works, but why not use a simple map of nutrients to units so that it can be retrieved directly? (and that we can have the map defined in alphabetic order, to avoid bugs like having the same nutrient defined in 2 groups).

lib/model/Nutrient.dart
monsieurtanuki: > On the website, default units depends on countries. e.g. sodium is in grams in EU, and mg in US.

@stephanegigandet The thing is that:
* regardless of the country, the default unit provided there is a good (good enough?) start
* we do take into consideration the country when we use https://fr.openfoodfacts.org/cgi/nutrients.pl but there are no default units there, where default units would make sense

monsieurtanuki:
@stephanegigandet @M123-dev 
I've just upgraded to dart 2.17, and that's helped me write a much nicer code.
Regarding the unification of code all around OFF, we cannot bend all programming languages to respect a unique format, but my solution here is to add an explicit `offTag` variable, which is what should be common in OFF:
```dart
// ...
  /// Fats
  fat(typicalUnit: Unit.G, offTag: 'fat'),

  /// Saturated Fats
  saturatedFat(typicalUnit: Unit.G, offTag: 'saturated-fat'),

  /// Proteins
  proteins(typicalUnit: Unit.G, offTag: 'proteins'),

  /// Energy in kcal
  energyKCal(typicalUnit: Unit.KCAL, offTag: 'energy-kcal'),

  /// Energy in kj
  energyKJ(typicalUnit: Unit.KJ, offTag: 'energy-kj'),
//...
```
What do you think of that?

lib/model/Nutriments.dart
```diff
@@ -1,1480 +1,1704 @@
...
+    final PerSize perSize,
+  ) =>
+      _getValue(_getTag(nutrient, perSize));
+
+  double? _getValue(final String tag) =>
+      JsonObject.parseDouble(_getDynamicValue(tag));
+
+  dynamic _getDynamicValue(final String tag) => _map[tag];
+
+  /// Returns the value in grams of that [nutrient] for that [perSize].
```
stephanegigandet, line 55:
```suggestion
  /// Sets the value in grams of that [nutrient] for that [perSize].
```

stephanegigandet:
> my solution here is to add an explicit `offTag` variable, which is what should be common in OFF:

Sounds good, thanks.

monsieurtanuki:
@M123-dev @stephanegigandet The controversial thing about this PR is that we cannot set nutrient units anymore. It was a deliberate choice of mine, as I believe it creates confusion with a limited added value.

Feel free to approve this PR or to tell me what to change.

Side note: yesterday I was lucky enough to have the tests passed after half a day and that was painful :)

M123-dev (approved):
Codewise nothing that shocks me here therefore the approval.

Just one thing regarding the depreciation and dart version increase, I'd prefer to make it a "feat!:" as this has the potential to break apps or maybe just even ci pipelines as we have in Smoothie if they call `pub upgrade`

monsieurtanuki:
> Just one thing regarding the depreciation and dart version increase, I'd prefer to make it a "feat!:" as this has the potential to break apps or maybe just even ci pipelines as we have in Smoothie if they call `pub upgrade`

I cannot say that I don't agree with you, but still, I'm a bit reluctant to make it a breaking change "just for the dart version upgrade".
In addition to that, 95% of the pain in the neck for this PR was to keep old fields in order to avoid the breaking change.

Or, let's
* remove the old fields
* make it a breaking change
* quickly fix #450 (in a different PR of course)

@M123-dev What do you think about it?

M123-dev:
Then let's rather postphone it to when we have some other really breaking changes as well

monsieurtanuki:
> Then let's rather postphone it to when we have some other really breaking changes as well

I see nothing bigger than #450 that would justify a breaking change. And there will never be a good reason to do it at any given time.
As you said before (I think), we constantly postpone breaking changes.
In addition to that, I would like to merge this PR - no big rush but we need it for Smoothie.

M123-dev:
Yes, sorry I wasn't exact enough, I meant let's merge this one, possibly without a breaking change

monsieurtanuki:
Thank you @M123-dev!

PR #574, opened 2022-10-07
feat: #573 - getRandomRobotoffQuestion with optional insight types

### LINKED ISSUE #573: Allow to query Robotoff without specifying InsightTypes
### Description
While coding https://github.com/openfoodfacts/smooth-app/pull/3102, I had the necessity of getting 3 random questions from robotoff, without specifying the product nor the insight type. 

The method `getRandomRobotoffQuestion` has a required parameter `types` that even when provided with an empty list is still added  to the product query, making robotoff return 0 questions.

#### Example of the query:
```
https://robotoff.openfoodfacts.org/api/v1/questions/random?lang=en&count=3&insight_types
```

### Expected behavior
The method should accept a null value for `types` and not add it to the query.

monsieurtanuki:
Impacted files:
* `api_getRobotoff_test.dart`: added a test for "no `'insight_types'`"; refactored
* `openfoodfacts.dart`: potentially no `'insight_types'` in `getRandomRobotoffQuestion`

### What
- We needed to be able not to specify insight types  in `getRandomRobotoffQuestion`

### Fixes bug(s)
- Closes #573

### Part of 
- https://github.com/openfoodfacts/smooth-app/pull/3102

### DIFF
--- a/lib/openfoodfacts.dart
+++ b/lib/openfoodfacts.dart
@@ -585,17 +585,18 @@ class OpenFoodAPIClient {
       count = 1;
     }
 
-    List<String?> typesValues = [];
-    for (var t in types) {
-      typesValues.add(t.value);
+    final List<String> typesValues = [];
+    for (final InsightType t in types) {
+      final String? value = t.value;
+      if (value != null) {
+        typesValues.add(value);
+      }
     }
-
-    String parsedTypes = typesValues.join(',');
 
     final Map<String, String> parameters = <String, String>{
       'lang': lang,
       'count': count.toString(),
-      'insight_types': parsedTypes.toString()
+      if (typesValues.isNotEmpty) 'insight_types': typesValues.join(',')
     };
 
     var robotoffQuestionUri = UriHelper.getRobotoffUri(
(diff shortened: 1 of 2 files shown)


VaiTon (commented):
Hey @monsieurtanuki! Thank you for your fast fix! 🚀 

I have a suggestion about the parameters: why don't we make types default to `[]`, so that it can be left out when invoking the method and you get every insight by default?

monsieurtanuki:
> I have a suggestion about the parameters: why don't we make types default to `[]`, so that it can be left out when invoking the method and you get every insight by default?

You're right, I was afraid it would be a (minor) breaking change, but in that direction it would not: the existing code is currently forced to provide a parameter, and it won't anymore.
About to code it...

VaiTon (approved):
Thank you!

monsieurtanuki:
Thank you @VaiTon for your review and suggestions!

Cannot merge for the moment because of unrelated failing tests - either #566 needs to be approved or I'll have a look - again - at the failing tests (probably tomorrow).

monsieurtanuki:
@M123-dev That's good news! :)

PR #576, opened 2022-10-10
fix: KP panel groups sometimes have null title

monsieurtanuki:
Impacted files:
* `KnowledgePanelElement.dart`: minor fix on `KnowledgePanelPanelGroupElement`
* `KnowledgePanelElement.g.dart`: generated

### What
- The way it was coded, `KnowledgePanelPanelGroupElement` were supposed to have a mandatory `'title'`.
- This afternoon, that was not always the case in TEST env (https://fr.openfoodfacts.net/api/v2/product/3270160717323?fields=knowledge_panels): 
```json
"panel_group_element": {
	"panel_ids": ["nutrition_facts_table"]
}
```
- As a side effect, the product could not be downloaded.
- The PR is about making the title optional.
- What I can say about PROD is that for the same product, `'title'` is always populated (https://fr.openfoodfacts.orgapi/v2/product/3270160717323?fields=knowledge_panels)

### DIFF
--- a/lib/model/KnowledgePanelElement.dart
+++ b/lib/model/KnowledgePanelElement.dart
@@ -114,7 +114,7 @@ class KnowledgePanelImageElement extends JsonObject {
 @JsonSerializable()
 class KnowledgePanelPanelGroupElement extends JsonObject {
   /// Title of the panel group. Example: "Carbon Footprint" or "Labels" etc.
-  final String title;
+  final String? title;
 
   /// List of panel ids that belong to this panel group.
   @JsonKey(name: 'panel_ids')
--- a/lib/model/KnowledgePanelElement.g.dart
+++ b/lib/model/KnowledgePanelElement.g.dart
@@ -58,7 +58,7 @@ Map<String, dynamic> _$KnowledgePanelImageElementToJson(
 KnowledgePanelPanelGroupElement _$KnowledgePanelPanelGroupElementFromJson(
         Map<String, dynamic> json) =>
     KnowledgePanelPanelGroupElement(
-      title: json['title'] as String,
+      title: json['title'] as String?,
       panelIds:
           (json['panel_ids'] as List<dynamic>).map((e) => e as String).toList(),
     );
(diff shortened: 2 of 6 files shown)


VaiTon:
I think we need to know when and if the title attribute is an optional value, so I invoke the help of @stephanegigandet and @alexgarel

monsieurtanuki:
Null title is something that did happen in TEST env:
* is that normal from the server side? - good question, indeed we should ask them
* should product downloads crash when the title is null? I think not, we need to be more resilient
* Actually in Smoothie there was already a test "is title empty?", which kind of means that not populated titles are probable. Therefore I just added the null case to the "not null but empty" case
________________________________
De : VaiTon ***@***.***>
Envoyé : lundi 10 octobre 2022 19:19
À : openfoodfacts/openfoodfacts-dart ***@***.***>
Cc : monsieurtanuki ***@***.***>; Assign ***@***.***>
Objet : Re: [openfoodfacts/openfoodfacts-dart] fix: KP panel groups sometimes have null title (PR #576)


I think we need to know when and if the title attribute is an optional value, so I invoke the help of @stephanegigandet<https://github.com/stephanegigandet> and @alexgarel<https://github.com/alexgarel>

stephanegigandet:
For knowledge panels, pretty much every field is optional, if something is not there (e.g. title, subtitle, content), then the meaning is that it should not be displayed.

Panel groups can indeed have no title. They possibly could have an empty list of panels too.

I'm in the process of writing a JSON schema for knowledge panels:
https://github.com/openfoodfacts/openfoodfacts-server/tree/main/docs/reference/schemas/knowledge_panels

stephanegigandet:
> This is because of the website redesign, a few panels may have changed.

I will add empty titles in order not to break the currently deployed apps, but please keep your changes to make them optional.

M123-dev (approved):

monsieurtanuki:
Thanks to all of you guys!
I'm about to create a dedicated issue about resilience in off-dart.

PR #583, opened 2022-10-12
feat: #579 - new method to log in, that returns data about user

### LINKED ISSUE #579: New login method
### Why - Problem description
cf. https://github.com/openfoodfacts/smooth-app/issues/3118#issuecomment-1274220392
Code evolution on the server side: now the login (`'/cgi/auth.pl'`) may return a JSON (an interesting one btw) if successful, instead of an empty body. Problem: we used to say "if the body is not empty, it's not good it's an error"

### What - Proposed solution
As @stephanegigandet suggested, the current "empty" algorithm still works but will soon be deprecated.
The new solution is:
* add `?body=1` to the URL
* check the body, and its possible typical values
  * `{"status":0,"status_verbose":"user not signed-in"}` (wrong user and/or password)
  * `{"status":1,"status_verbose":"user signed-in","user":{"email":"blababla@gmail.com","name":"Mr. John Doe"},"user_id":"gqwbgsvvod"}` (success)
  * this way we can even retrieve the email, the name and the user id (before, we got nothing, and even a possible confusion between email and user id)

monsieurtanuki:
New file:
* `LoginStatus.dart`: Status after an attempt to log in.

Impacted files:
* `api_getProduct_test.dart`: unrelated typo fix
* `openfoodfacts.dart`: new method `login2` that returns a `LoginStatus`
* `user_management_test_prod.dart`: now uses `login2` instead of `login`
* `user_management_test_test_env.dart`: now uses `login2` instead of `login`

### What
- New method `login2` that returns data about user

### Fixes bug(s)
- Closes #579

### DIFF
--- a/lib/model/LoginStatus.dart
+++ b/lib/model/LoginStatus.dart
@@ -42,7 +42,7 @@ class LoginStatus {
       );
 
   /// Was the login successful?
-  bool get ok => status == 1;
+  bool get successful => status == 1;
 
   @override
   String toString() => 'LoginStatus('
--- a/lib/model/LoginStatus.dart
+++ b/lib/model/LoginStatus.dart
@@ -1,0 +1,55 @@
+import 'package:openfoodfacts/interface/JsonObject.dart';
+
+/// Status after an attempt to log in.
+///
+/// Typical JSON:
+/// * wrong user and/or password
+///   {
+///     "status":0,
+///     "status_verbose":"user not signed-in"
+///   }
+/// * successful login:
+///   {
+///     "status":1,
+///     "status_verbose":"user signed-in",
+///     "user_id":"gqwbgsvvod",
+///     "user":{
+///       "email":"blababla@gmail.com",
+///       "name":"Mr. John Doe"
+///     }
+///   }
+class LoginStatus {
+  LoginStatus({
+    required this.status,
+    required this.statusVerbose,
+    this.userEmail,
+    this.userName,
+    this.userId,
+  });
+
+  final int status;
+  final String statusVerbose;
+  final String? userEmail;
+  final String? userName;
+  final String? userId;
+
+  factory LoginStatus.fromJson(Map<String, dynamic> json) => LoginStatus(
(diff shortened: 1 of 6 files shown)


lib/model/LoginStatus.dart
```diff
@@ -0,0 +1,55 @@
...
+
+  final int status;
+  final String statusVerbose;
+  final String? userEmail;
+  final String? userName;
+  final String? userId;
+
+  factory LoginStatus.fromJson(Map<String, dynamic> json) => LoginStatus(
+        status: JsonObject.parseInt(json['status'])!,
+        statusVerbose: json['status_verbose'] as String,
```
M123-dev, line 38: We were talking about not failing if some parameter is missing, we could make this nullable but to be honest it's not that important here

lib/model/LoginStatus.dart
```diff
@@ -0,0 +1,55 @@
...
+        userId: json['user_id'] as String?,
+        userEmail: json['user']['email'] as String?,
+        userName: json['user']['name'] as String?,
+      );
+
+  /// Was the login successful?
+  bool get ok => status == 1;
```
M123-dev, line 45: Ok is si generic

I would prefer loggedIn or correctCredentials or something like that

M123-dev (approved):
Perfect @monsieurtanuki looks good to me, I was thinking about how to name the method while not removing the old one. Login2 is perfectly fine when we switch when all users should have migrated.


Just one suggestion to keep the OpenFoodApiClient (even if just a little shorter) we can make `login` call and return `login2.ok`

monsieurtanuki:
Thank you @M123-dev for your review! I applied some of your suggestions.
I'll merge this PR when the tests are OK, or within 2 hours, whichever comes first ;)

monsieurtanuki:
@M123-dev Feel free to merge 1.26.0.

monsieurtanuki:
@M123-dev But you'd better be ready for Smoothie and remove the possible warnings (e.g. deprecation) the new release creates.

Issue #553, opened 2022-09-12
Check we are not sending credentials in the wrong conditions


teolemon:
### What

- [ ] Check we are not sending credentials in the wrong conditions
  - [ ] is provoking https://github.com/openfoodfacts/smooth-app/issues/2793
  - [ ] is a privacy issue
- [ ] Check we're sending the right kind of authentication (eg Robotoff does not need the regular auth)
  - [ ] Probably provoking openfoodfacts/smooth-app#2954 where Robotoff considers the query a vote, meaning it did not recognize the user

teolemon:
anything that's READ should not have any auth
anything WRITE: user_id + password
Robotoff: 

  static Future<Status> postInsightAnnotation(
    String? insightId,
    InsightAnnotation annotation, {
    User? user,
    String? deviceId,
    bool update = true,
    final QueryType? queryType,
  }) async {
    var insightUri = UriHelper.getRobotoffUri(
      path: 'api/v1/insights/annotate',
      queryType: queryType,
    );

  static Future<SpellingCorrection?> getIngredientSpellingCorrection({
    String? ingredientName,
    Product? product,
    User? user,
    QueryType? queryType,
  }) async {
    Map<String, String> spellingCorrectionParam;

teolemon:
https://github.com/openfoodfacts/robotoff/blob/master/robotoff/app/api.py#L209

raphael0202:
Linked to https://github.com/openfoodfacts/openfoodfacts-dart/issues/560
Robotoff question answers are still sent as vote. It's because we don't seem to send any authentication (https://github.com/openfoodfacts/openfoodfacts-dart/blob/a826add65c6b993f748b2874ef501a5efc900497/lib/openfoodfacts.dart#L618).

Robotoff supports HTTP Basic authentication, so we can authenticate the user this way:
send a `Authorization:'basic {ENCODEDBASE64}' `header where `{ENCODEDBASE64}` is an base64-encoded string of `user:password`. The answer from the server should be (in case of a positive annotation):

```json
{
    "description": "the annotation was saved and sent to OFF",
    "status": "updated",
    "status_code": 2
}
```

monsieurtanuki:
@M123-dev I don't know how exactly they are related but since #590 is merged what is the next step here?

M123-dev:
This should be mostly done, the only open question from #590 was if we should also send credentials to read queries

raphael0202:
[...]

PR #590, opened 2022-10-17
fix: Sending credentials to robotff

M123-dev:
### What
Sending credentials in the correct format to Robotff

See: https://github.com/openfoodfacts/openfoodfacts-dart/issues/553

### DIFF
--- a/lib/utils/HttpHelper.dart
+++ b/lib/utils/HttpHelper.dart
@@ -88,6 +88,11 @@ class HttpHelper {
     User? user, {
     QueryType? queryType,
     required bool addCredentialsToBody,
+
+    /// Typically used for Robotoff,
+    /// See: https://github.com/openfoodfacts/openfoodfacts-dart/issues/553#issuecomment-1269810032
+    /// User or globalUser needed
+    bool addCredentialsToHeader = false,
   }) async {
     if (addCredentialsToBody) {
       if (user != null) {
@@ -98,11 +103,13 @@ class HttpHelper {
     http.Response response = await http.post(
       uri,
       headers: _buildHeaders(
-          user: user,
-          isTestModeActive:
-              OpenFoodAPIConfiguration.getQueryType(queryType) == QueryType.PROD
-                  ? false
-                  : true),
+        user: user,
+        isTestModeActive:
+            OpenFoodAPIConfiguration.getQueryType(queryType) == QueryType.PROD
+                ? false
+                : true,
+        addCredentialsToHeader: addCredentialsToHeader,
+      ),
       body: addUserAgentParameters(body),
     );
     return response;
@@ -172,6 +179,11 @@ class HttpHelper {
   Map<String, String>? _buildHeaders({
     User? user,
     bool isTestModeActive = false,
+
+    /// Typically used for Robotoff,
+    /// See: https://github.com/openfoodfacts/openfoodfacts-dart/issues/553#issuecomment-1269810032
+    /// User or globalUser needed
+    bool addCredentialsToHeader = false,
   }) {
     Map<String, String>? headers = {};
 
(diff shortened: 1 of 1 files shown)


M123-dev:
@raphael0202 one question for that:
I noticed that the auth header for the test server (also base64 encoded `off:off`) and the robotoff auth header are named the same. Is this test authentication needed for the test robotff host (robotoff.openfoodfacts.net)

raphael0202:
No Robotoff preprod API (robotoff.openfoodfacts.net) is available without authentication, unlike Product Opener preprod

lib/utils/HttpHelper.dart
```diff
@@ -88,6 +88,11 @@ class HttpHelper {
     User? user, {
     QueryType? queryType,
     required bool addCredentialsToBody,
+
+    /// Please only use for Robotoff,
```
monsieurtanuki, line 92: I believe you should move this comment upwards, before the method.

lib/utils/HttpHelper.dart
```diff
@@ -172,6 +179,11 @@ class HttpHelper {
   Map<String, String>? _buildHeaders({
     User? user,
     bool isTestModeActive = false,
+
+    /// Please only use for Robotoff,
```
monsieurtanuki, line 183: Please move the comment upwards, before the method.
Actually as you repeating yourself, I would put the bulk of the comment only once (here), and just put a `/// Please only use for Robotoff, see [_buildHeaders] for details`

lib/utils/HttpHelper.dart
```diff
@@ -182,10 +194,17 @@ class HttpHelper {
...
 
-    if (isTestModeActive) {
+    if (isTestModeActive && !addCredentialsToHeader) {
       var token = 'Basic ${base64Encode(utf8.encode('off:off'))}';
       headers.addAll({'authorization': token});
     }
+
+    if (addCredentialsToHeader) {
+      final String? userId = OpenFoodAPIConfiguration.getUser(user)?.userId;
+      final String? password = OpenFoodAPIConfiguration.getUser(user)?.password;
```
monsieurtanuki, line 204:
```suggestion
      final User? myUser = OpenFoodAPIConfiguration.getUser(user);
      final String? userId = myUser?.userId;
      final String? password = myUser?.password;
```

lib/utils/HttpHelper.dart
```diff
@@ -182,10 +194,17 @@ class HttpHelper {
...
+      var token = 'basic ${base64Encode(utf8.encode('$userId:$password'))}';
+      headers.addAll({'Authorization': token});
```
monsieurtanuki, line 206: I don't know who's right but just a dozen line upwards you spell it `Basic` and `authorization`. Please check the correct spellings and be coherent.

monsieurtanuki (commented):
Hi @M123-dev!

There are 4 Robotoff GET methods in off-dart that you don't change because you don't change GET queries, and 2 Robotoff POST methods (`postInsightAnnotation`, `getIngredientSpellingCorrection`) that you don't change because you don't pass the extra parameter...

My suggestions:
* change at least the POST method calls in order to include your extra parameter
* check if it's ok to leave the GET methods as they are

lib/utils/HttpHelper.dart
VaiTon: From https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization

> The [Authentication scheme](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#authentication_schemes) that defines how the credentials are encoded. Some of the more common types are (case-insensitive): [Basic](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#basic_authentication_scheme), Digest, Negotiate and AWS4-HMAC-SHA256.

but I guess we should go with the more used titlecase `Basic`

lib/utils/HttpHelper.dart
```diff
@@ -182,10 +194,17 @@ class HttpHelper {
       'From': OpenFoodAPIConfiguration.getUser(user)?.userId ?? FROM,
     });
 
-    if (isTestModeActive) {
+    if (isTestModeActive && !addCredentialsToHeader) {
       var token = 'Basic ${base64Encode(utf8.encode('off:off'))}';
       headers.addAll({'authorization': token});
```
VaiTon, line 199:
```suggestion
      headers.addAll({'Authorization': token});
```

lib/utils/HttpHelper.dart
VaiTon:
```suggestion
      var token = 'Basic ${base64Encode(utf8.encode('$userId:$password'))}';
      headers.addAll({'Authorization': token});
```

M123-dev:
Thanks for your review @monsieurtanuki @VaiTon 

For the read/get queries it's indeed interesting, @raphael0202 what about adding the credentials there?

(I thought maybe there is some server side logic to not show the same question to a user two times, but then I noticed when one user vote is submitted it's directly applied so no need for that, just for anonym votes)

monsieurtanuki (approved):

PR #591, opened 2022-10-18
style: no_leading_underscores_for_local_identifiers

M123-dev:
### What
Fixes new infos in the dart analyzer: [no_leading_underscores_for_local_identifiers](https://dart-lang.github.io/linter/lints/no_leading_underscores_for_local_identifiers.html)

### DIFF
--- a/test/api_addProductImage_test.dart
+++ b/test/api_addProductImage_test.dart
@@ -28,7 +28,7 @@ void main() {
   /// * index 1: image height in pixels
   /// * index 2: file size in bytes
   /// Inspiration found in https://github.com/CaiJingLong/dart_image_size_getter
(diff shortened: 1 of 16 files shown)


monsieurtanuki (approved):
Hi @M123-dev!
To be honest, I'm not convinced at all by this new check, as `dart` considers that methods embedded inside other methods as variables - and actually you impacted only "private variable methods".
That said, it means that the rest of the code was clean (no illegitimate private _fields_), and it only impacted the test code.
So let's go!

monsieurtanuki:
Hi @M123-dev!
Merging myself as we need this PR for the next PRs to pass.
Still not convinced by the "new" style, but that's another story.

PR #595, opened 2022-10-21
fix: #565 - added product field "nutrition data", better null value management

### LINKED ISSUE #565: Fix nutrient input with 100g/serving
### Description

Create the following product fields:
* "nutrition_data" (either "on" or "")
* "nutrition_data_per" (either "100g" or "serving")

cf. https://github.com/openfoodfacts/smooth-app/issues/3046#issuecomment-1261992853

> * in most cases, there's only one input column (nutrient values), for the "product as sold", with a boolean "per 100g" / "per serving", that will populate directly fields like `energy-kcal_100g` or `energy-kcal_serving`
> * in exceptional cases, there's another input column (nutrient values), for the "product as prepared", with a boolean "per 100g" / "per serving", that will populate directly fields like `energy-kcal_prepared_100g` or `energy-kcal_prepared_serving`
>   => for the record this data is not available yet in openfoodfacts-dart
> * there's another field called `serving_size` - a free text that helps compute a `double` field `serving_quantity`
> * that's the way it looks on the website
>   Capture d’écran 2022-09-29 à 11 05 33
> 
> My additional investigations:
> 
> * if the "product as sold" column is populated, the field "nutrition_data" is "on" (and "" means "not populated")
> * the "nutrition_data_per" field will be populated by either "100g" or "serving", if relevant
> * if the "product as prepared" column is populated, the field "nutrition_data_prepared" is "on" (and "" means "not populated")
> * the "nutrition_data_prepared_per" field will be populated by either "100g" or "serving", if relevant
> * there are additional "imported" fields, like "nutrition_data_per_imported" and "nutrition_data_prepared_per_imported", but that's another story

monsieurtanuki:
Impacted files:
* `JsonHelper.dart`: `checkboxFromJSON` can now return `null`
* `Nutrient.dart`: added static method `fromOffTag`
* `Nutriments.dart`: now we deal with `null` values
* `OffTagged.dart`: added static method `fromOffTag`
* `OrderedNutrient.dart`: new getter `nutrient`
* `PerSize.dart`: added static method `fromOffTag`
* `Product.dart`: minor refactoring
* `ProductFields.dart`: added 'nutrition_data'; refactored "à la 2.17"
* `ProductSearchQueryConfiguration.dart`: minor refactoring

### Note
A quick review would be appreciated, as there's a pending bug on Smoothie with nutrients in mg and microg that I'm about to solve, which requires this PR.

### What
There were tiny (but important) fixes needed around `Nutriments` recent changes.

We needed a new product field ("nutrition data") - without it we cannot get the value from the server

While testing, I noticed that we needed to improve the management of `null` values.
For instance, it was not possible to erase a value, as either you put a `null` that it interpreted by the server as `0`, or you don't put a value which is interpreted by the server as "this value doesn't change".
The solution was a better `null` management here in off-dart, and setting a `''` to explicitly tell the server to erase the value.

### Fixes bug(s)
- Closes #565

### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/3046


VaiTon (approved):

monsieurtanuki:
Thank you @VaiTon for your review!

PR #604, opened 2022-11-06
fix: Added response for incorrect email

Akashsri3bi:
### What
- SignupStatusError Response gave long html response which was actually use of incorrect email address which was not added as a response .


monsieurtanuki (approved):
Thank you @Akashsri3bi, looks good to me!

Akashsri3bi:
> Thank you @Akashsri3bi, looks good to me!

Thanks for suggestions @monsieurtanuki !

PR #624, opened 2022-11-19
feat: Add copyWith method to all models

M123-dev:
### What
Adds copyWith methods to all JsonSerializable models we have

My first try was to use [freezed](https://pub.dev/packages/freezed) as this can add a copyWith method to JsonSerializable but because we have a lot of customization around languages and so on I decided to go for this simpler method.

For most classes it was just adding a `@CopyWith()` decorator others needed a change in the constructor to use named parameter.

I added it to all classes in a first run, if there are some were we don't need it I'am happy to remove them

### For
https://github.com/openfoodfacts/smooth-app/pull/3059
https://github.com/openfoodfacts/smooth-app/issues/3263
...

### DIFF
--- a/lib/model/TaxonomyAdditive.dart
+++ b/lib/model/TaxonomyAdditive.dart
@@ -1,10 +1,11 @@
+import 'package:copy_with_extension/copy_with_extension.dart';
 import 'package:json_annotation/json_annotation.dart';
 import 'package:openfoodfacts/interface/JsonObject.dart';
 import 'package:openfoodfacts/model/OffTagged.dart';
 import 'package:openfoodfacts/openfoodfacts.dart';
 import 'package:openfoodfacts/utils/CountryHelper.dart';
-import 'package:openfoodfacts/utils/TaxonomyQueryConfiguration.dart';
 import 'package:openfoodfacts/utils/TagType.dart';
+import 'package:openfoodfacts/utils/TaxonomyQueryConfiguration.dart';
 
 part 'TaxonomyAdditive.g.dart';
 
@@ -62,9 +63,10 @@ enum TaxonomyAdditiveField implements OffTagged {
 ///
 /// See [OpenFoodAPIClient.getTaxonomy] for more details on how to retrieve one
 /// of these.
+@CopyWith()
 @JsonSerializable()
 class TaxonomyAdditive extends JsonObject {
-  TaxonomyAdditive(
+  TaxonomyAdditive({
     this.additivesClasses,
     this.carbonFootprintFrFoodgesIngredient,
     this.carbonFootprintFrFoodgesValue,
@@ -94,7 +96,7 @@ class TaxonomyAdditive extends JsonObject {
     this.vegan,
     this.vegetarian,
     this.wikidata,
-  );
+  });
 
   factory TaxonomyAdditive.fromJson(Map<String, dynamic> json) {
     return _$TaxonomyAdditiveFromJson(json);
--- a/lib/model/TaxonomyAllergen.dart
+++ b/lib/model/TaxonomyAllergen.dart
@@ -1,10 +1,11 @@
+import 'package:copy_with_extension/copy_with_extension.dart';
 import 'package:json_annotation/json_annotation.dart';
 import 'package:openfoodfacts/interface/JsonObject.dart';
 import 'package:openfoodfacts/model/OffTagged.dart';
 import 'package:openfoodfacts/openfoodfacts.dart';
 import 'package:openfoodfacts/utils/CountryHelper.dart';
-import 'package:openfoodfacts/utils/TaxonomyQueryConfiguration.dart';
 import 'package:openfoodfacts/utils/TagType.dart';
+import 'package:openfoodfacts/utils/TaxonomyQueryConfiguration.dart';
 
 part 'TaxonomyAllergen.g.dart';
 
@@ -31,13 +32,14 @@ enum TaxonomyAllergenField implements OffTagged {
 ///
 /// See [OpenFoodAPIClient.getTaxonomy] for more details on how to retrieve one
 /// of these.
+@CopyWith()
 @JsonSerializable()
 class TaxonomyAllergen extends JsonObject {
-  TaxonomyAllergen(
+  TaxonomyAllergen({
     this.name,
     this.synonyms,
     this.wikidata,
-  );
+  });
 
   factory TaxonomyAllergen.fromJson(Map<String, dynamic> json) {
     return _$TaxonomyAllergenFromJson(json);
--- a/lib/model/TaxonomyCategory.dart
+++ b/lib/model/TaxonomyCategory.dart
@@ -1,10 +1,11 @@
+import 'package:copy_with_extension/copy_with_extension.dart';
 import 'package:json_annotation/json_annotation.dart';
 import 'package:openfoodfacts/interface/JsonObject.dart';
 import 'package:openfoodfacts/model/OffTagged.dart';
 import 'package:openfoodfacts/openfoodfacts.dart';
 import 'package:openfoodfacts/utils/CountryHelper.dart';
-import 'package:openfoodfacts/utils/TaxonomyQueryConfiguration.dart';
 import 'package:openfoodfacts/utils/TagType.dart';
+import 'package:openfoodfacts/utils/TaxonomyQueryConfiguration.dart';
 
 part 'TaxonomyCategory.g.dart';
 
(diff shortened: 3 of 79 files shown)


lib/model/TaxonomyAdditive.dart
```diff
@@ -62,9 +63,10 @@ enum TaxonomyAdditiveField implements OffTagged {
 ///
 /// See [OpenFoodAPIClient.getTaxonomy] for more details on how to retrieve one
 /// of these.
+@CopyWith()
 @JsonSerializable()
 class TaxonomyAdditive extends JsonObject {
-  TaxonomyAdditive(
+  TaxonomyAdditive({
```
monsieurtanuki, line 69: Breaking change.

lib/model/TaxonomyAllergen.dart
```diff
@@ -31,13 +32,14 @@ enum TaxonomyAllergenField implements OffTagged {
 ///
 /// See [OpenFoodAPIClient.getTaxonomy] for more details on how to retrieve one
 /// of these.
+@CopyWith()
 @JsonSerializable()
 class TaxonomyAllergen extends JsonObject {
-  TaxonomyAllergen(
+  TaxonomyAllergen({
```
monsieurtanuki, line 38: Breaking change.

lib/model/TaxonomyCategory.dart
```diff
@@ -58,9 +59,10 @@ enum TaxonomyCategoryField implements OffTagged {
 ///
 /// See [OpenFoodAPIClient.getTaxonomy] for more details on how to retrieve one
 /// of these.
+@CopyWith()
 @JsonSerializable()
 class TaxonomyCategory extends JsonObject {
-  TaxonomyCategory(
+  TaxonomyCategory({
```
monsieurtanuki, line 65: Breaking change.

lib/model/TaxonomyCountry.dart
```diff
@@ -33,9 +34,18 @@ enum TaxonomyCountryField implements OffTagged {
 ///
 /// See [OpenFoodAPIClient.getTaxonomy] for more details on how to retrieve one
 /// of these.
+@CopyWith()
 @JsonSerializable()
 class TaxonomyCountry extends JsonObject {
-  TaxonomyCountry();
+  TaxonomyCountry({
```
monsieurtanuki, line 40: Did `copy_with` force you to do that?

lib/model/TaxonomyIngredient.dart
```diff
@@ -78,9 +79,10 @@ enum TaxonomyIngredientField implements OffTagged {
 ///
 /// See [OpenFoodAPIClient.getTaxonomy] for more details on how to retrieve one
 /// of these.
+@CopyWith()
 @JsonSerializable()
 class TaxonomyIngredient extends JsonObject {
-  TaxonomyIngredient(
+  TaxonomyIngredient({
```
monsieurtanuki, line 85: Breaking change.

lib/model/TaxonomyLabel.dart
```diff
@@ -52,9 +53,10 @@ enum TaxonomyLabelField implements OffTagged {
 ///
 /// See [OpenFoodAPIClient.getTaxonomy] for more details on how to retrieve one
 /// of these.
+@CopyWith()
 @JsonSerializable()
 class TaxonomyLabel extends JsonObject {
-  TaxonomyLabel(
+  TaxonomyLabel({
```
monsieurtanuki, line 59: Breaking change.

lib/model/TaxonomyLanguage.dart
```diff
@@ -32,14 +33,15 @@ enum TaxonomyLanguageField implements OffTagged {
 ///
 /// See [OpenFoodAPIClient.getTaxonomy] for more details on how to retrieve one
 /// of these.
+@CopyWith()
 @JsonSerializable()
 class TaxonomyLanguage extends JsonObject {
-  TaxonomyLanguage(
+  TaxonomyLanguage({
```
monsieurtanuki, line 39: Breaking change.

lib/model/TaxonomyOrigin.dart
```diff
@@ -26,9 +27,15 @@ enum TaxonomyOriginField implements OffTagged {
 ///
 /// See [OpenFoodAPIClient.getTaxonomy] for more details on how to retrieve one
 /// of these.
+@CopyWith()
 @JsonSerializable()
 class TaxonomyOrigin extends JsonObject {
-  TaxonomyOrigin();
+  TaxonomyOrigin({
```
monsieurtanuki, line 33: Did `copy_with` force you to do that?

lib/model/TaxonomyPackaging.dart
```diff
@@ -41,9 +42,16 @@ enum TaxonomyPackagingField implements OffTagged {
 ///
 /// See [OpenFoodAPIClient.getTaxonomy] for more details on how to retrieve one
 /// of these.
+@CopyWith()
 @JsonSerializable()
 class TaxonomyPackaging extends JsonObject {
-  TaxonomyPackaging();
+  TaxonomyPackaging({
```
monsieurtanuki, line 48: idem

lib/model/TaxonomyPackagingMaterial.dart
```diff
@@ -26,9 +27,15 @@ enum TaxonomyPackagingMaterialField implements OffTagged {
 ///
 /// See [OpenFoodAPIClient.getTaxonomy] for more details on how to retrieve one
 /// of these.
+@CopyWith()
 @JsonSerializable()
 class TaxonomyPackagingMaterial extends JsonObject {
-  TaxonomyPackagingMaterial();
+  TaxonomyPackagingMaterial({
```
monsieurtanuki, line 33: idem

lib/model/TaxonomyPackagingRecycling.dart
```diff
@@ -28,9 +29,17 @@ enum TaxonomyPackagingRecyclingField implements OffTagged {
 ///
 /// See [OpenFoodAPIClient.getTaxonomy] for more details on how to retrieve one
 /// of these.
+@CopyWith()
 @JsonSerializable()
 class TaxonomyPackagingRecycling extends JsonObject {
-  TaxonomyPackagingRecycling();
+  TaxonomyPackagingRecycling({
```
monsieurtanuki, line 35: idem

lib/model/TaxonomyPackagingShape.dart
```diff
@@ -26,9 +27,15 @@ enum TaxonomyPackagingShapeField implements OffTagged {
 ///
 /// See [OpenFoodAPIClient.getTaxonomy] for more details on how to retrieve one
 /// of these.
+@CopyWith()
 @JsonSerializable()
 class TaxonomyPackagingShape extends JsonObject {
-  TaxonomyPackagingShape();
+  TaxonomyPackagingShape({
```
monsieurtanuki, line 33: idem

monsieurtanuki (commented):
Hi @M123-dev!
A lot to review, but each time it's rather quick!

TBH, I cannot picture exactly the added value of that "CopyWith" feature, more specifically in the context of https://github.com/openfoodfacts/smooth-app/pull/3059 and https://github.com/openfoodfacts/smooth-app/issues/3263.

In those Smoothie issues, we need to "merge" two `Product`s (the second on top on the first, only for the second's non null fields), like `inputProduct.merge(product: changes)`, or `Product.merge(product1, product2)`.
And from what I saw in the generated code it's just about individual fields. Like `product1.copyWith(lang: 'de')`.
If so, that would mean that we would need to list all the fields of the second product and then use `copyWith` just on it, which is more a pain in the neck than a good feature.

And in dart, you can already write easily `Product(barcode: '1234')..lang='de'..website='https://...'`.

Therefore, I think we need at least a relevant example of the use of `copyWith`. Would you please share one?

Beyond that, you introduced slight breaking changes, but as it's for classes that are not very interesting I'm sure you'd agree to rollback your changes there, as you suggested in the OP.

M123-dev:
Ohh, my bad @monsieurtanuki it makes sense that these methods do not make sense (at least for us), I was a bit hasty with the implementation.

monsieurtanuki:
Thank you anyway for your attempt @M123-dev!

M123-dev:
So looking at the problem how we actually have it. 
- We need a way to combine two objects
- Without specifying every property
- Preferably without having to write all properties
- My idea for this is to convert both passed objects to JSON and then into to a map and then just calling 
Map.addAll(), this doesn't work for nested properties though. Do you have another idea?

monsieurtanuki:
@M123-dev Actually I think I had a specific need like that in https://github.com/openfoodfacts/smooth-app/pull/3339 (I created a `TODO` about it): a clean copy of a `Product`(like you suggested, a sort of `fromJson(toJson))`.

Should we generalize that? I don't know. For the moment we only have this specific need for `Product`.

Another grander idea would be to code a bit like in `Nutriments`: just keep the json map and create getter/setter for all "json fields".
But it would be a bit out of proportions: refactoring all our classes because for one single need in Smoothie we may find interesting to use a `copyWith`/`merge` method on some fields of `Product`...

If you feel like refactoring in general, I would suggest #450. Would impact _all_ files, with breaking changes. But the code would be cleaner afterwards.

PR #626, opened 2022-11-22
fix: #625 - no user credentials in header if null

### LINKED ISSUE #625: Invalid authentication sent to Robotoff during question answering
See https://sentry.io/organizations/openfoodfacts/issues/3749073432/events/2638e00077e54c64abf983aaedb71838/?project=1415205&referrer=previous-event

'null' is sometimes sent as username in during `/insights/annotate` call.

monsieurtanuki:
Impacted file:
* `HttpHelper.dart`

### What
- The user credentials are added in the request header only for Robotoff, for `api/v1/insights/annotate` and `api/v1/predict/ingredients/spellcheck`
- In the special case when the user is `null`, we should NOT try to populate the request header with it - that's what this PR fixes

### Fixes bug(s)
- Closes #625

### DIFF
--- a/lib/utils/HttpHelper.dart
+++ b/lib/utils/HttpHelper.dart
@@ -174,9 +174,9 @@ class HttpHelper {
     return status;
   }
 
-  /// build the request headers
-  /// By default isTestMode is false
-  /// Note: [addCredentialsToHeader] and [isTestModeActive] exclude each other
+  /// Returns the request headers.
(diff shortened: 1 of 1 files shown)


M123-dev (approved):
LGTM, thanks @monsieurtanuki

monsieurtanuki:
Thank you @M123-dev!

PR #628, opened 2022-11-24
chore: move tests from Smoothie

monsieurtanuki:
New file:
* `registration_login_test.dart`

### What
- There were tests in Smoothie that should be here in off-dart.
- They have a tendency to fail (relying on unstable TEST env), so they failed in Smoothie.
- I moved here those tests - they still fail when they have to, but at least it's not connected to Smoothie.

### DIFF
--- a/test/registration_login_test.dart
+++ b/test/registration_login_test.dart
@@ -1,0 +1,66 @@
+import 'dart:math';
+
+import 'package:openfoodfacts/model/LoginStatus.dart';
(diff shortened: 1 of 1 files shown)


M123-dev:
Don't we have these tests already in
```
test/user_management_test_prod.dart
user_management_test_test_env.dart
```

monsieurtanuki:
That's correct, I've just double-checked and all tests added in this PR were redundant with either the TEST or the PROD tests.

Issue #575, opened 2022-10-08
❌ test/api_getRobotoff_test.dart: OpenFoodAPIClient get robotoff insights get random insight (failed)


teolemon:
### Description
❌ test/api_getRobotoff_test.dart: OpenFoodAPIClient get robotoff insights get random insight (failed)
  Expected: not null
    Actual: <null>
  
  package:test_api                      expect
  test/api_getRobotoff_test.dart 111:7  main.<fn>.<fn>

monsieurtanuki:
Fixed in #570.

PR #639, opened 2022-12-09
fix: 575 - less expected populated fields from Robotoff

monsieurtanuki:
Impacted file:
* `api_getRobotoff_test.dart`

### What
- For historical reasons we were expecting many populated fields from Robotoff, and in some cases - especially with the "give me some random data" test - some fields are _not_ populated.
- Those "unsafe" fields are now removed from the tests.

### Fixes bug(s)
- Fixes: #575

### DIFF
--- a/test/api_getRobotoff_test.dart
+++ b/test/api_getRobotoff_test.dart
@@ -93,7 +93,7 @@ void main() {
   });
 
   group('$OpenFoodAPIClient get robotoff insights', () {
-    test('FABRICE get random insight', () async {
+    test('get random insight', () async {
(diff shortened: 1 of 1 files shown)


test/api_getRobotoff_test.dart
```diff
@@ -93,7 +93,7 @@ void main() {
   });
 
   group('$OpenFoodAPIClient get robotoff insights', () {
-    test('get random insight', () async {
+    test('FABRICE get random insight', () async {
```
M123-dev, line 96:
```suggestion
    test('get random insight', () async {
```

monsieurtanuki:

Issue #634, opened 2022-12-02
energy / energyKj confusion fix


monsieurtanuki:
### Description
cf. https://github.com/openfoodfacts/smooth-app/pull/3399
* Depending on the country, the order and the list of nutrients vary.
* There is a bit of confusion as for historical reasons, we have both energy (in KJ) and energyKJ (erm... in KJ) in off, but in off-dart we only have energyKJ
* At least for Canada, there's no energyKJ, only energy, that should match with energyKJ but does not.

### Expected behavior
Probably, the best solution would be to map tag `'energy'` to Nutrient `energyKj`, in `Nutrient.fromOffTag`.
Another solution would be to create an additional `energy` nutrient (in addition to `energyKj` and `energyKCal`), but that's a bit confusing.

### Additional context
Nutrients currently supported by countries: https://github.com/openfoodfacts/openfoodfacts-server/blob/main/lib/ProductOpener/Food.pm#L522-L834

PR #644, opened 2022-12-17
fix: 634 - mapped tag 'energy' to Nutrient energyKj.

monsieurtanuki:
Impacted file:
* `Nutrient.dart`: now manages the special duplicate case `'energy'` / `'energy-kj'`

### What
- In recent refactoring we got rid of the duplicate case `'energy'` / `'energy-kj'` (both in kj)
- This PR fixes a forgotten case: ok we use only one `Nutrient` called `energyKj`, that not only matches with offTag `'energy-kj'` but also with `'energy'`

### Fixes bug(s)
- Fixes: #634


M123-dev (approved):

monsieurtanuki:
Thank you @M123-dev!

PR #646, opened 2022-12-19
feat: 539 - specific test for nova group filter

monsieurtanuki:
Impacted file:
* `api_searchProducts_test.dart`: added a specific test for nova group filter; switched tests to PROD env whenever possible

### What
- Just added a test on nova group filter - the filtering was already there (but was not really tested).

### Fixes bug(s)
- Closes: #539

### DIFF
--- a/test/api_searchProducts_test.dart
+++ b/test/api_searchProducts_test.dart
@@ -21,7 +21,8 @@ import 'package:test/test.dart';
 import 'test_constants.dart';
 
 void main() {
-  OpenFoodAPIConfiguration.globalQueryType = QueryType.TEST;
+  OpenFoodAPIConfiguration.globalQueryType = QueryType.PROD;
+  OpenFoodAPIConfiguration.globalUser = TestConstants.PROD_USER;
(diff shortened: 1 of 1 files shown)


monsieurtanuki:
> Why the switch to the prod environment though, isn't the test env exactly for this?

That's an interesting question.

My point of view: in off-dart we're not here to test the server's own TEST env, we're here to test if what we coded works in the normal conditions (PROD), as _integration_ or _functional_ tests.
We are virtually forced to run some tests "not on PROD" though, when we save data.
Besides, the TEST env is rather unstable (not a big deal, it's a TEST env), and I prefer to crash the tests with our own bugs with PROD data than because of a test platform instability.

We could also:
1. test whatever changed in server TEST that is not yet in server PROD (that would mean calling also in TEST our current tests in PROD), a bit like _non regression_ tests
2. completely ignore what the server says and use fake http calls - that's what some tests were about (and I got rid of them because they were actually testing `jsonDecode`), more like _unit_ tests.

PR #647, opened 2022-12-21
fix: little noNutritionData addition

M123-dev:
This was the my first guess while looking for an issue in smooth_app (https://github.com/openfoodfacts/smooth-app/issues/3445) There I stumbled across this. Just a change in return value

### DIFF
--- a/lib/model/Product.dart
+++ b/lib/model/Product.dart
@@ -748,6 +748,8 @@ class Product extends JsonObject {
       return _noNutritionData!;
     } else if (_nutriments != null) {
       return false;
+    } else if (_nutriments == null) {
+      return true;
     } else {
       return null;
     }
--- a/lib/model/Product.dart
+++ b/lib/model/Product.dart
@@ -743,16 +743,11 @@ class Product extends JsonObject {
     toJson: JsonHelper.checkboxToJSON,
     fromJson: JsonHelper.checkboxFromJSON,
   )
-  bool? get noNutritionData {
+  bool get noNutritionData {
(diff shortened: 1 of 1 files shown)


lib/model/Product.dart
```diff
@@ -748,6 +748,8 @@ class Product extends JsonObject {
       return _noNutritionData!;
     } else if (_nutriments != null) {
       return false;
+    } else if (_nutriments == null) {
```
monsieurtanuki, line 751: Then the last `else` is useless.
Rather return directly `_nutriments == null`.

monsieurtanuki (commented):
Hi @M123-dev!
Could you please provide some context for this PR?
Besides, please take into consideration the fact that there are probably explicit fields - at least on the server side - that have similar goals, like "We know for a fact there are no nutrition facts written on the product packaging".

M123-dev:
For context, this was the my first guess while looking for https://github.com/openfoodfacts/smooth-app/issues/3445 where I stumbled over it

monsieurtanuki:
> For context, this was the my first guess while looking for [openfoodfacts/smooth-app#3445](https://github.com/openfoodfacts/smooth-app/issues/3445) where I stumbled over it

OK. Please add it to the PR description.

From what we discussed in https://github.com/openfoodfacts/smooth-app/issues/3445, it looks like it's not enough: who would care about an empty (and not null) nutriment list?

I guess we need some more thinking about what we want (e.g. "I have at least one nutrient data"), which fields the server provides, and what we currently have in off-dart.

PR #653, opened 2022-12-25
feat!: 450 - Name source files using "lowercase_with_underscores"

monsieurtanuki:
Impacted files:
* `analysis_options.yaml`: remove the line that disabled the filename check
* `fake_http_helper.dart`: unrelated removal "en passant"
* `pubspec.yaml`: upgrade of `lints` in order to be as close as possible to `pub.dev` parameters
* 181 renamed files
* 22 files importing renamed files

### What
- BREAKING CHANGE!
- This is only a part of #450, which will be implement PR by PR.
- This first PR is "only" about renaming the files in order to match the "lowercase_with_underscores" spec
- Next PRs will be about `export`ing all those files, so that `import`ing only `openfoodfacts/openfoodfacts.dart` would be enough.
- In the end, the impact on all projects that use off-dart will be that: replacing all their individual file imports with only `openfoodfacts/openfoodfacts.dart`
- One goal is to get the maximum score on pub.dev.
- A side effect is that we'll then have full flexibility for refactoring files into functional folders, as it will have no impact on the package users.

### Part of 
- #450


M123-dev (approved):
Looks great @monsieurtanuki 👍🏼 

As you said, together with the exports we'll have a lot more freedom on what to do

monsieurtanuki:
Now I can work on the next steps.

PR #655, opened 2022-12-26
feat!: 450 - import only "openfoodfacts.dart"; all code in "src" folder

monsieurtanuki:
New file:
* `open_food_api_client.dart`: contains the _code_ previously stored in `openfoodfacts.dart`

Impacted files:
* `openfoodfacts.dart`: now only `export`ing all relevant files
* test files: now they import only `openfoodfacts/openfoodfacts.dart`
* `product_state.dart`: renamed from `state.dart` to `ProductState`, as `State` is used for `StatefulWidget`s and that would make upgrade painful
* `nutrient_levels.dart`: renamed `Level` as `NutrientLevel` to fix confusion with `Level` in `knowledge_panel.dart`
* all code files: moved to `src` folder; now `import`ing relative files instead of openfoodfacts package files

### What
- With this PR all developers using off-dart will only have to import `openfoodfacts/openfoodfacts.dart` (and remove the other `openfoodfacts/*` imports)
- Upgrade is smooth but repetitive - for the record in Smoothie that would mean 98 impacted files

### Part of 
- #450


M123-dev (approved):
Looks great, the renaming makes sense, thanks @monsieurtanuki

monsieurtanuki:
I think I'm done with #450.
While we're here, are there other small breaking changes we could include here?
Perhaps add some consistency with `User? user` parameters in the `open_food_api_client.dart` methods?

Issue #514, opened 2022-07-12
Use shared Lint rules


g123k:
We have Flutter rules for our projects.
It would be nice to also enforce rules for Dart project, including this one

Issue #339, opened 2021-12-20
Further splitting up of openfoodfacts.dart


M123-dev:
As said in #158 `openfoodfacts.dart` has gotten really big overtime there therefore we decided (for the time beeing) to code the folksonomy engine API into a separate file.

 I would suggest that we take this even further. Namely that we move the Robotoff methods to a separate file and separate class like `RobotoffApiClient` or so and depending on what the opinions are here do it like most dart packages I have seen and use the main file only for exports. This would mean that we move the OpenFoodAPIClient class to its own file. This should not even break anything since the package user only imports the main file with all the exports.

@monsieurtanuki @teolemon @peterwvj what are your opinions on that?

peterwvj:
I fully support this. Even if it requires breaking changes.

monsieurtanuki:
Currently `openfoodfacts.dart` weighs 1092 lines of code.

Inside this file, there are 6 robotoff methods (200 lines of code).  That means, still 900 lines of code in `openfoodfacts.dart` if we move away robotoff.

I don't know the exact relationships between OFF, robotoff and folksonomy, and if I decided to create a separate file for folksonomy it was because it seemed to me like a satellite project with different users (maybe I'm wrong). Then that made sense to create a distinct file in the same library. A different library would have been OK too.
But if for the library user we're playing - for both OFF and robotoff - with the same data and the same OFF users, I'm not sure it's a good idea to split OFF and robotoff.

I mean, regardless of technical questions on our side, what would make sense for the library user? OFF, robotoff and folksonomy in the same class/file? Or would that bring confusion to mix unrelated methods? OFF, robotoff and folksonomy in different classes/files? Or would that add complexity to library users: after all, should they care - when they code in dart - if for historical reasons the subdomain is different on the server side?
That was for the library user side.

Regarding "our" technical side, I'm OK with @M123-dev's suggestions. I'm a bit reluctant with a breaking change, but that would be a good opportunity to test the `feat!:` keyword and to switch to version `2.0.0`.

M123-dev:
I would even wait a little longer with the breaking changes. That is, until we have finished most of the restructuring, which hopefully won't take that long. Then we can remove all other pending deprecated fields directly with it. 

I understand your point @monsieurtanuki but somehow Robotoff is still a bit different even though it uses the same users. All the methods still have Robotoff in their name.  It's probably not helpfull that robotoff is found by sdk users but classes are not designed for that anyway.

[...]

PR #656, opened 2022-12-26
refactor!: 450 - finalizing the breaking changes

monsieurtanuki:
Deleted files:
* `page.dart`: deprecated from 2022-04-20
* `product_list_query_configuration.dart`: deprecated from 2022-08-03
* `to_be_completed_configuration.dart`: deprecated from 2022-08-03

Impacted files:
* `image_helper.dart`: removed code deprecated from 2022-08-18, and so we can get rid of package `image`
* `insight.g.dart`: generated
* `nutriments.dart`: removed 1655 (!) lines of code deprecated from 2022-10-04 - it was not urgent but now seemed a good opportunity
* `nutriments_helper.dart`: removed code deprecated from 2022-10-04 - it was not urgent but now seemed a good opportunity
* `open_food_api_client.dart`: removed code deprecated from 2022-01-22 and 2022-07-26
* `openfoodfacts.dart`: removed export deprecated from 2022-04-20
* `product.dart`: fixed typo
* `product.g.dart`: generated
* `pubspec.yaml`: removed deprecated dependency on `image`; upgraded versions
* `README.md`: added info about how to deal with 2.x.x breaking changes

### What
- Some minor cleaning after the most important PRs are merged.

### Fixes bug(s)
- Fixes: #450


M123-dev (approved):
Looks great @monsieurtanuki

monsieurtanuki:
Thank you @M123-dev: now I think we're ready for 2.0.0!
No rush though, as I've noticed that Smoothie only used 1.27.1. I'm going to upgrade Smoothie to off-dart 1.30.1 first.

M123-dev:
Yeah, if I have time later I'm going to work on https://github.com/openfoodfacts/openfoodfacts-dart/issues/339 and maybe https://github.com/openfoodfacts/openfoodfacts-dart/issues/514

monsieurtanuki:
Regarding #514: we cannot apply the same rules for off-dart and for Smoothie, for at least one good reason - for libraries like off-dart inner `import`s should use the relative syntax, and in the current Smoothie rules we are supposed to use the package syntax. Maybe "dart library" specific rules? But basically that would mean whatever pub.dev requires, perhaps a bit stronger.

Regarding #339, last time we discussed it was a year ago. Please check if that would still make sense, and state what you plan to code (would be "create a separate class for all the Robotoff methods", wouldn't it?).

M123-dev:
Your worries for #514 make sense, that's why it was more of an optional thing, I have to check if all the rules make sense and probably override some of them.

monsieurtanuki:
@M123-dev It was not a worry, it was a fact: I tried and saw tons of warnings!

I believe we don't have to wait for you to implement #514 for version 2.0.0, because:
* either the impact will be HUGE (e.g. all constants in camelCase) and that wouldn't be a good idea at all (too much unpleasant work for developers that use off-dart)
* or there will be no impact at all, breaking change wise (e.g. specifying the type before each empty `[]`)

M123-dev:
Okay makes sense, if there is no urge I would start writing some documentation today and tomorrow
Also thinking about the two different classes for robot off, I don't think I make a big difference for us, it just hides it more from the package users. If it's in the same class, they might stumble upon it, so I have crossed it off my list

Issue #650, opened 2022-12-25
Unstable TEST env causes pub.dev: "Run sdk tests: failing"


monsieurtanuki:
### Description
`Run sdk tests: failing` on https://pub.dev/packages/openfoodfacts:
![Capture d’écran 2022-12-25 à 11 36 01](https://user-images.githubusercontent.com/11576431/209464651-c4f654e4-0a9c-404c-a394-6d46a98919cf.png)

Going further: https://github.com/openfoodfacts/openfoodfacts-dart/actions/workflows/test-sdk.yml

Going further: https://github.com/openfoodfacts/openfoodfacts-dart/actions/runs/3773476930/jobs/6414944936
Just one test failed:
![Capture d’écran 2022-12-25 à 11 43 21](https://user-images.githubusercontent.com/11576431/209464878-38966b56-24d8-4e0d-bb44-70393dbe2667.png)

In short, the only tests that pass in `api_saveProducts_test.dart` are tests that are either skipped or do not depend on an actual "save to server" action.

### Expected behavior
We should be able to rely on an env where we can save data, at least for tests on a minimal list of products.

teolemon:
I added that to the Infrastructure project, cc @alexgarel . Meanwhile, would a barcode change help ?

monsieurtanuki:
It looks like we're talking mainly about time-outs, so I don't believe a barcode change would help.
Perhaps on DEV? (potentially smaller, therefore faster).
Again, for off-dart it's only a matter of a dozen of tests.

teolemon:
❌ test/api_search_products_test.dart: OpenFoodAPIClient search products nova filter (failed)
  FormatException: Unexpected character (at character 1)
  <h1>Software error:</h1>
  ^
  
  dart:convert                                                JsonCodec.decode
  package:openfoodfacts/src/open_food_api_client.dart 344:60  OpenFoodAPIClient.searchProducts
  ===== asynchronous gap ===========================
  test/api_search_products_test.dart 797:37                   main.<fn>.<fn>

monsieurtanuki:
If I remember well that one is about a search on NOVA, that sometimes ends with a Mongo DB time-out.
We may make the search faster with additional search parameter, like a less data-crowded country. Would Mongo filter first on the country then on nova or the other way around? Is the problem about sorting the too numerous results?

PR #668, opened 2022-12-31
fix: 650 - less server demanding nova integration test

monsieurtanuki:
Impacted file:
* `api_search_products_test.dart`: slightly edited a test in order to be less intense for the server

### What
- Filter on NOVA looks very demanding on the server resources, and for NOVA 4 it may lead to Mongo DB crash.
- The fix here is about switching to a country with less data than France - in this case, Poland.

### Fixes bug(s)
- Fixes: #650

### DIFF
--- a/test/api_search_products_test.dart
+++ b/test/api_search_products_test.dart
@@ -784,12 +784,14 @@ void main() {
     });
 
     test('nova filter', () async {
-      // approximated min counts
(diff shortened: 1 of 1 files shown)


teolemon (approved):
fyi @alexgarel @stephanegigandet

monsieurtanuki:

PR #669, opened 2023-01-01
fix: #670 : Adding packagings to Product object

MohamedFBoussaid:


monsieurtanuki:
Happy new year @MohamedFBoussaid!

While I admit that deprecating `packaging` was perhaps a bit too fast a decision, your PR does not really help.
If you want to set `packagings`, you just have to erm... set packagings, like `product.packagings = ...` or `Product()..packagings = ...`.
I think that systematically adding parameters to the constructor is not really a good idea, as it repeats code with 0 added value.
Or maybe there's something that I've missed?

monsieurtanuki:
> If that is the case, I will delete my PR, but maybe the Decrepit message need to be updated, as it is miss leading.
Bcs on my code I was using this :
 `Product product = Product(...., packaging:  _productPackigingController.text, ....);`

> And I understood (from 'Use packagingS instead' that the packgaingS will have the same logic

I understand your questioning, but anyway it was not possible to just replace `packaging` with `packagings`, as for the moment we cannot save `packagings` with the standard `saveProduct` method (only with the new temporary V3 save product method), and we cannot either save the whole `Product` with the new method.
If you have a better message for developers about how to save `packagings`, don't hesitate to PR it.

PR #671, opened 2023-01-02
refactor: lint - as close to pub.dev as possible in order to pass the checks

monsieurtanuki:
Impacted files:
* `analysis_options.yaml`: as close as the recommendations as possible
* `barcodes_validator.dart`: minor refactoring
* `product.g.dart`: ugly minor refactoring
* `product_result.g.dart`: ugly minor refactoring
* `robotoff_question.dart`: minor refactoring
* `status.dart`: minor refactoring
* `user.dart`: minor refactoring
* `user_agent.dart`: minor refactoring

### What
- Unexpectedly with 2.0.0 we still haven't reached the max score of 140 on pub.dev.
- Part of the reason is that pub.dev does check the generated files, and we did not.
- Now our check parameters are as close to pub.dev as possible.
- Hopefully when this PR is merged into pub.dev we get the max score!

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,16 +1,5 @@
 include: package:lints/recommended.yaml
 
-analyzer:
-  exclude:
-    - "**/*.g.dart"
-  errors:
-    omit_local_variable_types: ignore
-    annotate_overrides: ignore
-    use_function_type_syntax_for_parameters: ignore
-    missing_required_param: warning
-    missing_return: warning
-    prefer_single_quotes: warning
-
 linter:
   rules:
     constant_identifier_names: false
--- a/lib/src/model/product.g.dart
+++ b/lib/src/model/product.g.dart
@@ -1,3 +1,6 @@
+// ignore_for_file: deprecated_member_use_from_same_package
+// That's a bit ugly, but we need the previous line in order to pass the pub.dev
+// tests, so put that line back after code generation.
 // GENERATED CODE - DO NOT MODIFY BY HAND
 
 part of 'product.dart';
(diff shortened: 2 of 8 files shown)


analysis_options.yaml
```diff
@@ -1,16 +1,5 @@
...
-analyzer:
-  exclude:
-    - "**/*.g.dart"
-  errors:
-    omit_local_variable_types: ignore
-    annotate_overrides: ignore
-    use_function_type_syntax_for_parameters: ignore
-    missing_required_param: warning
-    missing_return: warning
-    prefer_single_quotes: warning
```
M123-dev, line 12: I would definitely leave these 3 inside

monsieurtanuki: I assumed that not setting parameters meant leaving the default values: is it not the case?
The point of this PR is to match the pub.dev settings, so the less we set ourselves, the better.

M123-dev: Probably is, nevertheless, i am in favor of stronger rules. These will probably not deduct points

M123-dev (approved):
However, would not like to stop our beautiful points with such trifles. Looks good, thanks @monsieurtanuki

analysis_options.yaml
monsieurtanuki: How do you know "our" rules are stronger and will always be stronger? They change the rules regularly. Maybe one day they will say "missing_return: error" instead of "warning", and we would miss it if we override with only "warning".

If I started this PR it's because our rules were less strict (more specifically, we ignored generated files, they don't).

In addition to that - but it's not 100% related to this PR - I think there are several rules we apply in Smoothie that are neither useful nor relevant, e.g. having to specify the type of an empty `List`.
Like `List<int> list = <int>[];`
I prefer `List<int> list = [];`, easier to write and to read, and I don't see what we're missing here. But that's another story.

monsieurtanuki:
Thank you @M123-dev for your review and your remarks!

monsieurtanuki:
@M123-dev No new version like 2.0.1?

M123-dev:
Refactor isn't an accepted keyword in off-dart, here we just have fix and feat

In smoothie we manually extended this list

Issue #643, opened 2022-12-17
January 1st 2023 - Upgrade copyright year in LICENSE


monsieurtanuki:
### Why - Problem description
The copyright years are written in the `LICENSE` file.
The value must be upgraded to reflect each new year.

### What - Proposed solution
Just remember upgrading the copyright year.

PR #672, opened 2023-01-02
fix: 643 - updated the copyright year in the LICENSE

monsieurtanuki:
Impacted file:
* `LICENSE`: updated the copyright year

### What
- Now it's 2023!
- And that would trigger a new version.

### Part of 
- #643

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2018-2022 OpenFoodFacts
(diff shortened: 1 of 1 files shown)


M123-dev (approved):

monsieurtanuki:
@M123-dev Dammit, still not OK:
Capture d’écran 2023-01-03 à 16 31 18

I understand the concept of the rule: basically we should never write `final this.` in a constructor, as `this.` implies `final`.
The problem is that we don't detect that locally.
About to PR. Again.

PR #686, opened 2023-01-09
fix: correct spelling of User-Agent

monsieurtanuki:
Impacted file:
* `http_helper.dart`

### What
- We experience problems in Smoothie regarding the value of user agent for some queries.
- We recently switched those queries from POST to GET.
- And anyway the spelling of `User-Agent` was wrong.
- cf. https://github.com/openfoodfacts/smooth-app/pull/3546#issuecomment-1376043349

### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/3534

### DIFF
--- a/lib/src/utils/http_helper.dart
+++ b/lib/src/utils/http_helper.dart
@@ -203,7 +203,7 @@ class HttpHelper {
 
     headers.addAll({
       'Accept': 'application/json',
-      'UserAgent':
+      'User-Agent':
           OpenFoodAPIConfiguration.userAgent?.toValueString() ?? USER_AGENT,
       'From': OpenFoodAPIConfiguration.getUser(user)?.userId ?? FROM,
     });


teolemon (approved):
😫

monsieurtanuki:
Maybe the oldest bug in off-dart!
Capture d’écran 2023-01-09 à 19 23 25

raphael0202:
I just checked the server, and with the latest version (4.2.1+910) the user agent is indeed filled, good job! :tada: 
However, the user agent string is the following:
`- Smoothie - OpenFoodFacts - 4.2.1+910 - android+HD1903_11_F.17 - https://world.openfoodfacts.org/ -  - Smoothie - OpenFoodFacts - 4.2.1+910 - android+HD1903_11_F.17`

I find it a bit weird that it starts with an hyphen.

Issue #692, opened 2023-01-27
pub.dev score back to 130 instead of 140 :(


monsieurtanuki:
### Description
In https://pub.dev/packages/openfoodfacts we're back to 130.
It looks like new rules are applied.

### Expected behavior
A perfect score of 140!

### Screenshot
![Capture d’écran 2023-01-27 à 07 45 24](https://user-images.githubusercontent.com/11576431/215025633-71f96410-69a8-4c5b-a858-a3d4e80b3caa.png)

monsieurtanuki:
Due to the new version 4.8.0 of [json_annotation](https://pub.dev/packages/json_annotation/changelog) that requires dart 2.18 at least. Which is not possible with flutter 3.0.5.
We'll be able to fix this issue after the upgrade to flutter 3.7.

Issue #691, opened 2023-01-27
Implement "imagesToJson"


monsieurtanuki:
### Description
For some reason `JsonHelper.imagesFromJson` is not implemented yet, there's even a `TODO` about it.
That prevents us from correctly storing a `Product` in a `String` (e.g. in a database), more specifically to retrieve the `images` field.
And now we need `images` in Smoothie...

### Expected behavior
`JsonHelper.imagesFromJson`  should be implemented.

PR #693, opened 2023-01-27
fix: 691 - implemented "imagesToJson"

monsieurtanuki:
New file:
* `api_json_to_from_test.dart`: unit tests about to/from json conversions

Impacted files:
* `api_search_products_test.dart`: unrelated fix
* `json_helper.dart`: implemented method `imagesToJson`; refactored method `imagesFromJson`
* `product_images.dart`: added fields width and height; added `hashcode` and `==` operator

### What
- `imagesFromJson`was already implemented, but not `imagesToJson`.
- That was problematic if we wanted to store locally in a `String` a `Product` with field `images`.

### Fixes bug(s)
- Closes: #691
- Closes: #692

### DIFF
--- a/lib/src/model/product_image.dart
+++ b/lib/src/model/product_image.dart
@@ -134,6 +134,8 @@ class ProductImage {
     this.y1,
     this.x2,
     this.y2,
+    this.width,
+    this.height,
   });
 
   final ImageField field;
@@ -165,6 +167,12 @@ class ProductImage {
   /// Crop coordinate y2, compared to the uploaded image
   int? y2;
 
+  /// Image width.
+  int? width;
+
+  /// Image height.
+  int? height;
+
   @override
   String toString() => 'ProductImage('
       '${field.offTag}'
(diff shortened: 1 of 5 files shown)


lib/src/model/product_image.dart
```diff
@@ -179,5 +187,36 @@ class ProductImage {
       '${y1 == null ? '' : ',y1=$y1'}'
       '${x2 == null ? '' : ',x2=$x2'}'
       '${y2 == null ? '' : ',y2=$y2'}'
+      '${width == null ? '' : ',width=$width'}'
+      '${height == null ? '' : ',height=$height'}'
       ')';
+
+  @override
+  int get hashCode =>
+      '${field.offTag}_${language?.code}_${size?.offTag}'.hashCode;
```
M123-dev, line 196: Can't we just use the toString method here

M123-dev (approved):

lib/src/model/product_image.dart
monsieurtanuki: In that case we could probably, but I don't think a hashcode should depend on other unrelated methods. What if the implementation of `toString` is removed from that class?

Fun facts:
1. I don't care about `hashcode`, it's just that the operator `==` triggers a warning about un-implemented `hashcode` method
2. with your logic I could have implemented `==` with `toString` too. That would have been ugly

monsieurtanuki:

Issue #699, opened 2023-02-09
Create MatchedScoreV2 and deprecate MatchedProductV2


monsieurtanuki:
### Why - Problem description
In Smoothie we had performance problems with `MatchedProductV2`, that made us implement instead `MatchedScoreV2`, as explained in Smoothie's `MatchedScoreV2` comments:
> This is a copy of the algorithm in off-dart's [MatchedProductV2].
> The idea behind is to do the same computing, but without having to store
> the product. Because in the rest of the app we never need all the products
> in memory.
> Another solution would be to store, instead of the product, just the
> attributeGroups (id => match and status)
> Should be merged into off-dart this way:
> * add [MatchedScoreV2]
> * make [MatchedProductV2] extend [MatchedScoreV2]
> * deprecated some methods too, if possible.

PR #702, opened 2023-02-11
feat: 699 - new class MatchedScoreV2 with better perf than MatchedProductV2

monsieurtanuki:
Impacted files:
* `api_matched_product_v2_test.dart`: refactored around the test for new class `MatchedScoreV2`
* `matched_product_v2.dart`: new class `MatchedScoreV2`
* `pubspec.yaml`: unrelated upgrade

### What
- When we have tons of `Product`s to sort according to scores, it is useful to compute the score without storing the product (which can be fat), for memory performance reason.
- The new class `MatchedScoreV2` computes the score but stores just the barcode, not the whole `Product`.
- The previous class `MatchedProductV2` still exists but just extends `MatchedScoreV2`.

### Fixes bug(s)
- Closes: #699

### DIFF
--- a/lib/src/personalized_search/matched_product_v2.dart
+++ b/lib/src/personalized_search/matched_product_v2.dart
@@ -30,11 +30,14 @@ enum MatchedProductStatusV2 {
   DOES_NOT_MATCH,
 }
 
-class MatchedProductV2 {
-  MatchedProductV2(
-    this.product,
+/// Score of a product according to preferences.
+///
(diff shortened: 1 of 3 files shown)


M123-dev (approved):
Looks good👍🏼 

I just checked the docs. We don't have the MatchedProductV2 in it so no "real" breaking change. But we could add the MatchedScoreV2  to the docs

monsieurtanuki:
Actually I've never mentioned any breaking change: it's rather elegant OOP ;)

PR #705, opened 2023-02-18
fix: rollback to dart 2.17 / flutter 3.0.5

monsieurtanuki:
Impacted file:
* `pubspec.yaml`

### What
- If we don't rollback here too, we cannot use off-dart in Smoothie...


M123-dev (approved):

monsieurtanuki:
Thank you @M123-dev!

Issue #719, opened 2023-03-24
Autosuggest works erratically if the responses are not received in the right order


teolemon:
### Description
- Reported by @alexgarel, unsure how often it happens
- If the request for the first character is sent, then the second request with two characters is sent, and the client receives the response of the two-character request and then the response of the one-character request, then the auto-suggestion will occur. behave erratically. The solution to fix the problem is to check the timeline of requests.

### Part of 
- #218

monsieurtanuki:
Technically it has not that much to do with OFF typical added value.

Theoretically it's quite interesting though. It would be like the combination of a cache of results and the latest typed string.

Currently the suggestions are called each time the user types a new character (or something like that):
```dart
                  return OpenFoodAPIClient.getSuggestions(
                    tagType!,
                    language: ProductQuery.getLanguage()!,
                    country: ProductQuery.getCountry(),
                    limit: 15,
                    input: input,
                  );
```

The _displayed_ suggestions are refreshed by the `return` statement.
It looks like the problem is when "old suggestions" are eventually returned (a bit too late), they are displayed and therefore bypass the current suggestions.
I would say that in an ideal world that should be dealt with by the `Autocomplete` widget.

I can code a `SuggestionManager` to deal with that. Should it be in OFF-dart or in Smoothie, that's another story.
Let's try with OFF-dart first.

monsieurtanuki:
@stephanegigandet Just curious: have suggestions evolved recently?
Our previous tests said something like "if you type 'M', then all the suggestions will contain 'M'".
But it does not seem to be the case anymore.
Doesn't look that bad, for instance "Guyane" is suggested from "FR".
Should we then avoid the "'M' in input means 'M' in suggestions" tests in off-dart?

stephanegigandet:
@monsieurtanuki We now also return taxonomy entries that have one of their synonyms match the input string, even if it doesn't match the main name of the entry. e.g. for countries, we will also match country codes, so looking for "fr" also returns the Faroese Islands because they have the country code FRO.

https://world.openfoodfacts.org/api/v3/taxonomy_suggestions?tagtype=countries&string=fr

 Same for "us", we return "United States" and "Soviet Union" (which matches "USSR"): https://world.openfoodfacts.org/api/v3/taxonomy_suggestions?tagtype=countries&string=us

In the future, we might adjust the filtering and ranking of some taxonomies, based on context, popularity etc. like we do for packaging shapes and materials.

[...]

PR #720, opened 2023-03-25
feat: 719 - Suggestion Manager with suggestions for the latest input

monsieurtanuki:
New files:
* `api_suggestion_manager_test.dart`: Unit tests for `SuggestionManager`
* `suggestion_manager.dart`: Manager that returns the suggestions for the latest input.

Impacted files:
* `open_food_api_client.dart`: added a comment mentioning `SuggestionManager`.
* `openfoodfacts.dart`: exported new file `suggestion_manager.dart`

### What
- New class `SuggestionManager` that was needed for autocomplete widgets.
- We already had a method in order to get suggestions.
- With the manager, the user sees the most relevant suggestions (that match the latest input), regardless of the time taken by each individual call to the `getSuggestions` method (typically, when the oldest call takes more time than a more recent call).

### Fixes bug(s)
- Closes: #719


lib/src/open_food_api_client.dart
```diff
@@ -1131,10 +1143,12 @@ class OpenFoodAPIClient {
       queryType: queryType,
       addCredentialsToBody: false,
     );
-    return json.decode(response.body);
+    return HttpHelper().jsonDecode(response.body);
   }
 
   /// cf. https://openfoodfacts.github.io/openfoodfacts-server/reference/api-v3/#get-/api/v3/taxonomy_suggestions
+  ///
+  /// Consider using [SuggestionManager].
```
M123-dev, line 1151: Could you update the README

test/api_get_suggestions_test.dart
```diff
@@ -257,7 +260,7 @@ void main() {
     test('Suggestions for additives', () async {
       List<String> result = await OpenFoodAPIClient.getSuggestions(
         TagType.ADDITIVES,
-        language: OpenFoodFactsLanguage.RUSSIAN,
```
M123-dev, line 260: What impact does this have

M123-dev (approved):
Heyy @monsieurtanuki looks great besides some missing documentation in the README


Also, sorry for my late response, but now I'm done with exams for some time at least :D

test/api_get_suggestions_test.dart
monsieurtanuki: For some reason the test does not pass anymore in Russian.
As we more or less agreed that it was more relevant to run integration tests with server data than unit tests with local json files, once in a while we have to "fix" tests.

lib/src/open_food_api_client.dart
monsieurtanuki: > Could you update the README

I will once the code is merged, as I cannot point to the doc of a class that does not exist yet.

monsieurtanuki:

monsieurtanuki:
@M123-dev When is the doc refreshed btw?

PR #732, opened 2023-05-21
fix: typo for mono and poly unsaturated fat tags

monsieurtanuki:
### What
- There was a typo for mono and poly unsaturated fat tags: a trailing `-tag` was missing in both cases.
- This is fixed in this PR.

### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/3980


teolemon (approved):
👍

PR #734, opened 2023-05-22
chore: Google in-house release workflow

M123-dev:
### What
It's a new workflow provided by google itself. It seems like it doesn't require credentials. I am unsure how this will work but let's give it a try. This, we don't depend on me updating the secrets with new credentials every couple months


monsieurtanuki (approved):
@M123-dev I'm not an expert on `yml` but I approve if that can help your tests.

teolemon:
@M123-dev Would you have the source workflow ? 
I'd like to PR against our scanner dependancy so that we can get more timely releases: https://github.com/juliansteenbakker/mobile_scanner/pull/262/files

M123-dev:
@monsieurtanuki thanks. It is failing due to our changelog. We have [x.x.x](link...) but they expect just the version as header. I have proposed a PR.


@teolemon what do you mean by source workflow. You can just use what I proposed here:

``` yaml
# A CI configuration to auto-publish pub packages.
# https://github.com/dart-lang/ecosystem/tree/main/pkgs/firehose
# https://github.com/dart-lang/ecosystem/wiki/Publishing-automation

name: Publish

on:
  pull_request:
    branches: [ master ]
    types: [opened, synchronize, reopened, labeled, unlabeled]
  push:
    tags: [ 'v[0-9]+.[0-9]+.[0-9]+' ]

jobs:
  publish:
    uses: dart-lang/ecosystem/.github/workflows/publish.yaml@main
```

As far as I can tell they don't need credentials or something like that

monsieurtanuki:
Hi @M123-dev!
I would like to release 2.6.0.
Would be a good opportunity to check if your automatic release system does work, wouldn't it?

monsieurtanuki:
@M123-dev Maybe it's just bad luck...
Capture d’écran 2023-06-14 à 14 29 34

PR #736, opened 2023-05-25
feat: Added enum for product result status instead of string

ketanchoyal:
### What

- Used enum instead of string as the status of a `ProductResultV3` to handle it easily


ketanchoyal:
No problem, thanks for reviewing it

PR #748, opened 2023-06-23
feat: Allow to pass `insight_types` to `api/v1/questions/$barcode`

g123k:
Hi everyone 🖐️!

We can now filter questions by giving one or more insight types.
This is not yet deployed (waiting for https://github.com/openfoodfacts/robotoff/pull/1134), so please don't merge this PR yet.

But we'll be ready once it's OK.

### DIFF
--- a/lib/src/robot_off_api_client.dart
+++ b/lib/src/robot_off_api_client.dart
@@ -4,15 +4,15 @@ import 'package:http/http.dart';
 
 import 'model/insight.dart';
 import 'model/robotoff_question.dart';
+import 'model/robotoff_question_order.dart';
 import 'model/status.dart';
 import 'model/user.dart';
 import 'utils/country_helper.dart';
 import 'utils/http_helper.dart';
 import 'utils/language_helper.dart';
 import 'utils/query_type.dart';
-import 'model/robotoff_question_order.dart';
-import 'utils/uri_helper.dart';
 import 'utils/server_type.dart';
+import 'utils/uri_helper.dart';
 
 class RobotoffAPIClient {
   RobotoffAPIClient._();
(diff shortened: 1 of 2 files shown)


lib/src/robot_off_api_client.dart
```diff
@@ -4,15 +4,15 @@ import 'package:http/http.dart';
 
 import 'model/insight.dart';
 import 'model/robotoff_question.dart';
+import 'model/robotoff_question_order.dart';
```
monsieurtanuki, line 7: Strange that the lint check didn't complain before your PR. Or is the check only for "packages" and not for local files?

test/api_get_robotoff_test.dart
```diff
@@ -58,6 +59,22 @@ void main() {
       }
     });
 
+    test('get questions with a given insight type', () async {
```
monsieurtanuki, line 62: Will there always be category questions for this specific product?

A stronger test would perhaps
* extract insights without barcodes (`getQuestions`) - we get barcodes and insight types
* check with `getProductQuestions` (with barcode and _insight_ parameter) that we do have a question for each barcode/insight type

monsieurtanuki (approved):
@g123k LGTM!

g123k:
FYI tests are now OK for the insights.
It's complaining, because of exisiting tests.

lib/src/robot_off_api_client.dart
g123k: Good question, maybe it's also because I use Dart 3.
And I know that some rules have changed

test/api_get_robotoff_test.dart
g123k: Indeed, that's a better implementation.
I will do it that way

monsieurtanuki:
> FYI tests are now OK for the insights. It's complaining, because of exisiting tests.

For the record in off-dart we can merge without all the tests passing.

test/api_get_robotoff_test.dart
g123k: I have changed the test @monsieurtanuki, please tell me if that is what you were expected

monsieurtanuki: Thank you @g123k, that's more or less what I had in mind!

M123-dev:
Robotoff 1.30.0 seems to be published can we merge

g123k:
> Robotoff 1.30.0 seems to be published can we merge

It seems OK I think

PR #749, opened 2023-06-24
fix: get user products tests

monsieurtanuki:
### What
- While reviewing #748, I noticed an unrelated bug in the tests.
- The fix is only to set a higher number of downloaded items, so that the algorithm of the test does work.

### Impacted file
* `api_get_user_products_test.dart`: fixed a test bug (set a higher limit)


monsieurtanuki:
@teolemon Given that the TEST env is again a bit slow or broken, couldn't we use a dedicated range of fake/unreachable barcodes in PROD for all our off-dart WRITE tests? Like `978` book barcodes?

M123-dev (approved):

PR #751, opened 2023-06-27
fix: ImageHelper refactoring

monsieurtanuki:
### What
Some refactoring: in Smoothie we had to duplicate some code from off-dart just because it was private. Now it's made public.

### Impacted file
* `image_helper.dart`: new method `getBarcodeSubPath`; new `root` parameter


g123k (approved):

monsieurtanuki:

PR #761, opened 2023-07-17
chore: add permissions to release_to_pub

alexgarel:


.github/workflows/release_to_pub.dev.yml
```diff
@@ -6,11 +6,18 @@ name: Publish
 
 on:
   pull_request:
-    branches: [ master ]
+    branches: 
+      # release please PR, will only run "validate"
+      - release-please--branches--master--components--openfoodfacts
```
M123-dev, line 11: Isn't this branch already covered by the line below

M123-dev (approved):
Thanks a lot @alexgarel looks about right to me

.github/workflows/release_to_pub.dev.yml
alexgarel: I don't think so. I think that currently it was never running on any PRs because there was master (but we have no PR where "origin" branch is master !)

And I don't think it's useful to run it for every branch, or would you like to ?

PR #766, opened 2023-07-20
fix: issue with converting LocalizedTags to JSON

### LINKED ISSUE #765: ProductPacking does not convert LocalizedTags to JSON
The problem can be demonstrated using he following test, which is expected to pass, but currently does not:
```dart
import 'package:test/test.dart';

void main() {
  test('ProductPackaging bug - covers issue #752', () {
    const expectedProductName = 'Orange';
    final testProduct = Product()..productName = expectedProductName;
    const expectedQuantity = '75cl';
    final packing = ProductPackaging()..quantityPerUnit = expectedQuantity;
    const expectedId = '4';
    packing.shape = LocalizedTag()..id = expectedId;
    testProduct.packagings = [packing];
    final productJson = testProduct.toJson();
    final productRestored = Product.fromJson(productJson);
    expect(productRestored.productName, expectedProductName);
    final packingRestored = productRestored.packagings!.first;
    expect(packingRestored.quantityPerUnit, expectedQuantity);
    expect(packingRestored.shape!.id, expectedId);
  });
}
```
Error:
```
type 'LocalizedTag' is not a subtype of type 'Map<String, dynamic>' in type cast
package:openfoodfacts/src/model/product_packaging.g.dart 13:49  _$ProductPackagingFromJson
package:openfoodfacts/src/model/product_packaging.dart 62:7     new ProductPackaging.fromJson
dart:_internal                                                  ListIterable.toList
package:openfoodfacts/src/model/product.g.dart 120:12           _$ProductFromJson
package:openfoodfacts/src/model/product.dart 539:28             new Product.fromJson
test/json_conversion_issues_test.dart 14:37                     main.<fn>
```
This ticket is related to #752 which fixed another issue with converting `ProductPackaging` objects to JSON.

peterwvj:
Fixes #765


lib/src/model/localized_tag.dart
```diff
@@ -25,6 +25,9 @@ class LocalizedTag extends JsonObject {
   factory LocalizedTag.fromJson(Map<String, dynamic> json) =>
       _$LocalizedTagFromJson(json);
 
+  static Map<String, dynamic> objToJson(LocalizedTag? tag) =>
+      tag != null ? tag.toJson() : {};
+
   @override
   Map<String, dynamic> toJson() => _$LocalizedTagToJson(this);
```
monsieurtanuki, line 32: I'm no going to reopen our initial arguments regarding the use of `json_serializable` but needing both `static Map<String, dynamic> objToJson` and `Map<String, dynamic> toJson` is not among the best code I've ever read.

test/json_conversion_issues_test.dart
monsieurtanuki, line 1: I wouldn't call it `json_conversion_issues_test.dart` as in a few hours that won't be "issues" anymore but just "tests".

monsieurtanuki (approved):
Thank you @peterwvj!
Again, I'm a bit surprised that it wasn't detected in Smoothie before, as we do use localized tags there.

test/json_conversion_issues_test.dart
peterwvj: I agree, the file has now been renamed.

lib/src/model/localized_tag.dart
peterwvj: Yeah, it does (generally) feel a bit awkward working with JSON conversion is Dart/Flutter

PR #777, opened 2023-08-08
feat: Registration/Password lost: better catch 504 errors from the server

g123k:
Hi everyone,

Due to the current situation with the infrastructure, we often have 504 errors.
Here is a better catch of the errors for the registration and the password lost.
It's here limited to the user management, but it's more critical here.

### DIFF
--- a/lib/src/open_food_api_client.dart
+++ b/lib/src/open_food_api_client.dart
@@ -1224,7 +1224,7 @@ class OpenFoodAPIClient {
         body:
             'An email with a link to reset your password has been sent to the e-mail address associated with your account.',
       );
-    } else if (status.status is int && status.status != 500) {
+    } else if (status.status is int && status.status < 500) {
       return status.copyWith(status: 400);
     } else {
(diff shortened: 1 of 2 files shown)


lib/src/open_food_api_client.dart
```diff
@@ -1224,8 +1224,11 @@ class OpenFoodAPIClient {
         body:
             'An email with a link to reset your password has been sent to the e-mail address associated with your account.',
       );
-    } else {
+    } else if (status.status is int && status.status != 500) {
       return status.copyWith(status: 400);
+    } else {
+      /// Trigger real 5xx errors
+      return status;
```
M123-dev, line 1231: Say the status is 502 then this case would never trigger. Everything in the besides 500 Returns a 400

Is that your wanted behavior

g123k: Oops my bad, I created this PR while discussing and clearly that was not a good idea.

PR #779, opened 2023-08-09
feat: Multipart requests: ensure to always get the full response

g123k:
Hi everyone,

We have a major issue with POST requests, where we only get a part of the response (`first` from a `Stream`).
This PR ensures to get the full response

### DIFF
--- a/lib/src/utils/http_helper.dart
+++ b/lib/src/utils/http_helper.dart
@@ -183,7 +183,7 @@ class HttpHelper {
       try {
         return Status.fromJson(jsonDecode(responseBody));
       } catch (e) {
-        //When the server returns html instead of json
+        //When the server returns html instead of JSON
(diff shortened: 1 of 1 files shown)


lib/src/utils/http_helper.dart
```diff
@@ -173,22 +176,32 @@ class HttpHelper {
...
-        return Status(
-            status: response.statusCode, error: response.reasonPhrase);
+    http.StreamedResponse response = await request.send();
+
+    if (response.statusCode == 200) {
+      String responseBody = await _extractResponseAsString(response);
+      try {
+        return Status.fromJson(jsonDecode(responseBody));
+      } catch (e) {
+        //When the server returns html instead of json
```
teolemon, line 186:
```suggestion
        //When the server returns html instead of JSON
```

PR #781, opened 2023-08-10
feat: Reset password: allow to receive "translated" email

g123k:
Hi everyone,

When we want to reset the password of the user, we use the world instance of OFF.
By providing a country and/or language, the user will then receive the localized content


teolemon (approved):

peterwvj (approved):
Thanks!

PR #783, opened 2023-08-12
fix: unit test values

monsieurtanuki:
### What
- Minor fixes about unit tests that rely on actual data. The data changed, we slightly change our tested values.


peterwvj (approved):
Thanks

monsieurtanuki:
Thank you @peterwvj for your review!

Issue #785, opened 2023-08-12
Listing unknown attributes for unknown matches


monsieurtanuki:
### Why - Problem description
cf. @teolemon's https://github.com/openfoodfacts/smooth-app/issues/1118
When we try to match a product with the user preferences, sometimes we get an "unknown match".
We just know it's unknown, and we don't list what is missing for a correct yes/no match.

### What - Proposed solution
Instead of just saying the match is unknown, we should list which attributes are missing.

teolemon:
Well, if we could have something pedagogical about the computation as well. I'm unsure whether we'll be easy showing it at first, but it would be interesting to have in Dev mode.

monsieurtanuki:
@teolemon Understood: I will extend that principle to "may not match" and "does not match".
For the record there are 8 different cases for 6 different match statuses (very good, good, poor, unknown, may not, does not), and in 3 cases we put "unknown" (no attributes at all, at least one mandatory attribute that is unknown, too many unknown attributes).

monsieurtanuki:
@teolemon @stephanegigandet Btw I think there's a flaw in the score computation, as we loop on the product attributes but not on the preference importances.

That means that a mandatory preference is ignored if the product doesn't have the related attribute at all.

It's OK as long as a product has
* either no attributes at all (probably a corrupted product), as we flag it as "unknown"
* or all the attributes populated.

stephanegigandet:
@monsieurtanuki The server always compute all attributes so we should be fine.

stephanegigandet:
I'm not sure if creating a separate list with explanations is the right approach: that kinds of duplicates the information we have on the attribute cards that are displayed on top of the product page.

e.g. on the website we have something like that on top of the product page:

![image](https://github.com/openfoodfacts/openfoodfacts-dart/assets/8158668/7dd87cc8-6a5a-4a3e-9cc8-6584893ba89b)

Related to that, is it planned to add the % score on the mobile app, like we have on the website?

Could we also adopt the same translations on the website + app for the strings like "Does not match" etc. e.g. in French on the app we have "Pas de compatibilité" while on the website we have "Ne correspond pas"

monsieurtanuki:
@stephanegigandet You're right, there's some kind of duplication here, and it may not be necessary.
Maybe I over-interpreted the initial needs.
I needed this step though, to understand the product score/match computations.

Basically there are 3 reasons why the product match is unknown:
[...]

PR #786, opened 2023-08-12
feat: 785 - list of problematic attributes as score explanation

monsieurtanuki:
### What
- The PR provides, as an option, lists of problematic attributes when the score is not explicit (unknown match, may not match, does not match).
- With this, we can be more verbose than "It's not OK", and provide localized explanations about the score, e.g. `'Caractère végétarien inconnu'` (FR) / `'Vegetarian status unknown'` (EN).

### Fixes bug(s)
- Closes: #785

### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/1118

### Impacted files
* `api_matched_product_v1_test.dart`: minor refactoring
* `api_matched_product_v2_test.dart`: added a test about score explanations
* `api_product_preferences_test.dart`: minor refactoring
* `available_attribute_groups.dart`: refactored with language instead of language code
* `available_preference_importances.dart`: refactored with language instead of language code
* `matched_product_v2.dart`: added an optional parameter that lists the problematic attributes when the status unknown, does not match and may not match

### DIFF
--- a/lib/src/personalized_search/available_attribute_groups.dart
+++ b/lib/src/personalized_search/available_attribute_groups.dart
@@ -1,5 +1,6 @@
 import '../model/attribute_group.dart';
 import '../utils/http_helper.dart';
+import '../utils/language_helper.dart';
 
 /// Referential of attribute groups, with loader.
 class AvailableAttributeGroups {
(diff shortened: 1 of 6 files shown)


stephanegigandet:
Just curious: where will those explanations be displayed?

monsieurtanuki:
I guess we would be better off trying to code it directly in Smoothie with incremental changes, and different UI/UX.
And then, if relevant, move some code from Smoothie to off-dart.
Therefore, closing here.

PR #787, opened 2023-08-16
fix: reset password checks with languages

monsieurtanuki:
### What
- Not that long ago there was no country or language parameters on reset password. We lousily tried to detect sentences in the resulting html.
- Now that we have country or language parameters, we get a localized html. The sentences we expected are localized. That's why in this PR we use html tags instead of localized terms.
- That's still very lousy: we would be better off calling a more appropriate server URL that would return a json. But that does not exist AFAIK.

### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/4549

### Impacted files
* `open_food_api_client.dart`: relying on html tags for error detection instead of English localizations
* `user_management_test_prod.dart`: additional tests with / without language/country and on existing / not existing users

### DIFF
--- a/lib/src/open_food_api_client.dart
+++ b/lib/src/open_food_api_client.dart
@@ -1186,8 +1186,10 @@ class OpenFoodAPIClient {
   }
 
   /// Uses reset_password.pl to send a password reset Email
-  /// needs only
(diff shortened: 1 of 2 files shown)


monsieurtanuki:

PR #790, opened 2023-08-22
feat!: A valid user agent is now required

g123k:
Hi everyone,

Starting with the next version of the SDK, all requests must have a valid `user agent` (at least with a `name`).
We don't want to implement complex things, like API keys, but due to massive spam, we need to ensure to identify each application.


lib/src/model/user_agent.dart
```diff
@@ -1,19 +1,28 @@
...
 
   const UserAgent({
-    this.name,
+    required this.name,
     this.version,
     this.system,
     this.url,
     this.comment,
-  });
+  }) : assert(name.length > 1, 'A non empty name is required');
```
monsieurtanuki, line 25:
```suggestion
  }) : assert(name.trim().isNotEmpty, 'A non empty name is required');
```

monsieurtanuki (commented):
Hi @g123k!

lib/src/model/user_agent.dart
g123k: Initially that was my change, but calling trim() will require to remove the `const` constructor.
I have no strong opinion on this

monsieurtanuki: I have a stronger opinion than you on that matter then: I prefer the `trim` that will reject junk data (and we may make the rules evolve in the future), to the usage of `const` and its highly hypothetical performance gains.

g123k: Ok, I will `trim` instead

g123k:
It should now be OK @monsieurtanuki

test/test_constants.dart
```diff
@@ -1,7 +1,8 @@
 import 'package:openfoodfacts/openfoodfacts.dart';
 
 class TestConstants {
-  static const UserAgent TEST_USER_AGENT = UserAgent(
+  // ignore: non_constant_identifier_names
+  static UserAgent TEST_USER_AGENT = UserAgent(
```
monsieurtanuki, line 5: Just curious: in which cases would it make sense to change the value? `const` seemed appropriate, didn't it?

test/configuration_test.dart
```diff
@@ -201,7 +215,14 @@ void main() {
...
     );
   });
 }
+
+String get _appNameValue => 'app_name=${Uri.encodeFull(
+      OpenFoodAPIConfiguration.userAgent!.name.replaceAll(
+        ' ',
+        '+',
+      ),
+    )}';
```
monsieurtanuki, line 228:
```suggestion
String get _appNameValue => 'app_name=${Uri.encodeComponent(
      OpenFoodAPIConfiguration.userAgent!.name,
    )}';
```

monsieurtanuki (commented):
Thank you @g123k for your changes!
I still don't agree with `assert`, that is not applied in `release` mode. You changed some `assert` to exceptions, and I guess you can change the remaining `assert`s too.
A couple of other minor suggested changes too.

test/configuration_test.dart
g123k: Unfortunately it won't work, because the url encoder, transform spaces in `%20`, whereas we need `+` here.

test/test_constants.dart
g123k: As I've previously mentioned, the `trim()` forces us to remove the `const`

test/configuration_test.dart
monsieurtanuki: @g123k Understood. Still, using `encodeComponent` instead of `encodeFull` is more readable.

test/test_constants.dart
monsieurtanuki: @g123k `final` would probably work, then.

g123k: Indeed, you're totally right

monsieurtanuki (approved):
Hi @g123k!
Thank you for taking into account my remarks!

I've just noticed that you make this PR a breaking change, which sort of makes sense but "is a bit expensive" for just the removal of a `?`.
Two separate suggestions:
* transform your `feat!` into a `feat` for not making it a breaking change
* keep your `feat!` without creating a new release immediately, so I can get rid of deprecated code (things that I do only in "breaking change" mode or after one year)

g123k:
> Hi @g123k! Thank you for taking into account my remarks!
> 
> I've just noticed that you make this PR a breaking change, which sort of makes sense but "is a bit expensive" for just the removal of a `?`. Two separate suggestions:
> 
> * transform your `feat!` into a `feat` for not making it a breaking change
> * keep your `feat!` without creating a new release immediately, so I can get rid of deprecated code (things that I do only in "breaking change" mode or after one year)

Actually this is a change from @M123-dev yesterday.
But it can indeed be considered as a breaking change, as we want a rapid move to this new mechanism.
In that case, I think you should deprecate/remove all the things you want at the same time

monsieurtanuki:
> I think you should deprecate/remove all the things you want at the same time

I'll do that today.

monsieurtanuki:
@g123k Btw the tests fail, at least for a `+` vs. `%20` confusion. Please fix them.

g123k:
> @g123k Btw the tests fail, at least for a `+` vs. `%20` confusion. Please fix them.

This is due to `Uri.encodeComponent`. I have to remove it and go back to the initial impl

test/configuration_test.dart
```diff
@@ -201,7 +215,13 @@ void main() {
...
     expect(
       uri1.toString(),
-      'http://world.openfoodfacts.org/test/test.pl?test=true&queryType=PROD',
+      'http://world.openfoodfacts.org/test/test.pl?test=true&queryType=PROD&$_appName',
     );
   });
 }
+
+String get _appName =>
+    'app_name=${OpenFoodAPIConfiguration.userAgent!.name.replaceAll(
```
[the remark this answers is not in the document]
monsieurtanuki: That won't work with special characters, e.g. `&`.

g123k:
> > @g123k Btw the tests fail, at least for a `+` vs. `%20` confusion. Please fix them.
> 
> This is due to `Uri.encodeComponent`. I have to remove it and go back to the initial impl

True, but that's only here for a hard-coded test

M123-dev:
Guess we can merge this now, can't we?

peterwvj:
Thank you for looking into this. Just for the sake of clarity, what information exactly would you prefer that the user agent contains? The bundle ID (iOS), package (Android),  the display name of the app, the version of the app, the platform it runs on?

monsieurtanuki:
> Guess we can merge this now, can't we?

@M123-dev Now we can't, because of https://github.com/openfoodfacts/openfoodfacts-dart/pull/790/files#r1305927902

@g123k Given that your current priorities may not let you fix this PR before several weeks, do you mind if I fix it?

g123k:
> > Guess we can merge this now, can't we?
> 
> @M123-dev Now we can't, because of [#790 (files)](https://github.com/openfoodfacts/openfoodfacts-dart/pull/790/files#r1305927902)
> 
> @g123k Given that your current priorities may not let you fix this PR before several weeks, do you mind if I fix it?

If that doesn't bother you, yes please do so.

README.md
```diff
@@ -21,6 +21,17 @@ We use the ability of the Open Food Facts API to return products results in JSON
 
 This plugin also allows you to edit a product or upload a new one to Open Food Facts. Using the same simple product structure you can create a product object or edit an existing one and send it to the API using a single function.
 
+## Migration to >= 2.11.x
```
monsieurtanuki, line 24:
```suggestion
## Migrating from 2.x.x to 3.x.x (breaking changes)
```

README.md
```diff
@@ -21,6 +21,17 @@ We use the ability of the Open Food Facts API to return products results in JSON
 
 This plugin also allows you to edit a product or upload a new one to Open Food Facts. Using the same simple product structure you can create a product object or edit an existing one and send it to the API using a single function.
 
+## Migrating from 2.x.x to 3.x.x (breaking changes)
+
+Starting with version 2.11, we now enforce all clients to provide a valid user agent.
```
monsieurtanuki, line 26:
```suggestion
Starting with version 3.0.0, we now enforce all clients to provide a valid user agent.
```

lib/src/model/user_agent.dart
```diff
@@ -1,19 +1,28 @@
...
-  const UserAgent({
-    this.name,
+  UserAgent({
+    required this.name,
     this.version,
     this.system,
     this.url,
     this.comment,
-  });
+  }) : assert(name.trim().length > 1, 'A non empty name is required');
```
monsieurtanuki, line 25:
```suggestion
  }) {
    if (name.trim().isEmpty) {
      throw Exception('A non empty name is required');
    }
  }
```

lib/src/utils/http_helper.dart
```diff
@@ -214,10 +211,14 @@ class HttpHelper {
   }) {
     Map<String, String>? headers = {};
 
+    assert(
+      OpenFoodAPIConfiguration.userAgent != null,
+      'A User-Agent must be set before calling this method',
+    );
```
monsieurtanuki, line 217:
```suggestion
    if (OpenFoodAPIConfiguration.userAgent == null) {
      throw Exception('A User-Agent must be set before calling this method');
    }
```

monsieurtanuki:
> Thank you for looking into this. Just for the sake of clarity, what information exactly would you prefer that the user agent contains? The bundle ID (iOS), package (Android), the display name of the app, the version of the app, the platform it runs on? Should the docs be more clear on this?

@peterwvj cf. `README.md`:
```dart
OpenFoodAPIConfiguration.userAgent = UserAgent(
  name: '<Name of your app>',
);
```

monsieurtanuki:
@g123k In the end I've coded only minor changes. I've finally understood what we couldn't agree about - my bad, the code I was not happy about was indeed only in the tests, and the real code will indeed escape all the potentially dangerous characters.

PR #791, opened 2023-08-25
fix!: removed deprecated code

monsieurtanuki:
### What
- Removed deprecated code, as we're about to merge another PR that is a breaking change too.


g123k (approved):
👍

monsieurtanuki:

peterwvj (approved):
Great, thanks!

monsieurtanuki:
Thank you @peterwvj for your review!

As it's a rather useless PR (only removed deprecated code) but still a breaking change, I prefer to wait for #790 (also a breaking change) so that we can make a single breaking change release with 2 breaking change PRs.

PR #795, opened 2023-09-07
fix!: fix for the uk/gb country code case

monsieurtanuki:
### What
- In Smoothie we detected a possible bug regarding our usage of 'uk' as United Kingdom country code - while the official country code is 'gb'.
- It got fixed in Smoothie - the current PR is about fixing it in off-dart too.
- Meanwhile, country related minor issues were fixed. Those are minor breaking changes, that will be embedded in the already "breaking change" next release.

### Impacted files
* `country_helper.dart`: `fromOffTag` now includes a fix for the uk/gb case; breaking change - removed class `CountryHelper`
* `open_food_api_client.dart`: breaking change - `getOrderedNutrients` now uses parameter `OpenFoodFactsCountry country`
* `ordered_nutrient_test.dart`: minor refactoring

### DIFF
--- a/lib/src/open_food_api_client.dart
+++ b/lib/src/open_food_api_client.dart
@@ -1084,14 +1084,14 @@ class OpenFoodAPIClient {
   ///   print(orderedNutrients.nutrients[10].name); // Vitamin A
   /// ```
   static Future<OrderedNutrients> getOrderedNutrients({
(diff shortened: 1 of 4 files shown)


M123-dev (approved):
👍🏼

monsieurtanuki:

Issue #69, opened 2021-01-21
[RFC] Support for Open Food Facts variants


teolemon:
Request for comments: 
Is there any way to add support for Open Food Facts variants such as ?

- Open Beauty Facts - https://world.openbeautyfacts.org/
- Open Pet Food Facts - https://world.openpetfoodfacts.org/
- Open Products Facts - https://world.openproductsfacts.org/

Is there a way to do it that would not break existing apps ? Or would we need to clone the package ?

monsieurtanuki:
@teolemon We now have a `ServerType` enum that would be helpful:
```dart
/// Type of Server (e.g. "food facts", "beauty facts", ...).
enum ServerType implements OffTagged {
  /// Open Food Facts
  openFoodFacts(offTag: 'off'),
  /// Open Beauty Facts
  openBeautyFacts(offTag: 'obf'),
  /// Open Pet Food Facts
  openPetFoodFacts(offTag: 'opff'),
  /// Open Product Facts
  openProductFacts(offTag: 'opf');
  const ServerType({
    required this.offTag,
  });
  @override
  final String offTag;
}
```

In off-dart this is how we compute the URLs:
```dart
        host: OpenFoodAPIConfiguration.getQueryType(queryType) == QueryType.PROD
            ? OpenFoodAPIConfiguration.uriProdHost // for the moment, 'world.openfoodfacts.org'
            : OpenFoodAPIConfiguration.uriTestHost, // for the moment, 'world.openfoodfacts.net'
```
We could add a `ServerType` parameter to get an easy access to other servers

I don't think it makes much sense maintaining two variables for the host (`uriProdHost`, `uriTestHost`) as nobody uses both at the same time.
I guess we would be better off with one variable (e.g. `host`) and 2 methods:
* `setForcedHost(final String host)` that would force the host to a specific value - e.g. for custom servers
* `setHost(final QueryType queryType, final ServerType serverType)` for "normal" use - e.g. I'm using open beauty facts on PROD

That would mean a (minor) breaking change.

monsieurtanuki:
Actually we can already point to open beauty facts today by changing `OpenFoodAPIConfiguration.uriProdHost`.
Would beauty product data fit in a food product data template? Probably not, and we need specs for that.

monsieurtanuki:
@teolemon As long as products have the same structure in all servers, yes we can.

PR #796, opened 2023-09-07
feat: 69 - method to get a server type from its off tag

monsieurtanuki:
### What
- We forgot to implement a method to get a server type from an off tag (e.g. `off`, `obf`, `opf` and `opff`).
- We need this method for Smoothie in order to store the server type into preferences.

### Part of 
- #69

### Impacted file
* `server_type.dart`: implemented the `fromOffTag` method


M123-dev (approved):
Haven't  checked the SeverType model but looks right to me 🔝

PR #799, opened 2023-09-07
feat: product "getAttribute" method

monsieurtanuki:
### What
- In Smoothie we needed this `product.getAttribute` method and had to implement it there.
- It's cleaner if it's in off-dart.

### Impacted file
* `product.dart`: new `getAttribute` method


teolemon (approved):
Thanks @monsieurtanuki 👌

PR #800, opened 2023-09-07
feat: method to get the Uri host (e.g. for ping)

monsieurtanuki:
### What
- This is a method we had to implement in Smoothie.
- Typical use-case: trying to ping the server after a connection failed.
- Now we can get the uri host directly.

### Impacted file
* `uri_helper.dart`: new `getUriHost` method

### DIFF
--- a/lib/src/utils/uri_helper.dart
+++ b/lib/src/utils/uri_helper.dart
@@ -23,15 +23,18 @@ class UriHelper {
   }) =>
       Uri(
         scheme: OpenFoodAPIConfiguration.uriScheme,
-        host: OpenFoodAPIConfiguration.getQueryType(queryType) == QueryType.PROD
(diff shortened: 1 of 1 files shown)


teolemon (approved):
Nice, @g123k is about to open a related issue in Smoothie (pinging other servers when the product is not found in Open Food Facts)

monsieurtanuki:
> Nice, @g123k is about to open a related issue in Smoothie (pinging other servers when the product is not found in Open Food Facts)

Thank you @teolemon for your reviews!

The thing is that we don't have the other servers' hosts for the moment.
Do we use the same world.openfoodfacts.org host for all, or different hosts? Something like
* world.openbeautyfacts.org
* world.openpetfoodfacts.org
* world.openproductsfacts.org

Issue #802, opened 2023-09-09
OBF OPF OPFF currently incompatible with api v3


monsieurtanuki:
### Problem
The main focus of Smoothie is currently OFF.
The next step is to explore OBF, OPF and OPFF too. Something like: "Hmm the barcode you've just scanned is not food, but pet food".
For that we need some consistency between OFF, OBF, OPF and OPFF.

The problem is that for OBF, even if we make an API V3 query, we get an old-fashioned API V1 result (https://world.openbeautyfacts.org/api/v3/product/3600550964738).
For instance, `status` is an `int` (and we expect a `String?` in API V3). And the packagings are different too.

### Proposed solution
Possible solutions:
1. just know that OBF, OPF and OPFF must be queried with an api version < 3
=> for that we would need to re-implement a now removed different `getProduct` method (V < 3) and maintain both methods. Not a lot of code, but possibly confusion for the users of the package.
2. implement a different file for each source - `open_food_api_client.dart`, `open_beauty_api_client.dart`, ...
=> we would have to play it smart and not recopy all the code. Would a "beauty" product have the same fields as a "food" product anyway?
3. ask the server to really implement api V3 for OBF OPF OPFF
=> minimum impact on openfoodfacts-dart. That still does not solve the possible differences between a "beauty" and a "food" product.

PR #806, opened 2023-09-20
feat: 802 - new "get product" method specific to OBF OPF OPFF

monsieurtanuki:
### What
- OBF OPF OPFF do no support API V3.
- As we switched for OFF to full API V3, we needed to reintroduce a support of API V<3.
- This is done with `getOldProduct`, that work with OBF OPF OPFF (and also OFF).
- Perhaps another solution would be to ignore the "get unique product(barcode)" method and consider it just a special case of "get all products(barcodes)".
- Anyway, from now on we can retrieve OBF OPF OPFF products from a single barcode with `getOldProduct`.

### Fixes bug(s)
- Closes: #802

### Files
New files:
* `api_not_food_get_product_test.dart`: tests around OBF OPF OPFF and `getOldProduct`
* `old_product_result.dart`: Product Result (old style).
* `old_product_result.g.dart`: generated

Impacted files:
* `open_food_api_client.dart`: new `getOldProduct` method, dedicated to OBF OPF OPFF, and supposed to disappear when they support API V3.
* `openfoodfacts.dart`: exported new class.

### DIFF
--- a/lib/openfoodfacts.dart
+++ b/lib/openfoodfacts.dart
@@ -28,6 +28,7 @@ export 'src/model/nutriments.dart';
 // export 'src/model/product_list.dart'; // not needed
 export 'src/model/ocr_ingredients_result.dart';
 export 'src/model/ocr_packaging_result.dart';
+export 'src/model/old_product_result.dart';
 export 'src/model/ordered_nutrient.dart';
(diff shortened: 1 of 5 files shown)


M123-dev (approved):
👍🏼

monsieurtanuki:

PR #809, opened 2023-09-27
chore: preparation of the 3.0.0 release

monsieurtanuki:
### What
- The initial idea of this PR was to add some items in the migration instruction, because of the breaking changes.
- As a side effect, `UriProductHelper` and `ImageHelper` were slightly refactored, with some clarifications about the places where the web stores the images.

### Impacted files
* `api_add_product_image_test.dart`: minor refactoring
* `api_get_product_test.dart`: minor refactoring
* `api_not_food_get_product_test.dart`: minor refactoring
* `configuration_test.dart`: minor refactoring
* `image_helper.dart`: methods `getProductImageRootUrl`, `getBarcodeSubPath` are moved to `UriProductHelper`; method `buildUrl` loses one parameter; method `buildUrl` renamed as `getLocalizedProductImageUrl`
* `open_food_api_configuration.dart`: minor refactoring
* `product_helper.dart`: minor refactoring
* `README.md`: added migration instructions
* `uri_helper.dart`: now we use `domain` instead of `host` for `UriProductHelper`; methods `getProductImageRootUrl`, `getBarcodeSubPath` are moved from `ImageHelper`

### DIFF
--- a/lib/src/utils/json_helper.dart
+++ b/lib/src/utils/json_helper.dart
@@ -138,8 +138,8 @@ class JsonHelper {
         }
         final int? width = JsonObject.parseInt(numberObject['w']);
         final int? height = JsonObject.parseInt(numberObject['h']);
+        final String? url = numberObject['url'];
 
-        // TODO(monsieurtanuki): add field "url"?
         var image = ProductImage(
           field: field,
           size: size,
@@ -154,6 +154,7 @@ class JsonHelper {
           y2: y2,
           width: width,
           height: height,
+          url: url,
         );
         imageList.add(image);
       }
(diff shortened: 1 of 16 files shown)


lib/src/utils/json_helper.dart
```diff
@@ -191,19 +192,22 @@ class JsonHelper {
         continue;
       }
       final Map<String, dynamic> item = <String, dynamic>{};
-      item['sizes'] = <String, Map<String, int>>{};
+      item['sizes'] = <String, Map<String, dynamic>>{};
```
g123k, line 195: `dynamic` implies the object can be null.
Is-it the case?

monsieurtanuki: It cannot be `null` but now it can also be a `String` (for `url`) (and not only `int` for sizes).
Would you prefer `Object` to `dynamic` or is `dynamic` finally good enough?

g123k: My point here was indeed that if it can't be null, please use an `Object` instead.

monsieurtanuki: > My point here was indeed that if it can't be null, please use an `Object` instead.

Good point!

g123k (approved):
LGTM 👍

monsieurtanuki:
Thank you @g123k for your review and your suggestion!

Issue #816, opened 2023-10-26
Default productName and languages


monsieurtanuki:
### Description
Starting about 3 days ago, the unit tests don't pass anymore, cf. https://github.com/monsieurtanuki/openfoodfacts-dart/actions/runs/6653878935/job/18080780143

When we ask for the `product_name`, we expect it to match the language (`lc`) when relevant.

In the following case, that means we expected `product_name` to be `"Pfannkuchen"`, as we ask for a German localization and there's a localization in German.
https://world.openfoodfacts.net/api/v3/product/2222222222225?fields=code,product_name,product_name_languages&lc=de&cc=de
```
{
    "code": "2222222222225",
    "errors": [],
    "product": {
        "code": "2222222222225",
        "product_name": "Pancakes",
        "product_name_languages": {
            "de": "Pfannkuchen",
            "en": "Pancakes",
            "ru": "Блинчики"
        }
    },
    "result": {
        "id": "product_found",
        "lc_name": "Produkt gefunden",
        "name": "Product found"
    },
    "status": "success",
    "warnings": []
}
```

### Expected behavior
We expect the unit tests to pass.
Other solution: if the business rules changed regarding the default language for `product_name`, we need to reflect that in our unit tests.

monsieurtanuki:
Maybe related to https://github.com/openfoodfacts/smooth-app/issues/4672

PR #817, opened 2023-10-28
fix: 816 - use localized versions of productName and ingredientsText

monsieurtanuki:
### What
- From about a week ago, `productName` and `ingredientsText` could not be 100% trusted anymore in a multilingual context.
- More specifically, you could not trust the product name to be in German if you asked for a list of languages with German first (on a product that has German and other translations).
- The solution is to use the localized version of `productName`. And as a fallback, `productName`.
- Same thing for `ingredientsText`.
- The unit tests failed and that's the reason of the current PR.

### Fixes bug(s)
- Fixes: #816

### Impacted files
* `api_get_save_product_test.dart`: removed tests on productName and ingredientsText; refactored
* `product.dart`: added comments about using rather the localized fields for productName and ingredientsText

### DIFF
--- a/lib/src/model/product.dart
+++ b/lib/src/model/product.dart
@@ -85,11 +85,17 @@ enum ProductImprovementCategory {
 /// or update data in specific language: https://github.com/openfoodfacts/openfoodfacts-dart/blob/master/DOCUMENTATION.md#about-languages-mechanics
 @JsonSerializable()
 class Product extends JsonObject {
+  /// Barcode of the product. Will very very very often be not null.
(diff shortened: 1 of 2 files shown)


g123k (approved):
Seems indeed better to add explanations in the SDK.
Thanks @monsieurtanuki!

monsieurtanuki:
> Seems indeed better to add explanations in the SDK. Thanks @monsieurtanuki!

Thank you @g123k for your review! If you have time please consider reviewing https://github.com/openfoodfacts/smooth-app/pull/4746 too.

PR #821, opened 2023-11-15
chore: pnnsGroup refactoring

monsieurtanuki:
### What
- Some refactoring around pnns groups, that were using a slightly obscure syntax that we can avoid now due to dart syntax evolution.

### Impacted files
* `pnns_group2_filter.dart`: minor refactoring
* `pnns_groups.dart`: used a clearer syntax

### DIFF
--- a/lib/src/utils/pnns_groups.dart
+++ b/lib/src/utils/pnns_groups.dart
@@ -1,85 +1,12 @@
+import 'package:openfoodfacts/src/model/off_tagged.dart';
+
 /// PNNS food groups
 ///
 /// PNNS is the French "Programme National Nutrition Santé"
 enum PnnsGroup1 {
-  BEVERAGES,
-  CEREALS_AND_POTATOES,
-  COMPOSITE_FOODS,
-  FAT_AND_SAUCES,
-  FISH_MEAT_AND_EGGS,
-  FRUITS_AND_VEGETABLES,
-  MILK_AND_DAIRIES,
-  SALTY_SNACKS,
-  SUGARY_SNACKS
-}
-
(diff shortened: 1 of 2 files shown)


monsieurtanuki:
> Error: Unable to resolve action `codecov/codecov-action@v4`, unable to find version `v4`

@teolemon Looks familiar?

teolemon:
ah snap, I forgot about this one, and merged it @monsieurtanuki

lib/src/utils/pnns_groups.dart
```diff
@@ -1,233 +1,276 @@
...
+  FRUIT_NECTARS(
+    offTag: 'fruit-nectars',
+    name: 'Fruits nectars',
+  ),
+  TEAS_AND_HERBAL_TEAS_AND_COFFEES(
+    offTag: 'teas-and-herbal-teas-and-coffees',
+    name: 'Teas and herbal teas and coffees',
+  ),
+
+  BREAKFAST_CEREALS(
```
M123-dev, line 137: Why are there sometimes new lines between the subgroups and sometimes not

M123-dev (approved):
Thanks @monsieurtanuki, just one minor inconsistency but thats nothing to block this PR on

lib/src/utils/pnns_groups.dart
monsieurtanuki: > Why are there sometimes new lines between the subgroups and sometimes not

The usual question: the empty lines were there the way it was written previously (in that specific case, between level-1 groups of level-2 groups).
It doesn't seem to trigger lints warning so why not. Of course we could get rid of those useless empty lines.

PR #823, opened 2023-11-15
feat: new field pageCount for searchResult

monsieurtanuki:
### What
- New field `pageCount` for `SearchResult`
- That will help us solve a minor rare discrepancy between the number of "returned products", the "number of returned products" (`pageCount`) and the "total number of products" (`count`).

### Impacted files:
* `api_get_user_products_test.dart`: now using `pageCount` instead of `count`
* `api_search_products_test.dart`: skipped a test failing because of the server
* `search_result.dart`: added field `pageCount`
* `search_result.g.dart`: generated

### DIFF
--- a/lib/src/model/search_result.dart
+++ b/lib/src/model/search_result.dart
@@ -15,6 +15,9 @@ class SearchResult extends JsonObject {
   @JsonKey(name: 'count', fromJson: JsonObject.parseInt)
   final int? count;
 
+  @JsonKey(name: 'page_count', fromJson: JsonObject.parseInt)
+  final int? pageCount;
+
   @JsonKey(name: 'skip', fromJson: JsonObject.parseInt)
   final int? skip;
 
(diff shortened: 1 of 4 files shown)


g123k (approved):
LGTM 👍
This may worth a bit of doc to explain the difference between count/pageCount

lib/src/model/search_result.dart
```diff
@@ -15,6 +15,9 @@ class SearchResult extends JsonObject {
   @JsonKey(name: 'count', fromJson: JsonObject.parseInt)
   final int? count;
```
teolemon, line 17: // page_count is the number of products requested per page. Please be mindful of server resources, and keep that below 200.

monsieurtanuki:

lib/src/model/search_result.dart
monsieurtanuki: > page_count is the number of products requested per page.

AFAIK it's rather the number of products _found_ within that result page.

PR #828, opened 2023-11-16
feat: new localized product fields for generic name

monsieurtanuki:
### What
- One test started to fail yesterday.
- Obviously something changed on the server side regarding `genericName`, the same way it changed weeks ago about `productName`. In short, now you're supposed to be explicit regarding the language you use, which is a good news. I fixed the test.
- En passant, I also added a localized version of "generic name".

### Impacted files
* `api_get_save_product_test.dart`: set explicit language for `genericName` test to pass; also tested new field `genericNameInLanguages`
* `product.dart`: added field `genericNameInLanguages`
* `product.g.dart`: generated
* `product_fields.dart`: added fields `GENERIC_NAME_IN_LANGUAGES` and `GENERIC_NAME_ALL_LANGUAGES`

### DIFF
--- a/lib/src/model/product.dart
+++ b/lib/src/model/product.dart
@@ -103,10 +103,17 @@ class Product extends JsonObject {
       includeIfNull: false)
   Map<OpenFoodFactsLanguage, String>? productNameInLanguages;
 
-  ///Common name
-  ///Example: Chocolate bar with milk and hazelnuts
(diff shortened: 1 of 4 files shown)


M123-dev (approved):
👍🏼

monsieurtanuki:
> 👍🏼

👍

Issue #829, opened 2023-11-17
Cannot login with a non-ASCII username


monsieurtanuki:
cf. https://github.com/openfoodfacts/smooth-app/issues/4813
cf. https://openfoodfacts.sentry.io/issues/3458725816/?project=5376745&referrer=github_integration
Managed to reproduce the issue trying to connect (`login2`) as `User(userId: 'លីវយី', password: '123')`

Looks like the problem comes from `From` (haha) in the http header.
It's expected to be an email address (cf. https://datatracker.ietf.org/doc/html/rfc9110#name-from).
But we put there the user name.

Looks like only ASCII or ISO-8859-1 are 100% expected to work.

g123k:
Correct me if I'm wrong, but it's more like a server issue than a Dart one?

monsieurtanuki:
> Correct me if I'm wrong, but it's more like a server issue than a Dart one?

I don't think so, because WE are sending a header that is an INVALID http header, regardless of the server.
That said, there may be an impact on the off server side depending on the way they need/use `From`.
@teolemon @stephanegigandet do you guys need that `From`, e.g. for your stats, and would a base64 version of the `From` be good enough for you?

monsieurtanuki:
Fixed by #830.

PR #830, opened 2023-11-20
fix: 829 - no more crash if the user name is not ASCII/ISO-8859-1

monsieurtanuki:
### What
- We populate the `From` http header with the user name
- Which may be problematic, as data in the http header is supposed to be ASCII/ISO-8859-1 encoded.
- In order to prevent the http header from being rejected, we transform user name if from another charset.
- cf. https://github.com/openfoodfacts/smooth-app/issues/4813

### Fixes bug(s)
- #829

### Impacted files
* `http_helper.dart`: safer `From` http header.
* `user_management_test_test_env.dart`: new test if the login crashes with a non-ASCII/ISO-8859-1 username

### DIFF
--- a/lib/src/utils/http_helper.dart
+++ b/lib/src/utils/http_helper.dart
@@ -96,7 +96,7 @@ class HttpHelper {
       }
     }
 
-    http.Response response = await http.post(
+    return http.post(
       uri,
       headers: _buildHeaders(
         user: user,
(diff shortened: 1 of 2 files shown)


g123k (approved):
The code is OK for me (and thanks for the fix).

I just wonder if we shouldn't document this somewhere to inform that we do this change

monsieurtanuki:

> I just wonder if we shouldn't document this somewhere to inform that we do this change

That's a good point, and we're in a grey area, as anyway
* `From` is supposed to be an email address - and it's not what we provide
* I don't know if the server accepts non-ASCII usernames, beyond the http header charset

Issue #620, opened 2022-11-11
Use the new ElasticSearch endpoint to provide search as you type product suggestions


teolemon:
### What
- Use the new ElasticSearch endpoint to provide search as you type product suggestions
- @alexgarel finished openfoodfacts-search staging deployment at https://search.openfoodfacts.net/ - it now receives live updates from https://world.openfoodfacts.net/
- [ ] Endpoint documentation https://search.openfoodfacts.net/docs#/default/autocomplete_autocomplete_post
- cc @simonj2

monsieurtanuki:
@teolemon @raphael0202 We don't seem to share the same taxonomy tags, e.g. `categories` vs `category`.
I don't really mind, it's just that we would need a new specific `TaxonomyType` class. Which kind of make sense anyway as we currently don't have a "brand" taxonomy type and Elastic Search does have one AFAIK.

monsieurtanuki:
@teolemon @raphael0202 @simonj2 For the record, following are the taxonomy tags we currently use - could you provide the list of the taxonomy tags used in ElasticSearch
1. states
1. languages
1. labels
1. categories
1. countries
1. ingredients
1. traces
1. additives
1. allergens
1. packaging
1. origins
1. packaging_shapes
1. packaging_materials
1. packaging_recycling
1. nova_groups
1. emb_codes

teolemon:
That's https://github.com/openfoodfacts/openfoodfacts-dart/issues/619

PR #834, opened 2023-11-22
feat: 620 - Elastic Search autocomplete for categories

monsieurtanuki:
### What
- First use of the Elastic Search API: autocomplete for categories in TEST env.
- TODO in other PRs
   - access to PROD env (so far, only TEST env works)
   - other fuzziness levels (so far, just `none`)
   - other taxonomy names (so far, just `category`)

### Part of 
- #620

### Files
New files:
* `api_search_test.dart`: Tests around the Elastic Search API.
* `autocomplete_search_result.dart`: Result of an autocomplete search on the Elastic Search API.
* `autocomplete_search_result.g.dart`: generated
* `autocomplete_single_result.dart`: Single item result of an autocomplete search on the Elastic Search API.
* `autocomplete_single_result.g.dart`: generated
* `fuzziness_level.dart`: Fuzziness Level for Elastic Search API.
* `open_food_search_api_client.dart`: Client calls of the Open Food Facts Elastic Search API.
* `taxonomy_name.dart`: Taxonomy Name for Elastic Search API.

Impacted files:
* `openfoodfacts.dart`: exported the new 5 files.
* `uri_helper.dart`: added a `forcedHost` parameter in order to deal with alternative subdomains like `search` for Elastic Search


g123k:
Apart from your code, I was wondering whether we shouldn't create an experimental part in the SDK.
That would make it possible to include this PR, but also the one for pricing.
The idea would be to say: you can test, but at any time the API can change or disappear.

raphael0202 (approved):
One of the advantage of this API is the support for brand, is it something you plan to add as well?

raphael0202:
> could you provide a list to all (or many) "taxonomy name"s, or a way to get them?

Yes, it's available here! If you need more, it's easy to index new ones
https://github.com/openfoodfacts/search-a-licious/blob/main/data/config/openfoodfacts.yml#L132

> could you provide a list of values and meanings of fuzziness levels?
Fuziness is an int, it's the maximum allowed levenshtein distance (=number of edits): https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html#fuzziness

> any idea when the PROD env will be available?
We plan to install search-a-licious on off1, the server that was used so far to host Product Opener (the migrations went smoothly yesterday :tada:). We need to perform software and hardware upgrades, it's expected to be ready in ~3 weeks.

PR #839, opened 2023-11-25
feat: simplified access to raw product images

monsieurtanuki:
### What
- This PR gives a direct access to "raw" (=uploaded) pictures at the same time as "main" pictures (e.g. 'front_en') (that are actually computed from raw pictures).
- Those pictures will be in field `images`, that used to have only "main" pictures for historical reasons.
- With this PR we have more flexibility: instead of having to load in a second step all the "raw" images, we directly get them from the `IMAGES` product field.
- The target is to simplify in Smoothie the access to the product image gallery: the main pictures and all the other pictures, without needing an extra server access.

### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/4674

### Impacted files
* `api_get_product_image_ids_test.dart`: used new method and extended tests to all/main/raw images
* `api_get_product_test.dart`: minor refactoring
* `api_json_to_from_test.dart`: extended tests to all/main/raw images
* `image_helper.dart`: minor refactoring
* `json_helper.dart`: integrated "raw" images in addition to "main" images
* `open_food_api_client.dart`: deprecated method `getProductImageIds`
* `product.dart`: integrated "raw" images in addition to "main" images
* `product.g.dart`: generated
* `product_helper.dart`: minor refactoring
* `product_image.dart`: now can include "raw" images


lib/src/model/product.dart
```diff
@@ -196,10 +196,33 @@ class Product extends JsonObject {
...
-      toJson: JsonHelper.imagesToJson)
+      fromJson: JsonHelper.allImagesFromJson,
+      toJson: JsonHelper.allImagesToJson)
+
+  /// All images in bulk. Will include "main" images and "raw" images.
+  ///
+  /// See also [getRawImages] and [getMainImages].
   List<ProductImage>? images;
 
+  /// "Raw" (uploaded) images, like "picture 12" for "400" size.
```
M123-dev, line 207:
```suggestion
  /// "Raw" (uploaded) images, for example "picture 12" cropped to "400" size.
```

lib/src/model/product.dart
```diff
@@ -196,10 +196,33 @@ class Product extends JsonObject {
...
+  List<ProductImage>? getRawImages() => _getImageSubset(false);
+
+  /// "Main" images, like "front" picture for "French" and "400" size.
```
M123-dev, line 210:
```suggestion
  /// "Main" images: The selcted image for certain criteria
  /// For example the "front" picture in "French"
  /// Images may be returned in multiple sizes
```

M123-dev (approved):
Thanks a lot @monsieurtanuki, I like the new implementation way more without the helper class. 

I've clarified some comments as I was kinda confused by them without reading the doc comment for `ProductImage`. Feel free to have a look at them

monsieurtanuki:

I've changed the comments according to your suggestions.

PR #840, opened 2023-11-25
ci: image labeler

teolemon:
### What
- ci: image labeler


monsieurtanuki (approved):
Hi @teolemon!
LGTM

Most of the other entries seem wrong, though, as they use the `lib` folder directly instead of `lib/src`.

teolemon:
Nice catch @monsieurtanuki 👍

Issue #773, opened 2023-08-07
Add product_name_brand_quantity function


ductung-bfast:
### Why - Problem description
Mobile app has no such full name as on the web.

### What - Proposed solution
We should replicate product_name_brand_quantity function as on the web

teolemon:
- For reference https://github.com/openfoodfacts/openfoodfacts-server/blob/0234c56635292483cd397aa5893610be39f256b9/lib/ProductOpener/Products.pm#L2679
- @stephanegigandet any comments on this one ?

stephanegigandet:
Is is a good idea I think

PR #844, opened 2023-12-02
feat: 773 - new method "getProductNameBrandQuantity"

monsieurtanuki:
### What
- New method `getProductNameBrandQuantity`.
- Added needed "abbreviated_product_name" product fields.

### Fixes bug(s)
- Closes: #773

### Impacted files
* `api_get_product_test.dart`: tested "abbreviated product name" fields
* `product.dart`: added method `getProductNameBrandQuantity`; added "abbreviated product name" fields
* `product.g.dart`: generated
* `product_fields.dart`: added "abbreviated product name" fields


lib/src/model/product.dart
```diff
@@ -925,4 +943,75 @@ class Product extends JsonObject {
     }
     _nutriments = nutriments;
   }
+
+  /// Returns the best version of a product name.
+  ///
+  /// cf. openfoodfacts-server/lib/ProductOpener/Products.pm
+  String getBestProductName(final OpenFoodFactsLanguage language) {
```
g123k, line 950: I would prefer to return a `null` value if there is no match; otherwise you force checking an empty string

monsieurtanuki:
Closing as stale.

lib/src/model/product.dart
```diff
@@ -951,4 +969,75 @@ class Product extends JsonObject {
...
+    }
+    if ((tmp = genericName)?.isNotEmpty == true) {
+      return tmp!;
+    }
+    if ((tmp = abbreviatedNameInLanguages?[language])?.isNotEmpty == true) {
+      return tmp!;
+    }
+    if ((tmp = abbreviatedName)?.isNotEmpty == true) {
+      return tmp!;
+    }
```
M123-dev, line 995: Can't we do a 

tmp ??= x;
tmp ??= y;

return tmp;

M123-dev (approved):
First of all sorry for my very irregular contributions, bout to write my finals in April and Mai, guess especially this PR somehow got  lost in the notifications. 

Added a small comment, otherwise good. Especially the new bestName method could be great for outside package users

lib/src/model/product.dart
monsieurtanuki: @M123-dev Looks tempting but we check "not null and not empty", not just "not null", therefore your suggestion would potentially return empty strings.

monsieurtanuki:
Thank you @M123-dev for your review, and good luck for your finals!

PR #857, opened 2023-12-16
fix: If `product -> images` is `null`, `getProductImageIds`shouldn't crash

g123k:
Hi everyone!

As noticed in the OFF app, when in the json, `product -> images` is `null`, the code crashes because we assign it to a non-null Map.


teolemon:
Just curious, how can this be reproduced ?

g123k:
You can test with `98765432186`

monsieurtanuki (approved):

teolemon:
Thanks @g123k 🙂

PR #858, opened 2023-12-22
fix: unit test with updated expected values

monsieurtanuki:
### What
- Had to update expected values for one unit test.


M123-dev (approved):

monsieurtanuki:
Thank you @M123-dev for your quick review!

M123-dev:
The only quick one 😅

PR #859, opened 2023-12-26
chore: test TEST server first before some tests around getProductV3

monsieurtanuki:
### What
- The purpose of this PR is to get rid of annoying emails about some specific tests non passing.
- In that particular case, the TEST server is often funny in its head, and `getProductV3` (and of course the corresponding URL) systematically fails with a `502 Bad Gateway` error.
- Instead of receiving error messages about problems that do not regard our code but the current status of the TEST env, we check first if a simple `getProductV3` call works. If it triggers a `502`, we ignore the test. If it passes, we run the test. IF there's another exception, we fail the test.

### DIFF
--- a/test/api_get_save_product_test.dart
+++ b/test/api_get_save_product_test.dart
@@ -31,8 +31,40 @@ void main() {
   const String russianIngredientsAll = 'Мука, вода';
   const List<String> russianIngredientsSplit = <String>['Мука', 'Вода'];
 
+  /// Returns true if the (TEST) server is in a good mood.
+  ///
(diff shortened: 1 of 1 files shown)


M123-dev (approved):
It would be amazing to stop these (for me push notifications from the Github App), so 100% for this check 💯

monsieurtanuki:

Let's extend the same principle each time we have "irrelevant" test errors. For that we'll need a simple query on the related server.
We can do that "on demand", next time the next tests fail for no "good" reason.

PR #861, opened 2023-12-31
chore: test Oxf servers

monsieurtanuki:
### What
- The purpose of this PR is to get rid of annoying emails about some specific tests non passing.
- In that particular case, the OxF servers are sometimes funny in their heads, and getProduct (and of course the corresponding URL) fails with a 504 Gateway Time-out.
- Instead of receiving error messages about problems that do not regard our code but the current status of the OxF servers, we ignore the tests when we have 504 exceptions.

### Impacted files
* `api_not_food_get_product_test.dart`: added tests around `504 Gateway Time-out`
* `pubspec.yaml`: minor package upgrades


monsieurtanuki:
Closing as stale.

Issue #862, opened 2024-01-03
Missing nutrients (e.g. Polyols)


monsieurtanuki:
### Description
cf. @CharlesNepote's https://github.com/openfoodfacts/smooth-app/issues/4943
There are some nutrients in [Food.pm](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/lib/ProductOpener/Food.pm) and https://world.openfoodfacts.org/cgi/nutrients.pl?lc=fr&cc=fr that we don't deal with in off-dart: they are not in off-dart's `Nutrient` `enum`.

### Expected behavior
We should be able to deal with all nutrients, at least most of them.

monsieurtanuki:
@CharlesNepote There are 47 "nutrients" found in Food.pm that are currently not in off-dart.
For each nutrient, we need to attach a `Unit` (KCAL, KJ, G, MILLI_G, MICRO_G, MILLI_L, L, PERCENT).
Would you list the important nutrients you would like to see added, each with their unit?

1. alpha-linolenic-acid
1. arachidic-acid
1. arachidonic-acid
1. behenic-acid
1. beta-carotene
1. bicarbonate
1. carbon-footprint
1. carbon-footprint-from-meat-or-fish
1. casein
1. cerotic-acid
1. chlorophyl
1. cocoa
1. collagen-meat-protein-ratio
1. dihomo-gamma-linolenic-acid
1. elaidic-acid
1. energy-from-fat
1. folates
1. fructose
1. fruits-vegetables-nuts
1. fruits-vegetables-nuts-dried
1. fruits-vegetables-nuts-estimate
1. gamma-linolenic-acid
1. glucose
1. glycemic-index
1. gondoic-acid
1. insoluble-fiber
1. lactose
1. lignoceric-acid
1. maltodextrins
1. maltose
1. mead-acid
1. melissic-acid
1. montanic-acid
1. nervonic-acid
1. nucleotides
1. nutrition-score-fr
1. nutrition-score-uk
1. omega-9-fat
1. ph
1. polyols
1. serum-proteins
1. silica
1. soluble-fiber
1. starch
1. sucrose
1. taurine
1. water-hardness

CharlesNepote:
[...]

PR #863, opened 2024-01-04
fix: 862 - added 18 nutrients

monsieurtanuki:
### What
Added 18 nutrients:
- Omega 9
- ß-Carotene
- Bicarbonate
- Sugar Alcohol (polyol)
- and 14 acids

### Part of 
- #862


M123-dev (approved):
Lots of acids

monsieurtanuki:

> Lots of acids

Yeah, it looked easier to manage those guys first, with gram as the suspected unit. Compared to `nutrition-score-uk`.

PR #865, opened 2024-01-12
fix: unit test value + better json decode

monsieurtanuki:
### What
- Fixed a unit test value
- Improved the json decoding when the server is down. Something like
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>ð Service Downtime Notification - Open Food Facts ð</title>
    <style>
```
image

image

### Impacted files
* `api_get_product_test.dart`: updated expected value
* `http_helper.dart`: handled the `<!DOCTYPE html>` returned from the server when jsondecoding

### DIFF
--- a/lib/src/utils/http_helper.dart
+++ b/lib/src/utils/http_helper.dart
@@ -254,6 +254,19 @@ class HttpHelper {
         throw Exception(
             'JSON expected, software error found: ${string.split('\n')[1]}');
       }
+      if (string.startsWith('<!DOCTYPE html>')) {
+        const String titleOpen = '<title>';
+        const String titleClose = '</title>';
+        int pos1 = string.indexOf(titleOpen);
+        if (pos1 >= 0) {
+          pos1 += titleOpen.length;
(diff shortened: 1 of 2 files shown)


M123-dev (approved):
We could also integrate this into the requests. Seeing multiple issues of format exception due to html responses in sentry

monsieurtanuki:

> We could also integrate this into the requests. Seeing multiple issues of format exception due to html responses in sentry

I'm not sure what you mean and/or how we should proceed, then, as we return a `Product` or throw an exception.
If we want to always return something, we should return a "meta-product" that contains a nullable product and a nullable "exception".
What is currently implemented is that almost every request uses a custom json decode method, that handles the potential `<html>` returned. In this PR I added a case in the common custom json decode method, therefore I think we should be fine from now on.
Feel free to open an issue if I've misunderstood your use-case.

PR #869, opened 2024-01-24
fix: 862 - added 1 nutrient

davidpryor:
### What
- Added enumeration support for "added-sugars"
- Nutrient already supported server-side https://github.com/openfoodfacts/openfoodfacts-server/blob/acf96bb5625f02ea1bc3bc0bcbdb82153a896517/taxonomies/nutrients.txt#L2098

### Screenshot
image

### Part of 
- #862


davidpryor:
> Looks good to me @davidpryor! Congratulations for your first contribution!
> 
> For the record you would be better off with a specific branch instead of `master`, but that's OK for this time I guess.

Haha thank you and I agree 😅
I was just being a little lazy!

Issue #872, opened 2024-01-29
Nutriment : PerSize.serving Potentially Broken


davidpryor:
### Description
When setting Nutriments to `PerSize.serving`, `Product.servingSize`, and `Product.nutrimentDataPer` to `PerSize.serving.offTag`. The server will not save the Nutriments. When I set the `PerSize` to `oneHundredGrams`, the server will accept it and properly scale it based on my `servingSize` attribute.

### Expected behavior
SDK/backend should accept Nutriments provided in `PerSize.serving`.

### Stacktraces
None

### Package information
Open Food Facts package version: 3.3.2

### Additional context
I see there is a test that sends a new food to the backend with `PerSize.serving`, but this test sends empty nutriments, so this particular scenario is not tested with the SDK (I think 😁 )

monsieurtanuki:
Hi @davidpryor!
Please write a unit test that fails according to you.

davidpryor:
@monsieurtanuki 
I have created a test here:
https://github.com/davidpryor/openfoodfacts-dart/commit/83e96dcc54cf69eb29871cd9ea28c762129c6a88

This was a copy of an existing test with the following changes:
```diff
-      const PerSize perSize = PerSize.oneHundredGrams;
+      const PerSize perSize = PerSize.serving;
```
```diff
-      final String nutrimentDataPer = PerSize.oneHundredGrams.offTag;
+      final String nutrimentDataPer = PerSize.serving.offTag;
```
```diff
        final Product newProduct = Product(
          barcode: BARCODE,
          productName: PRODUCT_NAME,
          nutrimentDataPer: nutrimentDataPer,
+          servingSize: "20g",
+          servingQuantity: 20,
          nutriments: nutriments,
        );
```
These fields on Product were communicated to me by Alex on slack https://openfoodfacts.slack.com/archives/C032PK9D8/p1706546238737309?thread_ts=1706301302.772189&cid=C032PK9D8


Here is the first incorrect expect check:
<img width="499" alt="image" src="https://github.com/openfoodfacts/openfoodfacts-dart/assets/79552751/acfde590-2e30-4aef-88b3-fd391ff2a5a4">

davidpryor:
I think I have discovered what is happening.
@stephanegigandet has clarified `nutriments_(nutrient)_100g` and `nutriments_(nutrient)_serving` are both read only and the server only cares about `nutrition_data_per` for writes.

[...]

PR #873, opened 2024-01-30
fix: failing nutriment persize serving test

davidpryor:
### What
Fixed inability to send nutrients to server that were tagged with PerSize.serving

### Fixes bug(s)
#872


test/api_save_product_test.dart
```diff
@@ -398,6 +398,82 @@ Like that:
...
+
+        expect(result.status, ProductResultV3.statusSuccess);
+        expect(result.barcode, BARCODE);
+        final Product? searchedProduct = result.product;
+        expect(searchedProduct != null, true);
+        if (searchedProduct != null) {
+          expect(searchedProduct.barcode, BARCODE);
+          expect(searchedProduct.productName, PRODUCT_NAME);
+          expect(searchedProduct.nutrimentDataPer, nutrimentDataPer);
+          var searchedNutriments = searchedProduct.nutriments;
```
monsieurtanuki, line 455:
```suggestion
          final searchedNutriments = searchedProduct.nutriments;
```

test/api_save_product_test.dart
```diff
@@ -398,6 +398,82 @@ Like that:
...
+            version: ProductQueryVersion.v3,
+          ),
+          user: USER,
+          uriHelper: uriHelper,
+        );
+
+        expect(result.status, ProductResultV3.statusSuccess);
+        expect(result.barcode, BARCODE);
+        final Product? searchedProduct = result.product;
+        expect(searchedProduct != null, true);
```
monsieurtanuki, line 450:
```suggestion
        expect(searchedProduct, isNotNull);
```

test/api_save_product_test.dart
```diff
@@ -398,6 +398,82 @@ Like that:
...
+        if (searchedProduct != null) {
```
monsieurtanuki, line 451: Please get rid of useless `if`s.

test/api_save_product_test.dart
```diff
@@ -398,6 +398,82 @@ Like that:
...
+          expect(searchedProduct.barcode, BARCODE);
+          expect(searchedProduct.productName, PRODUCT_NAME);
+          expect(searchedProduct.nutrimentDataPer, nutrimentDataPer);
+          var searchedNutriments = searchedProduct.nutriments;
+          expect(searchedNutriments, isNotNull);
+          if (searchedNutriments != null) {
```
monsieurtanuki, line 457: Please get rid of useless `if`s.

test/api_save_product_test.dart
```diff
@@ -398,6 +398,82 @@ Like that:
...
+            final List<Nutrient> nutrients = <Nutrient>[
+              Nutrient.energyKJ,
+              Nutrient.carbohydrates,
+              Nutrient.proteins,
+              Nutrient.fat,
+              Nutrient.vitaminB12,
```
monsieurtanuki, line 463: You probably should write your code a bit differently, as there's no vitamin b12 in your data. I guess an initial `Map<String, double>` would help you make sure you're testing data that you saved earlier.
Additional idea: add an `isNotNull` test for the values (which would have crashed for vitamin b12).

test/api_save_product_test.dart
```diff
@@ -398,6 +398,82 @@ Like that:
...
+      const double ENERGY = 365;
+      const double CARBOHYDRATES = 12;
+      const double PROTEINS = 6;
+      const double FAT = 0.1;
+      const String BARCODE = '01234567891';
+      const String PRODUCT_NAME = 'Test Food';
+      final String nutrimentDataPer = PerSize.serving.offTag;
+
+      const PerSize perSize = PerSize.serving;
+      for (int i = 2; i >= 0; i--) {
```
monsieurtanuki, line 412: I know you probably just copied that 2 down to 0 loop elsewhere, but that's very strange.
Please test if something less exotic like 0 to 2 works: if so fix this line and the other line in this file with the 2 to 0.

test/api_save_product_test.dart
```diff
@@ -398,6 +398,82 @@ Like that:
       }
     }, skip: 'Works randomly');
 
+    test('Test saving food with PerSize.serving Nutrients', () async {
```
monsieurtanuki, line 401: Instead of copying/pasting another test, I guess you could loop on `PerSize.values`.

lib/src/open_food_api_client.dart
```diff
@@ -109,6 +109,10 @@ class OpenFoodAPIClient {
         if (pos != -1) {
           key = key.substring(0, pos);
         }
+        final servingPost = key.indexOf('_serving');
+        if (servingPost != -1) {
+          key = key.substring(0, servingPost);
+        }
```
monsieurtanuki, line 115: It would be way much cleaner to loop on `PerSize.values` rather than testing on `_100g` and `_serving` that come out of nowhere.

monsieurtanuki (commented):
Thank you @davidpryor for finding and fixing that bug!
I have comments regarding the code itself though, regarding maintainability.
Please have a look at them and feel free to tell me what you think.

lib/src/open_food_api_client.dart
davidpryor: Agreed!

test/api_save_product_test.dart
davidpryor: Yeah I copied the whole test. I will rewrite it to be more applicable.

test/api_save_product_test.dart
davidpryor: I went ahead and combined the test with the test I copied from

test/api_save_product_test.dart
davidpryor: to remove the useless "if" statements, I changed this line to a null check with a `fail` call inside. This is to avoid the use of `!.`

test/api_save_product_test.dart
davidpryor: Flipped that for you!

test/api_save_product_test.dart
monsieurtanuki: > This is to avoid the use of !.

Actually we don't need to "avoid" `!`.
Well, tell me when your code is ready for a new review.

test/api_save_product_test.dart
```diff
@@ -329,23 +329,112 @@ Like that:
...
+        Nutrient.vitaminB12,
+      ];
+
       const String BARCODE = '7340011364184';
       const String PRODUCT_NAME = 'Chili beans';
-      final String nutrimentDataPer = PerSize.oneHundredGrams.offTag;
+      for (final option in PerSize.values) {
+        final String nutrimentDataPer = option.offTag;
+
+        PerSize perSize = option;
```
monsieurtanuki, line 345:
```suggestion
      for (final PerSize perSize in PerSize.values) {
        final String nutrimentDataPer = perSize.offTag;
```

test/api_save_product_test.dart
```diff
@@ -329,23 +329,112 @@ Like that:
...
+            expect(
+              actual,
+              isNotNull,
+              reason: 'should not be null for $nutrient',
+            );
+            expect(
+              searchedNutriments.getValue(nutrient, perSize),
+              nutriments.getValue(nutrient, perSize),
+              reason: 'should be the same values for $nutrient',
+            );
```
monsieurtanuki, line 410:
```suggestion
            final double? actual =
                searchedNutriments.getValue(nutrient, perSize);
            expect(
              actual,
              isNotNull,
              reason: 'should not be null for $nutrient',
            );
            expect(
              actual,
              nutriments.getValue(nutrient, perSize),
              reason: 'should be the same values for $nutrient',
            );
```

test/api_save_product_test.dart
```diff
@@ -329,23 +329,112 @@ Like that:
...
+    test('Test saving food with PerSize.serving Nutrients', () async {
+      const User USER = TestConstants.TEST_USER;
+      const double ENERGY = 365;
+      const double CARBOHYDRATES = 12;
+      const double PROTEINS = 6;
+      const double FAT = 0.1;
+      const String BARCODE = '01234567891';
+      const String PRODUCT_NAME = 'Test Food';
+
+      for (final option in PerSize.values) {
```
monsieurtanuki, line 425: Bad naming: "option". What about "var1", "possibleValue"?
Jokes aside, `perSize` would be easier to read, wouldn't it?

monsieurtanuki (commented):
Thank you @davidpryor for your changes!

I cannot see the added value of your additional "Test saving food with PerSize.serving Nutrients" test. Now that you loop in the already existing "confirm that nutrient fields are saved" test for all `PerSize` values, we don't need to copy/paste the test, right?
Explain that to me or fix that, and I'll approve your PR.

Less important: if you used a const map for initial values, you would never have potential discrepancies between the nutrient values you set and the list of nutrients you check:
```dart
const Map<Nutrient, double> rootNutrientValues = <Nutrient, double>{
  Nutrient.energyKJ: 365,
  Nutrient.carbohydrates: 12,
  Nutrient.proteins: 6,
  Nutrient.fat: 0.1,
};
```
E.g. using `rootNutrientValues.keys`.

davidpryor:
> I cannot see the added value of your additional "Test saving food with PerSize.serving Nutrients" test. Now that you loop in the already existing "confirm that nutrient fields are saved" test for all `PerSize` values, we don't need to copy/paste the test, right?
> Explain that to me or fix that, and I'll approve your PR.

I agree. Just missed deleting it.

monsieurtanuki (approved):
Thank you @davidpryor for your PR, looks good to me!

Issue #831, opened 2023-11-21
Adding support for the Prices project


teolemon:
### Description
Add support for the Prices project
https://prices.openfoodfacts.org/api/docs 
Points of contact: @raphodn @raphael0202 
Demonstrator by @monsieurtanuki : https://github.com/monsieurtanuki/openprices

monsieurtanuki:
> Adding support for the Prices project https://price.openfoodfacts.org/docs

@teolemon The link doesn't work.

teolemon:
I just updated it, we swapped the url this morning, since we're going to have many prices

monsieurtanuki:
@teolemon @raphodn @raphael0202 Btw is there already a project of app dedicated to prices, with
* an easy access to maps and OSM location ids
* a barcode scanner
* price input

raphodn:
Hey @monsieurtanuki !
You can find details about the project here : https://wiki.openfoodfacts.org/Project:Open-Prices
There's link to the repos & demos there :)

raphael0202:
> Btw is there already a project of app dedicated to prices, with

@monsieurtanuki That's interesting, can you give us the link? :)

monsieurtanuki:
@raphael0202 "is there?" (question), and not "there is" (affirmation)...

That said, I"d be glad to start a POC.
Assuming that we already have users with a relatively clear idea about their needs (e.g. "price" session in a specific shop).

raphael0202:
> "is there?" (question), and not "there is" (affirmation)...

Oh my bad, without a question mark at the end I misread it. Yes there is a JS POC available here: https://prices.openfoodfacts.net/app!

It's still a work in progress, but we would definitely need a real app to do this, as web barcode scanners are far from working correctly.

monsieurtanuki:
I guess we can close now, and create issues when relevant.

PR #884, opened 2024-02-14
feat: 831 - new price methods like add price and upload proof

monsieurtanuki:
### What
Now we're getting serious about prices with the following new methods:
* `getAuthenticationToken` to start a session with authentication
* `getUserSession` to check if a session is running
* `deleteUserSession` to close a session
* `createPrice` to add a price
* `deletePrice` to delete a price
* `uploadProof` to upload a proof
* `deleteProof` to delete a proof

### Part of 
- #831

### Files
New files:
* `maybe_error.dart`: Contains a successful value OR an error.
* `price_per.dart`: Price Per.
* `proof.dart`: Proof of a price.
* `proof.g.dart`: generated
* `proof_type.dart`: Type of a Proof.
* `session.dart`: Price session.
* `session.g.dart`: generated

Impacted files:
* `api_get_save_product_test.dart`: unrelated minor fix
* `api_prices_test.dart`: added tests for the new methods
* `http_helper.dart`: added optional "bearerToken" parameter to GET; new `doDeleteRequest`, `doPostJsonRequest` and `imagineMediaType` methods
* `open_prices_api_client.dart`: added parameters to `getPrices`; new methods `getAuthenticationToken`, `getUserSession`, `deleteUserSession`, `createPrice`, `deletePrice`, `uploadProof` and `deleteProof`; minor refactoring
* `openfoodfacts.dart`: added 5 new files to export
* `price.dart`: added fields.
* `price.g.dart`: generated
* `pubspec.yaml`: added package `http_parser` for `MediaType`
* `test_constants.dart`: unrelated minor fix


lib/src/open_prices_api_client.dart
```diff
@@ -98,10 +145,234 @@ class OpenPricesAPIClient {
...
+      return MaybeError<Proof>.value(Proof.fromJson(json));
+    }
+    return MaybeError<Proof>.error(
+      error: responseBody,
+      statusCode: response.statusCode,
+    );
+  }
+
+  /// Deletes a proof.
+  ///
```
M123-dev, line 357:
```suggestion
  /// A user can delete only owned proofs. Can delete only proofs that are not associated with prices. A moderator can delete not owned proofs.
```

lib/src/utils/http_helper.dart
```diff
@@ -198,14 +231,30 @@ class HttpHelper {
...
   /// Note: [addCredentialsToHeader] and [isTestModeActive] exclude each other.
   Map<String, String>? _buildHeaders({
     User? user,
     required final UriHelper uriHelper,
     required bool addCredentialsToHeader,
+    final String? bearerToken,
   }) {
+    if (bearerToken != null) {
+      return _getBearerHeader(bearerToken);
+    }
```
M123-dev, line 256: Don't we want a user agent and a "From" header. Just because prices does not use them yet it could be interesting

M123-dev (approved):
Thanks a lot @monsieurtanuki, added some minor comments 👍🏼

lib/src/open_prices_api_client.dart
```diff
@@ -1,35 +1,81 @@
...
 
 /// Client calls of the Open Prices API.
+///
+/// cf. https://prices.openfoodfacts.org/api/docs
 class OpenPricesAPIClient {
   OpenPricesAPIClient._();
 
-  /// Subdomain of the Elastic Search API.
+  /// Status when the server is running (cf. [getStatus]).
+  static const String statusRunning = 'running';
```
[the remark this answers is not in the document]
monsieurtanuki: @M123-dev Not sure it would be appropriate: basically we just want to know if the server is "running" (I even hesitated returning a `bool` instead).
Using an `enum` would be problematic: what if there's a new status we don't know? We should then maintain the values. For no added values. Let's be resilient.

lib/src/open_prices_api_client.dart
```diff
@@ -98,10 +145,234 @@ class OpenPricesAPIClient {
...
+        final dynamic decodedResponse = HttpHelper().jsonDecodeUtf8(response);
+        return MaybeError<String>.value(decodedResponse['status']!);
+      } catch (e) {
+        //
+      }
+    }
+    return MaybeError<String>.responseError(response);
+  }
+
+  /// Authentication: provide username/password and get a bearer token in return.
```
monsieurtanuki, line 159:
```suggestion
  /// Authentication: provide username/password and get a bearer token in return.
  ///
  /// [username]: Open Food Facts user_id (not email)
  /// [password]: user password (clear text, but HTTPS encrypted)
  /// A token is returned. If the set_cookie parameter is set to 1, the token is also set as a cookie named "session" in the response.
  /// To authenticate, you can either:
  /// - use the Authorization header with the Bearer scheme, e.g.: "Authorization: bearer token"
  /// - use the session cookie, e.g.: "Cookie: session=token"
```

lib/src/utils/http_helper.dart
monsieurtanuki: > Don't we want a user agent and a "From" header. Just because prices does not use them yet it could be interesting

Could be done later if needed, for stats I guess.
I don't know if `prices` plan to do something with it.
For the moment the code is clearer to read that way: "token? just return this! else do as usual".

monsieurtanuki:
Thank you very much @M123-dev for your quick (but detailed) review!

Regarding verbose comments, I don't know the best practice: is it relevant to copy the API comments here, as they can be outdated? Is it relevant to just point to the API docs?

raphael0202:
Great :tada:

monsieurtanuki:
@raphael0202 Today I'll PR a bigger `getPrices`, which will give us a feature we need in Smoothie: get the latest prices for a product (order by created desc, or created desc)

I'm still a bit stuck with the "get location" feature, that we need too:
* is there an easy way (OSM or prices) to get the OSM shops with a query like "monoprix, avenue daumesnil, paris"?
* is there an easy way (OSM or prices) to get the OSM shops close to a geopoint?

Perhaps we would need the prices for a product within 5km.

raphael0202:
> is there an easy way (OSM or prices) to get the OSM shops with a query like "monoprix, avenue daumesnil, paris"?

Currently we use OSM Nominatim, which does the job (but is far from perfect). Can you use it as well here?

> is there an easy way (OSM or prices) to get the OSM shops close to a geopoint?

Hum I'm not sure of this, maybe @raphodn will know better

PR #886, opened 2024-02-17
feat: 831 - more parameters for the "get prices" API query

monsieurtanuki:
### What
- More refined parameters for the "get prices" API query.
- Including the "order by" clause, useful for Smoothie (get the latest prices for that product).

### Part of 
- #831

### Files
New files:
* `get_parameters_helper.dart`: Helper class for API query parameters.
* `get_prices_order.dart`: Order clause for "get prices".
* `get_prices_parameters.dart`: Parameters for the "get prices" API query.

Impacted files:
* `api_prices_test.dart`: elaborated the test of "get prices"
* `open_prices_api_client.dart`: refactored
* `openfoodfacts.dart`: 2 new exported files

### DIFF
--- a/lib/openfoodfacts.dart
+++ b/lib/openfoodfacts.dart
@@ -89,6 +89,9 @@ export 'src/personalized_search/preference_importance.dart';
 export 'src/personalized_search/product_preferences_manager.dart';
 export 'src/personalized_search/product_preferences_selection.dart';
 export 'src/prices/currency.dart';
+// export 'src/prices/get_parameters_helper.dart'; // uncomment if really needed
+export 'src/prices/get_prices_order.dart';
+export 'src/prices/get_prices_parameters.dart';
 export 'src/prices/get_prices_result.dart';
 export 'src/prices/get_prices_results.dart';
 export 'src/prices/location.dart';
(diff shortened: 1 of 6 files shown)


teolemon (approved):

monsieurtanuki:

I think we're all set now for the MVP regarding prices in Smoothie:
* create/get/delete session
* create/get/delete price
* upload/delete proof

Still not implemented:
* get users (not really helpful in a first approach)
* get proof (but it's somehow embedded in get price)
* get product (but we can do that in "standard" off-dart)
* get locations (but we can use OSM methods and it's somehow embedded in get price)

monsieurtanuki:
@raphael0202 @raphodn In a first approach I can retrieve my "monoprix daumesnil paris" shop with a nominatim query like https://nominatim.openstreetmap.org/search.php?q=monoprix+daumesnil+paris&format=jsonv2

I can find the same shop with https://prices.openfoodfacts.org/app/add/single

But I cannot get that shop with prices API [`api/v1/locations`](https://prices.openfoodfacts.org/api/docs#/Locations/get_locations_api_v1_locations_get) (no results).
Am i doing something wrong with `api/v1/locations`?

Issue #891, opened 2024-03-12
Add field "uploaded timestamp" for "raw" images


monsieurtanuki:
### Why - Problem description
cf. #104 https://github.com/openfoodfacts/openfoodfacts-server/issues/9226 https://github.com/openfoodfacts/smooth-app/issues/4674
We need to know the uploaded timestamp of a raw image if we want to prevent the users from using it - or at least flag them - when outdated (e.g. 3 years old).

### What - Proposed solution
There's already a `uploaded_t` field in the json we get, we just currently don't store it in `ProductImage.raw`.

PR #892, opened 2024-03-13
feat: 891 - new "uploaded timestamp" field for raw images

monsieurtanuki:
### What
- Now we use the "uploaded timestamp" field that the server provided for raw images, but that we just didn't retrieve.
- With this field we'll be able to say if an uploaded image is "too old", e.g. if it was uploaded more than a year ago.

### Fixes bug(s)
- #891

### Part of 
- https://github.com/openfoodfacts/openfoodfacts-server/issues/9226
- https://github.com/openfoodfacts/smooth-app/issues/4674

### Impacted files:
* `api_json_to_from_test.dart`: added tests around new field `uploaded`
* `json_helper.dart`: added the `uploaded` field to from/to json conversion methods; refactored
* `product_image.dart`: added field `uploaded`, for uploaded images only

### DIFF
--- a/lib/src/model/product_image.dart
+++ b/lib/src/model/product_image.dart
@@ -130,6 +130,7 @@ class ProductImage {
     required String this.imgid,
     this.width,
     this.height,
+    this.uploaded,
   })  : language = null,
         field = null;
 
@@ -137,6 +138,9 @@ class ProductImage {
   final ImageSize? size;
   final OpenFoodFactsLanguage? language;
   String? url;
+
+  /// Upload timestamp, for uploaded images only, in seconds since Unix Epoch.
+  int? uploaded;
 
   /// Revision number
   int? rev;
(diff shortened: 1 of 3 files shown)


lib/src/model/product_image.dart
```diff
@@ -130,6 +130,7 @@ class ProductImage {
     required String this.imgid,
     this.width,
     this.height,
+    this.uploaded,
```
AshAman999, line 133: I think we can have a better variable name, maybe something like uploaded_on or created_at. Or is it the same var name in other places?

monsieurtanuki: Hi @AshAman999!

In this PR we have an `'uploaded_t'` JSON tag and an `int? uploaded` field.

If we compare to similar fields:
* we also have a `'created_t'` JSON tag and a `created` field (cf. in `Product`)
* but instead of `int?` in this PR, we do have a much more explicit `DateTime?`

Would you be OK with `DateTime? uploaded`? That would be consistent with the rest of the timestamp fields in the package, and wouldn't be confusing with a `DateTime?` type.

teolemon (approved):
per @AshAman999 's comment

monsieurtanuki:
Thank you @teolemon @AshAman999 for your comments!
I don't merge right now as `DateTime? uploaded` was indeed a good idea instead of `int?`. I'll implement that by tomorrow.

monsieurtanuki:
Thank you @teolemon @AshAman999: I've just merged the code with `DateTime? uploaded` field.

PR #894, opened 2024-03-19
test: Lang field read only

M123-dev:
### What

Related to: https://github.com/openfoodfacts/smooth-app/issues/4771#issuecomment-2004500769

```
Start lang: OpenFoodFactsLanguage.FRENCH

End lang: OpenFoodFactsLanguage.FRENCH

00:18 +0 -1: OpenFoodAPIClient Save product main language Save product main language [E]

  Expected: OpenFoodFactsLanguage:<OpenFoodFactsLanguage.GERMAN>
    Actual: OpenFoodFactsLanguage:<OpenFoodFactsLanguage.FRENCH>

00:18 +0 -1: Some tests failed.

```

It looks like we have a good old read only field

### DIFF
--- a/test/api_add_product_main_language.dart
+++ b/test/api_add_product_main_language.dart
@@ -1,0 +1,51 @@
+import 'dart:math';
+
+import 'package:openfoodfacts/openfoodfacts.dart';
+import 'package:test/test.dart';
+
+import 'test_constants.dart';
+
+void main() {
+  OpenFoodAPIConfiguration.userAgent = TestConstants.TEST_USER_AGENT;
+  const UriProductHelper uriHelper = uriHelperFoodTest;
+
+  group('$OpenFoodAPIClient Save product main language', () {
+    String barcode_1 = '3017620425035';
+
+    test('Save product main language', () async {
+      ProductResultV3 result =
+          await OpenFoodAPIClient.getProductV3(ProductQueryConfiguration(
+        barcode_1,
+        version: ProductQueryVersion.v3,
+      ));
+
+      Product? product = result.product;
+
+      expect(product, isNotNull);
+
+      print('Start lang: ${product!.lang}');
+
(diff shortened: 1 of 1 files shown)


monsieurtanuki:
@M123-dev Thank you for your test!

That said:
* it's MUCH better if you save data and get data from the same server, consistency-wise
```dart
      ProductResultV3 result = await OpenFoodAPIClient.getProductV3(
        ProductQueryConfiguration(
          barcode_1,
          version: ProductQueryVersion.v3,
        ),
        uriHelper: uriHelper, // was missing in your code
      );
```
* it's also a good practice to check return codes, e.g. for the saveProduct call
* that said, you're right!
* And wrong. For "your" barcode (`3017620425035`) and the first barcode I tried (`3350031897306`) the language didn't change and cause a test failure. But for another one (`8027498025058`) the test passed.

@stephanegigandet I already experienced some madness tests with fields that couldn't be changed for some products. I guess some fields are locked on some products, like "the producer says this is the right value and it's irrelevant to change it". Is there a way we can detect which fields are locked for which products?

monsieurtanuki:
@M123-dev Thank you for your PR!

`lang` isn't read-only.
It's just that it's sometimes protected as "owner field":
* `lang` is an owner field for [3017620425035](https://fr.openfoodfacts.org/api/v3/product/3017620425035?fields=owner_fields), and is read-only
* `lang` is an owner field for [3350031897306](https://fr.openfoodfacts.org/api/v3/product/3350031897306?fields=owner_fields), and is read-only
* `lang` is NOT an owner field for [8027498025058](https://fr.openfoodfacts.org/api/v3/product/8027498025058?fields=owner_fields), and is NOT read-only

Closing the PR, that's somehow a redundancy of a test of "owner field" that already exists in `api_get_product_test.dart`.

Issue #895, opened 2024-03-23
Add "expiration_date" product field


monsieurtanuki:
### Why - Problem description
cf. https://github.com/openfoodfacts/smooth-app/issues/5127

PR #898, opened 2024-03-30
feat: 895 - added product field `'expiration_date'`

monsieurtanuki:
### What
- New product field `'expiration_date'`

### Fixes bug(s)
- Fixes: #895

### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/5127

### Impacted files
* `api_get_product_test.dart`: added test for new product field about `'expiration_date'`
* `product.dart`: added product field for `'expiration_date'`
* `product.g.dart`: generated
* `product_fields.dart`: added product field for `'expiration_date'`

### DIFF
--- a/lib/src/model/product.dart
+++ b/lib/src/model/product.dart
@@ -550,6 +550,10 @@ class Product extends JsonObject {
   )
   bool? obsolete;
 
+  /// Expiration date / best before. Just a string, no format control.
+  @JsonKey(name: 'expiration_date', includeIfNull: false)
+  String? expirationDate;
+
   Product(
       {this.barcode,
       this.productName,
--- a/lib/src/model/product.g.dart
+++ b/lib/src/model/product.g.dart
@@ -165,7 +165,8 @@ Product _$ProductFromJson(Map<String, dynamic> json) => Product(
       ..origins = json['origins'] as String?
       ..novaGroup = json['nova_group'] as int?
       ..website = json['link'] as String?
-      ..obsolete = JsonHelper.checkboxFromJSON(json['obsolete']);
+      ..obsolete = JsonHelper.checkboxFromJSON(json['obsolete'])
+      ..expirationDate = json['expiration_date'] as String?;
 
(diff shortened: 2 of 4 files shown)


monsieurtanuki:
@teolemon You may be right, but frankly I believe you got mistaken with another PR merged this week: #892, about "uploaded timestamp" for raw images.
The way I interpret this "expiration date" field, if I remember what it looks like on the website, is "just in case you happen to know the expiration date, let us know, so we have a vaguely better idea of how recently we found this product in a shop". Could be dealt with much better by "prices" btw.
The sources of this PR are https://github.com/openfoodfacts/smooth-app/issues/5127 and @jusdekiwi's https://github.com/openfoodfacts/smooth-app/issues/5125

monsieurtanuki:
@teolemon Thank you for the approval btw. We can still add comments about this field, in dart and in Smoothie.

teolemon:
Alizarine explicitely told me she was using it for that purpose (the user feedback from last week) @monsieurtanuki

monsieurtanuki:
> Alizarine explicitely told me she was using it for that purpose (the user feedback from last week) @monsieurtanuki

"contributors estimate how old the photos are" with an unrelated "expiration date" field that has a unique value for the whole product, like `"02/04/2017"` for [`3661344961326`](https://fr.openfoodfacts.org/api/v3/product/3661344961326?fields=expiration_date)?
We have an "uploaded timestamp" for that on each uploaded photo.
Something I don't understand. That wouldn't be the first time, actually :)

rugk:
Coming here from a very related discussion (aka https://github.com/Forceu/barcodebuddy/issues/125)… Barcodebuddy allows to use OpenFoodFacts data to fill in the Grocy DB (kind of an ERP-management tool for your fridge etc.) by requesting it etc.

Now the missing one big piece is actually the expiration date / best before date (BFD). If that would be inside the OpenFoodFacts data or at least some generic estimate on how it is usually (since the time you buy a product), that would make it quite easy.

PR #902, opened 2024-04-09
chore: unit test fixes

monsieurtanuki:
### What
- Unit tests fixes regarding the latest failing tests.
- Still not working: the "emb" suggestions, but that's a problem on the server side.


monsieurtanuki:
Somehow dealt with by #926

Issue #893, opened 2024-03-14
Add support for the upcoming user info API


teolemon:
### Description
Add support for the upcoming user info API
https://github.com/openfoodfacts/openfoodfacts-server/pull/9918

stephanegigandet:
Note that we are planning to remove the email from the body. Is it needed for the app? cc @monsieurtanuki @g123k

monsieurtanuki:
> Note that we are planning to remove the email from the body. Is it needed for the app? cc @monsieurtanuki @g123k

@stephanegigandet In Smoothie we don't need the user email field returned by `/cgi/auth.pl`:
* we only use `/cgi/auth.pl` once, in off-dart's method `login2`
* we only use method `login2` once, in Smoothie's method `getLoginResult`
* inside method `getLoginResult`, we only extract the `user_id` field

In Smoothie the user may log in with the user id or the email (same password).

Actually, from https://github.com/openfoodfacts/smooth-app/issues/2464 https://github.com/openfoodfacts/smooth-app/pull/3499 we explicitly use the returned `user_id` field instead of whatever the user typed in (user id or email) because it created side-effects to use user email.

Possible confusion: is the returned `user_id` always a "decent" user id - and never an email?

stephanegigandet:
@monsieurtanuki Perfect, thanks for checking, I will remove the email field.
The returned user_id is always the user id, never an email

PR #903, opened 2024-04-09
feat: 893 - 4 new fields for user

monsieurtanuki:
### What
- Now when we log in we get 4 fields:
  - preferredLanguage
  - country
  - isModerator
  - isAdmin
- And we don't get the "email" field anymore.
- @stephanegigandet Still a problem on the server side: we get `country` as `en:france`, which is not very easy to convert, whereas `fr` would be perfect

### Fixes bug(s)
- Closes: #893

### Impacted files
* `login_status.dart`: added 4 fields and deprecated "email"
* `open_food_api_client.dart`: upgraded the max length of user name (60) and user id (40)
* `user_management_test_prod.dart`: tested the new 4 login status fields
* `user_management_test_test_env.dart`: tested the new 4 login status fields and remove a test on "email"


stephanegigandet:
@monsieurtanuki PR to add a cc field: https://github.com/openfoodfacts/openfoodfacts-server/pull/10152 ("world" can also be a value)

monsieurtanuki:
@stephanegigandet I've just PR'ed the use of json field `'cc'` instead of `'country'`.

As mentioned in https://github.com/openfoodfacts/openfoodfacts-server/pull/10152#issuecomment-2068091511, so far I couldn't test it successfully in PROD as the field doesn't seem to be available.

Besides, I'm not 100% sure it would be very clean to add a fake country called `'world'`.
Wouldn't `null` instead of `'world'` still work? For the record that is what is happening right now, as we cannot decode `'world`' as a country and decode it as `null`.

monsieurtanuki:
Thank you @stephanegigandet: I've just tested in PROD and the `cc` field is now populated.

monsieurtanuki:
ping

M123-dev (approved):
The dart side is looking fine 💯

monsieurtanuki:

Issue #890, opened 2024-03-11
Get a localized list of countries


monsieurtanuki:
### Why - Problem description
cf. https://github.com/openfoodfacts/smooth-app/issues/4921
In Smoothie we use a [package](https://pub.dev/packages/iso_countries) in order to localize the country names.
This package is only available for iOS and Android (not linux, macOS, web, windows).
And on the server we can retrieve localized countries, with https://world.openfoodfacts.org/cgi/countries.pl?lc=de

### What - Proposed solution
We should be able to use the server to localize country names.
Something like that:
```dart
  static Future<Map<OpenFoodFactsCountry, String>> getLocalizedCountries(
    final OpenFoodFactsLanguage language, {
    final UriProductHelper uriHelper = uriHelperFoodProd,
  }) async { //...
```

### Additional context
That could also be a good opportunity to check if some countries are missing in `OpenFoodFactsCountry.values`.

shubhsardana29:
Are there any specific considerations or requirements regarding how the localized country names should be handled in cases where there might be discrepancies or inconsistencies between the data provided by the server and the expected country names within the application's context?

monsieurtanuki:
@shubhsardana29 We just want to return a map of OpenFoodFactsCountry and String. We have no expectations regarding the localizations.
In the process, we may find unexpected countries: just report them and we'll see later what to do with them (e.g. we may not need Yugoslavia in 2024).

monsieurtanuki:
@shubhsardana29 Feel like starting to code? No hard sell.

monsieurtanuki:
@shubhsardana29 Too late ;)

PR #904, opened 2024-04-11
feat: 890 - new method getLocalizedCountryNames

monsieurtanuki:
### What
- New method `getLocalizedCountryNames`, that takes localized names of countries from the OFF server.
- Helpful specifically outside android and iOS - in both platforms country names can be taken from the device.
- Probably helpful for android and iOS too - in order to have more control on country names and to avoid weird cases like "Soviet Union" (cf. https://github.com/openfoodfacts/smooth-app/issues/4921)

### Fixes bug(s)
- Closes: #890

### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/4921

### Impacted files
* `api_get_product_test.dart`: test for new method `getLocalizedCountryNames`
* `open_food_api_client.dart`: new method `getLocalizedCountryNames`


M123-dev (approved):
Looks good, thanks 👍🏼

PR #911, opened 2024-04-19
feat: Allow to receive a cookie and send it later

g123k:
Hi everyone!

Some requests require to be authenticated via a cookie.
That's why, this PR allows to get it during the login and send it later


teolemon (approved):
Very much needed for Nutri-Score V2 previews. Thanks @g123k

Issue #922, opened 2024-05-09
Currency for country table


monsieurtanuki:
### Why - Problem description
cf. https://github.com/openfoodfacts/smooth-app/issues/5201: "Prepopulate currency based on country"

PR #923, opened 2024-05-09
feat: 922 - default currency for each country

monsieurtanuki:
### What
- Now each country has a default currency. Except Antarctica.

### Fixes bug(s)
- Fixes: #922

### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/5201

### Impacted files
* `api_off_tag_test.dart`: added a test about the relationship between country and currency
* `country_helper.dart`: added a `Currency` field to `OpenFoodFactsCountry`
* `currency.dart`: added fields in order to detect the most important currencies; added comments.

### DIFF
--- a/lib/src/prices/currency.dart
+++ b/lib/src/prices/currency.dart
@@ -4,311 +4,927 @@
 /// See https://en.wikipedia.org/wiki/ISO_4217 for a list of valid currency
 /// codes.
 enum Currency {
-  ADP,
-  AED,
-  AFA,
-  AFN,
-  ALK,
-  ALL,
-  AMD,
-  ANG,
-  AOA,
-  AOK,
-  AON,
-  AOR,
-  ARA,
-  ARL,
-  ARM,
-  ARP,
-  ARS,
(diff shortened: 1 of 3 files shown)


teolemon (approved):
Thank you so much @monsieurtanuki . Don't we have a satisfying route for this in the Open Prices API ? cc @raphodn

monsieurtanuki:

> Don't we have a satisfying route for this in the Open Prices API ? cc @raphodn

I'm not fluent in `. vue`, perhaps it does exist. Anyway I would spare a server call for that, in a first approach.

Meanwhile, I noticed currencies (taken from https://prices.openfoodfacts.org/api/docs / `CurrencyEnum`) that were probably not relevant:
* historical currencies like `MCF` "Franc monégasque"
* not ISO and unknown currencies like `ARM`
* currencies not useful for "prices" like `XBD` "European Unit of Account 17 (E.U.A.-17) (bond market unit)"

Therefore this PR was an opportunity to detect suspicious currencies.

Issue #933, opened 2024-05-20
Prices/Location: add countrycode field.


monsieurtanuki:
### Why - Problem description
Prices / Location now provides a "countrycode" field, and this field is helpful for us, so let's integrate it!
https://github.com/openfoodfacts/open-prices/pull/296

raphodn:
fyi the field name is `osm_address_country_code`

And there's also 2 new extra fields : `osm_tag_key` & `osm_tag_value` : https://github.com/openfoodfacts/open-prices/pull/294

PR #934, opened 2024-05-22
feat: 933 - added osm location fields (countryCode, osmKey, osmValue)

monsieurtanuki:
### What
- Added osm location fields `countryCode`, `osmKey` and `osmValue`
- Typical use cases for Smoothie:
  - displaying the flag of the country for prices
  - displaying an icon related to a location type (like, if it's a shop)

### Fixes bug(s)
- Closes: #933

### Impacted files
* `api_get_product_test.dart`: unrelated minor test fix
* `api_get_robotoff_test.dart`: unrelated minor test fix
* `api_matched_product_v1_test.dart`: unrelated minor test fix
* `location.dart`: added fields `countryCode`, `osmKey` and `osmValue`
* `location.g.dart`: generated

### DIFF
--- a/lib/src/prices/location.dart
+++ b/lib/src/prices/location.dart
@@ -46,6 +46,15 @@ class Location extends JsonObject {
   @JsonKey(name: 'osm_address_country')
   String? country;
 
+  @JsonKey(name: 'osm_address_country_code')
+  String? countryCode;
+
+  @JsonKey(name: 'osm_tag_key')
+  String? osmKey;
+
+  @JsonKey(name: 'osm_tag_value')
+  String? osmValue;
+
   @JsonKey(name: 'osm_lat')
   double? latitude;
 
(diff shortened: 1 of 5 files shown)


teolemon (approved):
Unrelated: have you spotted the blocklist of osm keys (to return supermarkets, and not  streets ?)

monsieurtanuki:

> Unrelated: have you spotted the blocklist of osm keys (to return supermarkets, and not streets ?)

You mean, as a filter? It looks like it's not supported by https://photon.komoot.io/, but we may filter the results returned by photon.
That said, eventually we may merge the results of several providers like photon and Nominatim. And then in Nominatim I think we can filter by key/value.

monsieurtanuki:
@teolemon Actually it's possible to filter by osm key: something like `osm_tag=shop` that I'm going to introduce in Smoothie.

PR #938, opened 2024-06-06
fix: added "app_name" parameter for prices methods uploadProof and createPrice

monsieurtanuki:
### What
- Now we pass an `app_name` parameter to prices methods `uploadProof` and `createPrice`, for tracking purposes.
- Meanwhile, some refactoring, including trying to make all tests pass.
- And some new prices methods somehow needed by Smoothie.

### Impacted files
* `api_get_taxonomy_origins_server_test.dart`: minor refactoring
* `api_get_user_products_test.dart`: set max page size to new server max page size (100)
* `api_search_products_test.dart`: set default page size to new server default page size (50)
* `api_suggestion_manager_test.dart`: now targeting PROD; minor refactoring
* `currency.dart`: new `fromName` method
* `open_prices_api_client.dart`: added `app_name` parameter for `uploadProof` and `createPrice`; refactored around new method `getUri`
* `page_size.dart`: added comments
* `price.dart`: added comments
* `proof.dart`: new `getFileUrl` method; added comments
* `proof_type.dart`: new `fromOffTag` method

### DIFF
--- a/lib/src/model/parameter/page_size.dart
+++ b/lib/src/model/parameter/page_size.dart
@@ -1,6 +1,8 @@
 import '../../interface/parameter.dart';
 
 /// "Page size" search API parameter
+///
+/// Typically defaults to 50 (used to be 24). Max value seems to be 100.
 class PageSize extends Parameter {
   @override
   String getName() => 'page_size';
--- a/lib/src/open_prices_api_client.dart
+++ b/lib/src/open_prices_api_client.dart
@@ -41,15 +41,26 @@ class OpenPricesAPIClient {
   static String _getHost(final UriProductHelper uriHelper) =>
       uriHelper.getHost(_subdomain);
 
+  static Uri getUri({
+    required final String path,
+    final Map<String, dynamic>? queryParameters,
+    final UriProductHelper uriHelper = uriHelperFoodProd,
+  }) =>
+      uriHelper.getUri(
+        path: path,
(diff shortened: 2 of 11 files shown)


monsieurtanuki:
cc @raphodn @raphael0202

raphodn:
Fyi in the prices web app I opted to pass the `app_name` in the URL of every request (so not only POST, but also GET).

I guess in Sentry if there are any errors, the User Agent is sent and available in any case. But i realised that some browsers do not allow to edit the UA (for instance Chrome).

monsieurtanuki:
> Fyi in the prices web app I opted to pass the `app_name` in the URL of every request (so not only POST, but also GET).

DELETE and PATCH too?

raphodn:
Yes every call to the API 👌

And you can send extra info like the `app_version` and any other relevant data as extra url params if you want.

I need to document at least the `app_name` in the API documentation

monsieurtanuki:
@raphodn I've just pushed a change where all prices URLs will include "user agent" fields (when populated):
* `app_name`
* `app_version` (optional)
* `app_uuid` (optional)
* `app_platform` (optional)
* `comment` (optional)

With the exception of the proof file URL where we don't add those parameters, e.g. https://prices.openfoodfacts.net/img/0002/yRw3zcph08.jpg

M123-dev (approved):

monsieurtanuki:

Issue #937, opened 2024-06-05
Flutter Web and the User Agent


g123k:
Hi everyone,

I have received several complaints about the SDK when used on the web.
Every time the error is `ClientException: XMLHttpRequest error.`.

After digging a little bit, the real Exception is:
`Refused to set unsafe header "User-Agent"` 

Today, the way we send the custom header on mobile and desktop is OK.
However, on the web, a User-Agent must follow a [strict format](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent).

So we have two choices here:
- Try to create a valid header entry
- Rename what we call "User-Agent" to another name

monsieurtanuki:
@g123k AFAIR the user-agent is used by the OFF server for stats, so any change would have an impact on the server stats, cc @stephanegigandet 
IMHO the easiest solution would be not to force the user-agent when used in web mode. I assume that the browser user-agent will overwrite that void, which would be a relatively relevant approximation for the OFF server. Or it will be kept null, which would also solve the current issue.
Btw we can easily detect "web" usage, we already do with code like that  - we consider the "js" version as "web":
```dart
import 'uri_reader_stub.dart'
    if (dart.library.io) 'uri_reader_io.dart'
    if (dart.library.js) 'uri_reader_js.dart';
```
Just my 2 cents.

PR #950, opened 2024-07-08
fix: 937 - removed home-made user agent header when for web

monsieurtanuki:
### What
- Our home-made user agent prevents our dart package to work when used in web mode.
- We use this specific user agent for stats.
- The solution implemented here removes the user agent http header when in web mode, so that the package can be used.
- Possible improvements would be to change the way we build our user agent string (but that would have an impact on our previous/next stats), or where we put that home-made string (e.g. different http header tag) (but that would have an impact on our stats consistency too)
- We'll see how many requests we get from web mode. Later we can select one of the suggested improvements, or something else.

### Fixes bug(s)
- Fixes: #937

### Impacted files
* `api_add_product_image_test.dart`: minor refactoring
* `http_helper.dart`: removed home-made user agent header when for web; minor refactoring
* `open_prices_api_client.dart`: minor refactoring
* `uri_reader.dart`: added an `isWeb` getter; minor refactoring
* `uri_reader_js.dart`: implemented the `isWeb` getter

### DIFF
--- a/lib/src/open_prices_api_client.dart
+++ b/lib/src/open_prices_api_client.dart
@@ -481,7 +481,7 @@ class OpenPricesAPIClient {
         if (currency != null) 'currency': currency.name,
       },
     );
-    final List<int> fileBytes = await UriReader.instance!.readAsBytes(imageUri);
+    final List<int> fileBytes = await UriReader.instance.readAsBytes(imageUri);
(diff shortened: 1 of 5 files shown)


g123k (approved):
Thanks for the change.
We were getting more and more complaints about this.

monsieurtanuki:

PR #953, opened 2024-07-19
feat: Add the name of the image contributor

g123k:
Hi everyone,

On the `images` field, the API returns the contributor's name.
Eg:

```json
        "images": {
            "1": {
                "sizes": {
                    "100": {
                        "h": 71,
                        "w": 100
                    },
                    "400": {
                        "h": 284,
                        "w": 400
                    },
                    "full": {
                        "h": 1089,
                        "w": 1536
                    }
                },
                "uploaded_t": 1345799270,
                "uploader": "openfoodfacts-contributors"


```

### DIFF
--- a/lib/src/utils/json_helper.dart
+++ b/lib/src/utils/json_helper.dart
@@ -98,6 +98,7 @@ class JsonHelper {
 
   // only for raw images
   static const String _ALL_IMAGES_TAG_UPLOADED = 'uploaded_t';
+  static const String _ALL_IMAGES_TAG_UPLOADER = 'uploader';
 
   // common to main and raw images
   static const String _ALL_IMAGES_TAG_IMAGE_ID = 'imgid';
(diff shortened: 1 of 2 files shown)


lib/src/utils/json_helper.dart
```diff
@@ -169,6 +170,7 @@ class JsonHelper {
               height: JsonObject.parseInt(numberObject[_ALL_IMAGES_TAG_HEIGHT]),
               url: numberObject[_ALL_IMAGES_TAG_URL],
               uploaded: uploaded,
+              contributor: fieldObject[_ALL_IMAGES_TAG_UPLOADER],
```
monsieurtanuki, line 173: `_ALL_IMAGES_TAG_UPLOADED` appears in both `allImagesToJson` and `allImagesFromJson`.
In your code `_ALL_IMAGES_TAG_UPLOADER` appears only once. Probably something needs to be fixed.
Additional unit tests would help, too.

g123k: I didn't include the 'to' method on purpose, because today, images are well assigned to the contributor.
But I can add do, no pb.

In terms of tests, despite checking the result of the server, I don't really see any benefits here.

M123-dev (approved):

teolemon:
Thanks @g123k @M123-dev 
Just to be sure, are we getting it as part of the standard payload ? Are there some cases where it might not be desirable (eg coke with loads of images ?)

PR #965, opened 2024-09-04
ci: Update github-projects.yml

teolemon:
### What
- Add dispatcher for Open Beauty, Pet Food, Products


monsieurtanuki (approved):
I don't speak `yaml` at all but that looks good.

PR #996, opened 2024-11-21
fix: product_type not included when null

monsieurtanuki:
### Impacted files:
* `product.dart`: made most fields as "includeIfNull: false", including product_type
* `product.g.dart`: impact only on product_type

### What
- The product field "product_type" was ALWAYS included in the "toJson" conversion, _even if null_.
- That had some problematic side effects on the server, that _always_ saw a product_type provided for every "save product details" operation in Smoothie. And there's only one moment when this field should be sent: when we _create_ the product. Afterwards, any save operation will be rejected if we populate the product_type, even with the same value.
- Now the product field "product_type" is only populated in the "toJson" conversion when it's not null, which only happens in Smoothie when we create the product in the correct OxF server.
- Beyond that, in order to avoid similar errors, a general code clean up should be performed around the "includeIfNull" clause for all classes, but that's another story.

### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/5864

### DIFF
--- a/lib/src/model/product.dart
+++ b/lib/src/model/product.dart
@@ -81,14 +81,15 @@ enum ProductImprovementCategory {
   NUTRI_SCORE,
 }
 
+// TODO(monsieurtanuki): define and implement a strategy for all classes regarding "includeIfNull: false"
 /// This class contains most of the data about a specific product.
 ///
 /// Please read the language mechanics explanation if you intend to display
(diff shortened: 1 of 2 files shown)


teolemon (approved):
👌 cc @stephanegigandet

monsieurtanuki:

Issue #1023, opened 2025-01-23
Unable to generate files using build_runner


PrimaelQuemerais:
### Description
When using a recent version of Flutter (Stable channel : 3.27.2, dart 3.6.1) we are unable to generate files using the command
`dart run build_runner build`

The command first outputs the following warning

```
The language version (2.17.0) of this package (openfoodfacts) does not match the required range `^3.0.0`.

Edit pubspec.yaml to include an SDK constraint of at least ^3.0.0.

environment:
  sdk: ^3.0.0
```

And then lists all the formatter exceptions.

Can we bump the minimum dart version of the packages?

### Expected behavior
Generate files

### Stacktraces
```
[SEVERE] json_serializable on lib/src/model/product.dart:
An error `FormatterException` occurred while formatting the generated source for
  `package:openfoodfacts/src/model/product.dart`
which was output to
  `lib/src/model/product.json_serializable.g.part`.
This may indicate an issue in the generator, the input source code, or in the
source formatter.
Could not format because the source could not be parsed:

line 108, column 56 of .: Expected to find ')'.
    ╷
108 │         if (_$ProductTypeEnumMap[instance.productType] case final value?) 'product_type': value,
    │                                                        ^^^^
    ╵
line 109, column 34 of .: Expected to find ')'.
    ╷
109 │         if (instance.productName case final value?) 'product_name': value,
    │                                  ^^^^
    ╵
line 110, column 77 of .: Expected to find ')'.
    ╷
110 │         if (LanguageHelper.toJsonStringMap(instance.productNameInLanguages) case final value?) 'product_name_in_languages': value,
    │                                                                             ^^^^
    ╵
line 111, column 34 of .: Expected to find ')'.
    ╷
111 │         if (instance.genericName case final value?) 'generic_name': value,
    │                                  ^^^^
    ╵
line 112, column 77 of .: Expected to find ')'.
    ╷
112 │         if (LanguageHelper.toJsonStringMap(instance.genericNameInLanguages) case final value?) 'generic_name_in_languages': value,
    │                                                                             ^^^^
    ╵
line 113, column 38 of .: Expected to find ')'.
    ╷
113 │         if (instance.abbreviatedName case final value?) 'abbreviated_product_name': value,
    │                                      ^^^^
    ╵
line 114, column 81 of .: Expected to find ')'.
    ╷
[...]

PR #1021, opened 2025-01-22
chore: 960 - refreshed pubspec.yaml

monsieurtanuki:
### What
- As we're between 2 major flutter versions, we need to be more precise regarding the package version numbers we use in `pubspec.yaml`.

### Part of 
- #960

cc. @AffanShaikhsurab

Issue #960, opened 2024-08-03
Add new localized ProductField's


monsieurtanuki:
### Why - Problem description
I've noticed new localized product fields:
* `conservation_conditions_fr`
* `customer_service_fr`

cf. https://fr.openfoodfacts.org/produit/3017620425035/nutella-ferrero.com?fields=conservation_conditions_fr,customer_service_fr

We should implement them as localized fields in `ProductField` and `Product`.

AffanShaikhsurab:
i would love you work on this issue

monsieurtanuki:
Hi @AffanShaikhsurab!
That should be easy - once you've understood how things work in that project for multilanguage fields.
Good luck!

AffanShaikhsurab:
I added the fields `conservation_conditions_fr` and `customer_service_fr` to the product object. However, I think the issue lies in the `conservation_conditions` field. The code seems to expect it to be of type `List<dynamic>` (probably because the field name is plural), but the data being returned is a `String`.

Here’s an example of the value being returned:  
`"conservation_conditions_fr": "A conserver au sec et à l'abri de la chaleur. Ne pas mettre au réfrigérateur."`

The error we’re encountering is:  
`Error fetching product: TypeError: "A conserver à l'abri du soleil et du gel, dans un endroit propre et sans odeur.": type 'String' is not a subtype of type 'List<dynamic>?`

How should we approach resolving this issue? Should we:
1. Update the API/logic to return a `List<String>` instead of a single `String` for `conservation_conditions_fr`?  
2. Change the code logic to handle `String` instead of `List<dynamic>` for this specific case?

monsieurtanuki:
> I added the fields `conservation_conditions_fr` and `customer_service_fr` to the product object.

@AffanShaikhsurab That may be part of the problem: we don't care about specifically French labels, we want _multilingual_ labels.
I should rather have mentioned https://fr.openfoodfacts.org/produit/3017620425035/nutella-ferrero.com?fields=conservation_conditions_languages,customer_service_languages
```json
{
  "code": "3017620425035",
  "product": {
    "conservation_conditions_languages": {
      "fr": "A conserver au sec et à l'abri de la chaleur. Ne pas mettre au réfrigérateur."
    },
    "customer_service_languages": {
      "fr": "FERRERO FRANCE COMMERCIALE - Service Consommateurs, CS 90058 - 76136 MONT SAINT AIGNAN Cedex"
    }
  },
[...]

PR #1024, opened 2025-01-23
chore: 960 - refreshed pubspec.yaml, again

monsieurtanuki:
### What
- Same as #1021, but with a little twist.
- Looks like recent versions of `json_serializable` use that brand new stupid syntax with `if(... case ...)`, which is not supported everywhere.
- Therefore, we have to use the best version of `json_serializable` without that syntax.

### Fixes bug(s)
- Fixes: #1023

### Part of 
- #960

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -17,7 +17,7 @@ dependencies:
 dev_dependencies:
   analyzer: 6.11.0
   build_runner: 2.4.9
-  json_serializable: 6.9.0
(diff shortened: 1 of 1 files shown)


PrimaelQuemerais (approved):
This change is required if we are not willing to bump dart to version 3.0.0

monsieurtanuki:
Thank you @PrimaelQuemerais for your review!

Issue #738, opened 2023-06-02
Implement Robotoff nutrition prediction route


teolemon:
### What
- Implement Robotoff nutrition prediction route
- https://openfoodfacts.github.io/robotoff/references/api/#tag/Predict/paths/~1predict~1nutrition/get
- Related smoothie issue: https://github.com/openfoodfacts/smooth-app/issues/3278

monsieurtanuki:
@teolemon @raphael0202 Not really sure what we try to achieve here, but maybe I'm mislead by my unsuccessful tests.

I get the following result with https://robotoff.openfoodfacts.org/api/v1/predict/nutrition?barcode=7300400481588&image_ids=55:
```json
{"predictions":[],"image_ids":["55"]}
```

While the image 55 is something like that:
![Capture d’écran 2023-07-30 à 18 12 00](https://github.com/openfoodfacts/openfoodfacts-dart/assets/11576431/3ac17ac1-7d5f-47bf-ab8d-ae55028c9eec)

Other points:
* we don't specify language or country, which is also disconcerting
* "imageIds" images can be big and include much more that nutrition facts, which may be problematic (for the AI) in order to focus on nutrition only

raphael0202:
The current endpoint only works well on "text" nutrition tables. I would be in favor of waiting for a proper generic model for nutrition information extraction to be developed before its integration in Smoothie.

teolemon:
@raphael0202 @monsieurtanuki We should integrate this route, which is now mature, and tested on Hunger Games.

https://robotoff.openfoodfacts.org/api/v1/insights?insight_types=nutrient_extraction&barcode={BARCODE}

raphael0202:
> One suggestion: make it possible to return all the values, and let the developer compare with what is already there in the product.

This is a valid point, what I fear if we have the current nutrient data saved in the insight is that it gets out of sync. Due to how Robotoff work, in case of massive updates, we can wait a long time before insights get refreshed.

We can add the data on the fly when querying the insight, but it would considerably slow down the request, I'm not sure it's a good strategy here.

monsieurtanuki:
@raphael0202 Misunderstanding alert!

My assumptions were:
* robotoff works on the nutrition facts image
* robotoff manages to OCR nutrition facts from that image
* robotoff would be able to return the nutrition facts OCR'ed from that image, et voilà!

I don't assume that robotoff should be aware of the nutrition facts _as current values at the product level_.
[...]

Issue #584, opened 2022-10-13
Nutrient: support xx_modifier


g123k:
Hi everyone,

On the website, we can input a modifier (eg: `>` or `<`) and it's available through a field called `modifier`.
Here is an example with the salt:
`"nutriments":{"fat":1,"fat_100g":1,"fat_modifier":">","fat_unit":"g","fat_value":1}`

We should support that value on the Dart SDK.
And will also fix https://github.com/openfoodfacts/smooth-app/issues/3135

monsieurtanuki:
@g123k I used to see those modifiers, but they are not in the website, are they?
![Capture d’écran 2022-10-13 à 15 15 40](https://user-images.githubusercontent.com/11576431/195606732-f394237c-9716-4c4e-98ca-1981a76d194f.png)

g123k:
On the website, just change the value to ">892" and it will appear.
By default, it's just like if it's an `=892`

monsieurtanuki:
Thank you @g123k!
So you mean that the full list of modifiers is as following, right?
* `<`
* `>`
* `~`
* `=` (default value, implicit when modifier is null)

g123k:
@stephanegigandet or @alexgarel Could you confirm this, please?

stephanegigandet:
There's no "=" value.

Also, not a modifier, but if you use "-" as the value, it means that the field is not specified on the packaging. (useful for fibers in particular).

stephanegigandet:
e.g. if you input:

"-" for fibers
"~0" for proteins
"<10" for sugars

you will get this:

```
nutriments: {
fiber_modifier: "-",
fruits-vegetables-nuts-estimate-from-ingredients_100g: 0,
fruits-vegetables-nuts-estimate-from-ingredients_serving: 0,
nova-group: 1,
nova-group_100g: 1,
nova-group_serving: 1,
proteins: 0,
proteins_100g: 0,
proteins_modifier: "~",
proteins_unit: "g",
proteins_value: 0,
sugars: 10,
sugars_100g: 10,
sugars_modifier: "<",
sugars_unit: "g",
sugars_value: 10
},
```

PR #1026, opened 2025-01-23
feat: Added robotoff nutrient extraction

PrimaelQuemerais:
### What
This PR adds the nutrient extraction Robotoff feature to the package

### Part of 
#738


lib/src/robot_off_api_client.dart
```diff
@@ -1,6 +1,7 @@
 import 'dart:convert';
 
 import 'package:http/http.dart';
+import 'package:openfoodfacts/src/model/robotoff_nutrient_extraction.dart';
```
monsieurtanuki, line 4:
```suggestion
import 'model/robotoff_nutrient_extraction.dart';
```

lib/src/robot_off_api_client.dart
```diff
@@ -192,6 +193,32 @@ class RobotoffAPIClient {
     return Status.fromApiResponse(response.body);
   }
 
+  /// Get nutrient extraction insights for a given barcode
+  /// cf. https://robotoff.openfoodfacts.org/api/v1/insights
+  static Future<RobotoffNutrientExtractionResult> getNutrientExtraction(
+    String barcode, {
```
monsieurtanuki, line 199:
```suggestion
    final String barcode, {
```

test/api_get_robotoff_test.dart
```diff
@@ -1,4 +1,5 @@
 import 'package:openfoodfacts/openfoodfacts.dart';
+import 'package:openfoodfacts/src/model/robotoff_nutrient_extraction.dart';
```
monsieurtanuki, line 2:
```suggestion
```

That file should be included in `openfoodfacts.dart`.

lib/src/model/old_product_result.g.dart
```diff
@@ -1,7 +1,5 @@
 // GENERATED CODE - DO NOT MODIFY BY HAND
 
-// ignore_for_file: deprecated_member_use_from_same_package
```
monsieurtanuki, line 3: Please keep it there, we need it.

lib/src/model/robotoff_nutrient_extraction.dart
```diff
@@ -0,0 +1,201 @@
+import 'package:json_annotation/json_annotation.dart';
+import '../interface/json_object.dart';
+
+part 'robotoff_nutrient_extraction.g.dart';
+
+@JsonSerializable()
+class RobotoffNutrientExtractionResult extends JsonObject {
```
monsieurtanuki, line 7: Generally speaking, I would prefer one file per class (I'm from Java) and an explicit `Robotoff` prefix for all class names.
Not the most important part of my review, though.

lib/src/model/robotoff_nutrient_extraction.dart
```diff
@@ -0,0 +1,201 @@
...
+  final List<NutrientExtractionInsight>? insights;
+
+  const RobotoffNutrientExtractionResult({
+    this.status,
+    this.count,
+    this.insights,
+  });
+
+  NutrientExtractionInsight? get getLatestInsights {
+    insights?.sort((a, b) => a.completedAt!.compareTo(b.completedAt!));
```
monsieurtanuki, line 19: Please deal with the case where `completedAt` is null, as you don't reject that possibility.

lib/src/model/robotoff_nutrient_extraction.dart
```diff
@@ -0,0 +1,201 @@
...
+    return insights?.last;
+  }
+
+  NutrientEntity? getNutrientEntity(String nutrientKey) {
```
monsieurtanuki, line 23: As with most methods regarding nutrients, please consider re-using or using something similar to `String _getTag(Nutrient nutrient, PerSize perSize)` in `nutriments.dart`.

lib/src/model/robotoff_nutrient_extraction.dart
```diff
@@ -0,0 +1,201 @@
...
+  Map<String, dynamic> toJson() => _$NutrientAnnotationDataToJson(this);
+}
+
+@JsonSerializable()
+class NutrientAnnotation {
+  final Map<String, NutrientAnnotationData>? nutrients;
+  @JsonKey(name: 'serving_size')
+  final String? servingSize;
+  @JsonKey(name: 'nutrition_data_per')
+  final String? nutritionDataPer;
```
monsieurtanuki, line 92: Please consider using `PerSize`.

lib/src/model/robotoff_nutrient_extraction.dart
```diff
@@ -0,0 +1,201 @@
...
+  factory NutrientEntity.fromJson(Map<String, dynamic> json) =>
+      _$NutrientEntityFromJson(json);
+
+  Map<String, dynamic> toJson() => _$NutrientEntityToJson(this);
+}
+
+@JsonSerializable()
+class NutrientAnnotationData {
+  final String? unit;
+  final String? value;
```
monsieurtanuki, line 73: Quite obscure `String? unit` and `String? value`.
Are these randoms strings, or are we supposed to get typical values from a list.

lib/src/model/robotoff_nutrient_extraction.dart
```diff
@@ -0,0 +1,201 @@
...
+  final int? uniqueScansN;
+  @JsonKey(name: 'reserved_barcode')
+  final bool? reservedBarcode;
+  final String? predictor;
+  @JsonKey(name: 'predictor_version')
+  final String? predictorVersion;
+  final List<String>? campaign;
+  final double? confidence;
+  @JsonKey(name: 'bounding_box')
+  final dynamic boundingBox;
```
monsieurtanuki, line 166: `dynamic`?

lib/src/model/robotoff_nutrient_extraction.dart
```diff
@@ -0,0 +1,201 @@
...
+@JsonSerializable()
+class NutrientExtractionInsight extends JsonObject {
+  @JsonKey(name: 'id')
+  final String? insightId;
+  final String? barcode;
+  final String? type;
+  final NutrientDataWrapper? data;
+  final String? timestamp;
+  @JsonKey(name: 'completed_at')
+  final String? completedAt;
```
monsieurtanuki, line 136: Is that a date?

lib/src/model/robotoff_nutrient_extraction.dart
```diff
@@ -0,0 +1,201 @@
...
+  @JsonKey(name: 'predictor_version')
+  final String? predictorVersion;
+  final List<String>? campaign;
+  final double? confidence;
+  @JsonKey(name: 'bounding_box')
+  final dynamic boundingBox;
+
+  const NutrientExtractionInsight({
+    this.insightId,
+    this.barcode,
```
monsieurtanuki, line 170: I'm not sure using `final` fields is very helpful.
That forces us to maintain the list of fields in the constructor.

lib/src/model/robotoff_nutrient_extraction.dart
```diff
@@ -0,0 +1,201 @@
...
+  factory NutrientAnnotation.fromJson(Map<String, dynamic> json) =>
+      _$NutrientAnnotationFromJson(json);
+
+  Map<String, dynamic> toJson() => _$NutrientAnnotationToJson(this);
+}
+
+@JsonSerializable()
+class NutrientDataWrapper {
+  final Map<String, List<NutrientEntity>>? entities;
+  final Map<String, NutrientEntity>? nutrients;
```
monsieurtanuki, line 109: That `String` key is a vague.
Please consider adding getters with `Nutrient` and `PerSize` parameters.

test/api_get_robotoff_test.dart
```diff
@@ -218,4 +219,20 @@ void main() {
...
+    final RobotoffNutrientExtractionResult result =
+        await RobotoffAPIClient.getNutrientExtraction(
+      barcode,
+    );
+
+    expect(result.status, isNotNull);
+    expect(result.status, 'found');
+    expect(result.insights!, isNotEmpty);
+    expect(result.insights![0].barcode, barcode);
+    expect(result.insights![0].data!.nutrients!['fat_100g']!.value, '0.5');
```
monsieurtanuki, line 235: Nutrient + PerSize

lib/src/model/robotoff_nutrient_extraction.dart
```diff
@@ -0,0 +1,201 @@
...
+}
+
+@JsonSerializable()
+class NutrientExtractionInsight extends JsonObject {
+  @JsonKey(name: 'id')
+  final String? insightId;
+  final String? barcode;
+  final String? type;
+  final NutrientDataWrapper? data;
+  final String? timestamp;
```
monsieurtanuki, line 134: Is that a date?

lib/src/model/robotoff_nutrient_extraction.dart
```diff
@@ -0,0 +1,201 @@
...
+
+  Map<String, dynamic> toJson() => _$NutrientDataWrapperToJson(this);
+}
+
+@JsonSerializable()
+class NutrientExtractionInsight extends JsonObject {
+  @JsonKey(name: 'id')
+  final String? insightId;
+  final String? barcode;
+  final String? type;
```
monsieurtanuki, line 132: Are there known types?

lib/src/model/robotoff_nutrient_extraction.dart
```diff
@@ -0,0 +1,201 @@
...
+  final String? timestamp;
+  @JsonKey(name: 'completed_at')
+  final String? completedAt;
+  final int? annotation;
+  @JsonKey(name: 'annotated_result')
+  final int? annotatedResult;
+  @JsonKey(name: 'n_votes')
+  final int? nVotes;
+  final String? username;
+  final List<String>? countries;
```
monsieurtanuki, line 143: Localized countries? Country codes?

monsieurtanuki (commented):
Thank you @PrimaelQuemerais for your PR!

lib/src/model/robotoff_nutrient_extraction.dart
PrimaelQuemerais: This is the kind of values in the list : "en:france"
Do we have an object to wrap this into? Or do you require to add a comment stating the format?

lib/src/model/robotoff_nutrient_extraction.dart
PrimaelQuemerais: Here the type is "nutrition_extractor" do we have an object for this?

lib/src/model/robotoff_nutrient_extraction.dart
monsieurtanuki: `"en:france"` is unfortunately somehow useless. `'fr'` is useful.
If you could convince the server to send https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2 codes instead that would be better. If it's not possible, at least add a comment.

lib/src/model/robotoff_nutrient_extraction.dart
monsieurtanuki: > Here the type is "nutrition_extractor" do we have an object for this?

I guess you'll have to create a new class or enum for that.

lib/src/model/robotoff_nutrient_extraction.dart
PrimaelQuemerais: I've changed unit to be of type Unit, but value cannot be parsed as a number as it contains symbols like '<'

monsieurtanuki: > value cannot be parsed as a number as it contains symbols like '<'

Understood. A bit problematic as we don't support '<' in off-dart and Smoothie AFAIK. cf. https://github.com/openfoodfacts/openfoodfacts-dart/issues/584

Would be a good opportunity to
* create a `NutrientModifier` enum (<, >, ~)
* in your class, create 2 getters (`NutrientModifier? getNutrientModifier`, `double? getNutrientValue`)

Or something like that.

PrimaelQuemerais: Sounds good, I'm on it

lib/src/model/robotoff_nutrient_extraction.dart
PrimaelQuemerais: I'm not sure what level of abstraction this represents. Is it okay to leave it as a string until we have a second endpoint at the same level?

monsieurtanuki: If you have a limited understanding of what this field is for, and as a consequence about its usage, it would be better to ignore that field temporarily (e.g. commenting and adding a `TODO`).
Once we've decided that a field has a given type, it's a breaking change to change its type. Let's avoid that.

PrimaelQuemerais: Good point, updating now

lib/src/model/nutrient.dart
```diff
@@ -254,3 +256,7 @@ enum Nutrient implements OffTagged {
       ? Nutrient.energyKJ
       : OffTagged.fromOffTag(offTag, Nutrient.values) as Nutrient?;
 }
+
+extension NutrientExtension on Nutrient {
```
monsieurtanuki, line 260: We don't need an extension, we could put it directly in the enum, couldn't we?

lib/src/model/robotoff_nutrient_extraction_annotation.dart
```diff
@@ -0,0 +1,46 @@
...
+import 'package:openfoodfacts/src/model/per_size.dart';
+import 'package:openfoodfacts/src/utils/unit_helper.dart';
+
+part 'robotoff_nutrient_extraction_annotation.g.dart';
+
+@JsonSerializable()
+class RobotoffNutrientAnnotationData {
+  @JsonKey(toJson: UnitHelper.unitToString, fromJson: UnitHelper.stringToUnit)
+  final Unit? unit;
+  final String value;
```
monsieurtanuki, line 12: What about `valueWithModifier`?

lib/src/model/robotoff_nutrient_extraction_annotation.dart
```diff
@@ -0,0 +1,46 @@
...
+  final NutrientModifier? modifier;
+  final double? numericalValue;
+
+  RobotoffNutrientAnnotationData({
+    this.unit,
+    required this.value,
+  })  : modifier = NutrientModifierExtension.fromValue(value),
+        numericalValue = double.tryParse(value.trim().substring(1));
```
monsieurtanuki, line 20: What if there's no modifier? Still starting at 1?
What if the string is empty?
Besides, getters would be good enough, instead of final fields.

lib/openfoodfacts.dart
```diff
@@ -58,6 +58,7 @@ export 'src/model/product_state.dart';
 export 'src/model/product_stats.dart';
 export 'src/model/product_tag.dart';
 export 'src/model/recommended_daily_intake.dart';
+export 'src/model/robotoff_nutrient_extraction.dart';
```
monsieurtanuki, line 61: All your new files would deserve to be here, as long as users need them.

lib/src/model/old_product_result.g.dart
```diff
@@ -1,9 +1,9 @@
...
-
 part of 'old_product_result.dart';
 
+// ignore_for_file: deprecated_member_use_from_same_package
```
monsieurtanuki, line 5: I'm not (100%) a maniac, but please put the "ignore" comment where it was.
Much better later when looking at the code we'll see that no one touched this file for months, instead of a recent useless change.

lib/src/model/robotoff_nutrient_extraction_insight.dart
```diff
@@ -0,0 +1,140 @@
+import 'package:json_annotation/json_annotation.dart';
+import 'package:openfoodfacts/src/model/nutrient.dart';
```
monsieurtanuki, line 2:
```suggestion
import 'nutrient.dart';
```

lib/src/model/robotoff_nutrient_extraction_insight.dart
```diff
@@ -0,0 +1,140 @@
...
+import 'package:openfoodfacts/src/model/per_size.dart';
```
monsieurtanuki, line 3:
```suggestion
import 'per_size.dart';
```

lib/src/model/robotoff_nutrient_extraction_insight.dart
```diff
@@ -0,0 +1,140 @@
...
+  String? insightId;
+  String? barcode;
+  // TODO: Figure out what this field is and what type it should be
+  // final String? type;
+  final RobotoffNutrientDataWrapper? data;
+  // @JsonKey(
+  //     fromJson: JsonHelper.timestampToDate, toJson: JsonHelper.dateToTimestamp)
+  // TODO ask server to send timestamp as milliseconds since epoch
+  // Current format is 2024-12-26T14:13:51.299483
+  final String? timestamp;
```
monsieurtanuki, line 63: We already manage those kinds of timestamp formats, I believe.
Please have a look at `JsonHelper`.

monsieurtanuki:
@PrimaelQuemerais Btw I would safer/cleaner to use a dedicated branch instead of master.
For the next PR...

lib/src/model/nutrient.dart
```diff
@@ -1,3 +1,5 @@
+import 'per_size.dart';
+
```
monsieurtanuki, line 2:
```suggestion
```

lib/src/model/nutrient_modifier.dart
```diff
@@ -0,0 +1,51 @@
+import 'package:openfoodfacts/src/model/off_tagged.dart';
+
+enum NutrientModifier implements OffTagged {
+  approximately,
+  superiorTo,
+  inferiorTo;
+
+  @override
+  String get offTag => symbol;
+}
```
monsieurtanuki, line 10:
```suggestion
enum NutrientModifier implements OffTagged {
  approximately(offTag: '~'),
  greaterThan(offTag: '>'),
  lessThan(offTag: '<');

  const NutrientModifier({
    required this.offTag,
  });

  @override
  final String offTag;

  /// Returns the first [NutrientModifier] that matches the [offTag].
  static NutrientModifier? fromOffTag(final String? offTag) =>
      OffTagged.fromOffTag(offTag, NutrientModifier.values) as NutrientModifier?;
}
```

I'm not saying what you coded is wrong, but what about replicating the way we code other `OffTagged`s?

Btw in English they rather say `greaterThan` and `lessThan` ;)

lib/src/model/nutrient_modifier.dart
```diff
@@ -0,0 +1,51 @@
...
+  String get symbol {
+    switch (this) {
+      case NutrientModifier.approximately:
+        return '~';
+      case NutrientModifier.superiorTo:
+        return '>';
+      case NutrientModifier.inferiorTo:
+        return '<';
+    }
+  }
```
monsieurtanuki, line 22: We don't need that "symbol", it's `offTag`.

lib/src/model/nutrient_modifier.dart
```diff
@@ -0,0 +1,51 @@
...
+      case '~':
+        return NutrientModifier.approximately;
+      case '>':
+        return NutrientModifier.superiorTo;
+      case '<':
+        return NutrientModifier.inferiorTo;
+      default:
+        return null;
+    }
+  }
```
monsieurtanuki, line 41: We don't need that, it's dealt with `fromOffTag`.

lib/src/model/robotoff_nutrient_extraction.dart
```diff
@@ -0,0 +1,88 @@
+import 'package:json_annotation/json_annotation.dart';
+import 'package:openfoodfacts/src/model/per_size.dart';
```
monsieurtanuki, line 2:
```suggestion
import 'per_size.dart';
```

lib/src/model/robotoff_nutrient_extraction.dart
```diff
@@ -0,0 +1,88 @@
...
+import 'package:openfoodfacts/src/model/robotoff_nutrient_extraction_annotation.dart';
```
monsieurtanuki, line 3:
```suggestion
import 'robotoff_nutrient_extraction_annotation.dart';
```

lib/src/model/robotoff_nutrient_extraction_annotation.dart
```diff
@@ -0,0 +1,54 @@
...
+  String valueWithModifer;
+
+  RobotoffNutrientAnnotationData({
+    this.unit,
+    required this.valueWithModifer,
+  });
+
+  get modifier => NutrientModifierExtension.fromValue(valueWithModifer);
+
+  get value {
```
monsieurtanuki, line 21: If possible I prefer explicit types rather than guessing them each time I read a code.
Besides, it's much safer - maintenance-wise.

lib/src/model/robotoff_nutrient_extraction_annotation.dart
```diff
@@ -0,0 +1,54 @@
...
+  @JsonKey(toJson: UnitHelper.unitToString, fromJson: UnitHelper.stringToUnit)
+  Unit? unit;
+  String valueWithModifer;
+
+  RobotoffNutrientAnnotationData({
+    this.unit,
+    required this.valueWithModifer,
+  });
+
+  get modifier => NutrientModifierExtension.fromValue(valueWithModifer);
```
monsieurtanuki, line 19: If possible I prefer explicit types rather than guessing them each time I read a code.
Besides, it's much safer - maintenance-wise.

lib/src/model/nutrient_modifier.dart
```diff
@@ -0,0 +1,51 @@
...
+enum NutrientModifier implements OffTagged {
+  approximately,
+  superiorTo,
+  inferiorTo;
+
+  @override
+  String get offTag => symbol;
+}
+
+extension NutrientModifierExtension on NutrientModifier {
```
monsieurtanuki, line 12: For the record we don't need an extension.
An `enum` can have `static`  methods.

test/api_get_robotoff_test.dart
```diff
@@ -218,4 +218,47 @@ void main() {
...
+      Nutrient.fat,
+      Nutrient.salt,
+      Nutrient.fiber,
+      Nutrient.sugars,
+      Nutrient.proteins,
+      Nutrient.energyKJ,
+      Nutrient.energyKCal,
+      Nutrient.carbohydrates,
+      Nutrient.saturatedFat,
+    ]) {
```
monsieurtanuki, line 259:
```suggestion
```

lib/src/utils/json_helper.dart
```diff
@@ -408,6 +408,12 @@ class JsonHelper {
   static DateTime? nullableStringTimestampToDate(dynamic json) =>
       json == null ? null : stringTimestampToDate(json);
 
+  static String dateToStringTimestamp(DateTime dateTime) =>
+      dateTime.toIso8601String();
+
+  static String? nullableDateToStringTimestamp(DateTime? dateTime) =>
+      dateTime == null ? null : dateToStringTimestamp(dateTime);
+
```
monsieurtanuki, line 416: Looks very similar to what is already coded.
If your methods are really useful, that means that parts of the existing code is wrong - which is of course possible.
Would you please check that we really need your methods, like, what if we transform an object into a json into an object, is the datetime kept?
```dart
  /// Returns a [DateTime] from a JSON-encoded String (e.g. '2021-10-29T11:00:56.177379')
  static DateTime stringTimestampToDate(dynamic json) =>
      DateTime.parse(json as String);

  /// Returns a [DateTime] from a JSON-encoded String (e.g. '2021-10-29T11:00:56.177379')
  static DateTime? nullableStringTimestampToDate(dynamic json) =>
      json == null ? null : stringTimestampToDate(json);
```

lib/src/model/robotoff_nutrient_extraction_insight.dart
```diff
@@ -0,0 +1,138 @@
...
+  final List<String>? campaign;
+  final double? confidence;
+  // TODO: Figure out what this field is and what type it should be
+  // @JsonKey(name: 'bounding_box')
+  // final dynamic boundingBox;
+
+  RobotoffNutrientExtractionInsight({
+    this.insightId,
+    this.barcode,
+    // this.type,
```
monsieurtanuki, line 106: No comments. Or add a `TODO` here too.

lib/src/model/robotoff_nutrient_extraction_insight.dart
```diff
@@ -0,0 +1,138 @@
...
+    this.sourceImage,
+    this.automaticProcessing,
+    this.serverType,
+    this.uniqueScansN,
+    this.reservedBarcode,
+    this.predictor,
+    this.predictorVersion,
+    this.campaign,
+    this.confidence,
+    // this.boundingBox,
```
monsieurtanuki, line 128: No comments. Or add a `TODO` here too.

lib/src/model/robotoff_nutrient_extraction.dart
```diff
@@ -0,0 +1,88 @@
...
+  final int? count;
+  final List<RobotoffNutrientExtractionInsight>? insights;
+
+  const RobotoffNutrientExtractionResult({
+    this.status,
+    this.count,
+    this.insights,
+  });
+
+  RobotoffNutrientExtractionInsight? get getLatestInsights {
```
monsieurtanuki, line 22: I don't find it very elegant to compute this again and again.
I'd prefer if you could compute that only once.
Perhaps with a bool private flag? That would prevent you from using a `const` class, which isn't a big loss, IMHO.
Please ignore this comment if you cannot find quickly a clean solution.

monsieurtanuki (commented):
Hi @PrimaelQuemerais!
Still some comments. We're close to the end ;)

PrimaelQuemerais:
Hi @monsieurtanuki 
Thanks for taking the time to review everything. I've made all the changes you suggested :)

lib/src/model/robotoff_nutrient_extraction.dart
```diff
@@ -0,0 +1,108 @@
...
+  RobotoffNutrientExtractionInsight? _latestInsight;
+
+  RobotoffNutrientExtractionResult({
+    this.status,
+    this.count,
+    this.insights,
+  });
+
+  RobotoffNutrientExtractionInsight? getLatestInsights({
+    bool recompute = false,
```
monsieurtanuki, line 24: I would remove that "recompute" possibility, because I can't see a good reason to recompute.

lib/src/model/robotoff_nutrient_extraction_annotation.dart
```diff
@@ -0,0 +1,56 @@
...
+  NutrientModifier? get modifier =>
+      NutrientModifier.fromValue(valueWithModifer);
+
+  double? get value {
+    if (valueWithModifer.trim().isEmpty) {
+      return null;
+    }
+    return modifier == null
+        ? double.tryParse(valueWithModifer.trim())
+        : double.tryParse(valueWithModifer.trim().substring(1));
```
monsieurtanuki, line 29:
```suggestion
    final String trimmed = valueWithModifer.trim();
    if (trimmed.isEmpty) {
      return null;
    }
    return modifier == null
        ? double.tryParse(trimmed)
        : double.tryParse(trimmed.substring(1));
```

lib/src/model/robotoff_nutrient_extraction_insight.dart
```diff
@@ -0,0 +1,132 @@
...
+  final DateTime? completedAt;
+  final int? annotation;
+  @JsonKey(name: 'annotated_result')
+  final int? annotatedResult;
+  @JsonKey(name: 'n_votes')
+  final int? nVotes;
+  final String? username;
+  // TODO: ask server to send countries as ISO 3166-1 alpha-2 codes
+  // Current format is "en:france"
+  final List<String>? countries;
```
monsieurtanuki, line 73: In order to avoid future breaking changes, I would comment this field too.

monsieurtanuki (approved):
Thank you @PrimaelQuemerais!
That's good! Please have a look at my comments, though, you may find some interesting points. Feel free to ignore them.

monsieurtanuki:
Thank you @PrimaelQuemerais for your latest changes, and feel free to merge your PR!
(I can do it in case for some reason you don't see the "merge" button)

PR #1034, opened 2025-02-06
fix: Robotoff Nutrient Extraction Improvement

PrimaelQuemerais:
Small improvement to robotoff nutrient extraction. The unit in RobotoffNutrientEntity is now of type Unit instead of String


monsieurtanuki (approved):
Hi @PrimaelQuemerais!
LGTM

PR #1033, opened 2025-02-04
feat: added product fields 'data_quality_{*_}tags'

g123k:
This PR allows to get all data quality tags

Issue #1032, opened 2025-01-30
Unable to generate code with the latest version of Flutter/Dart


g123k:
Hi!

I would like to implement #1031, but unfortunately, I can't due to this error:

```
Building package executable... 
Failed to build build_runner:build_runner:
../../../.pub-cache/hosted/pub.dev/dart_style-2.3.2/lib/src/source_visitor.dart:593:19: Error: The getter 'inlineKeyword' isn't defined for the class 'ClassDeclaration'.
 - 'ClassDeclaration' is from 'package:analyzer/src/dart/ast/ast.dart' ('../../../.pub-cache/hosted/pub.dev/analyzer-6.11.0/lib/src/dart/ast/ast.dart').
Try correcting the name to the name of an existing getter, or defining a getter or field named 'inlineKeyword'.
    modifier(node.inlineKeyword);
                  ^^^^^^^^^^^^^
Failed to update packages.
````

Is-it OK to upgrade the dependencies (at least `analyzer` & `test`)?

**Comment by monsieurtanuki** [comment]
@g123k Very strange, I've just tested:
```log
PS C:\Users\fabri\StudioProjects\openfoodfacts-dart> dart run build_runner build
Building package executable... (1.9s)
Built build_runner:build_runner.
[INFO] Generating build script completed, took 547ms
[INFO] Reading cached asset graph completed, took 223ms
[INFO] Checking for updates since last build completed, took 1.3s
[INFO] Running build completed, took 18.7s
[INFO] Caching finalized dependency graph completed, took 280ms
[INFO] Succeeded after 19.0s with 164 outputs (363 actions)
```

```log
PS C:\Users\fabri\StudioProjects\openfoodfacts-dart> flutter doctor
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.27.3, on Microsoft Windows [version 10.0.26100.2894], locale fr-FR)
[✓] Windows Version (Installed version of Windows is version 10 or higher)
[✓] Android toolchain - develop for Android devices (Android SDK version 35.0.1)
[✓] Chrome - develop for the web
[✗] Visual Studio - develop Windows apps
    ✗ Visual Studio not installed; this is necessary to develop Windows apps.
      Download at https://visualstudio.microsoft.com/downloads/.
      Please install the "Desktop development with C++" workload, including all of its default components
[✓] Android Studio (version 2024.2)
[✓] Connected device (4 available)
[✓] Network resources

[...]

Issue #1031, opened 2025-01-29
Add data quality tags


g123k:
Hi!

In the API, we have access to data quality tags:
```
data_quality_tags
data_quality_bugs_tags
data_quality_errors_tags
data_quality_info_tags
data_quality_warnings_tags
```

It would be really nice to add them to the Dart SDK.

[An example of response](https://world.openfoodfacts.org/api/v3/product/0098765432187.json):

```
    "data_quality_bugs_tags": [],
    "data_quality_errors_tags": [],
    "data_quality_info_tags": [
      "en:no-packaging-data",
      "en:environmental-score-extended-data-not-computed",
      "en:food-groups-1-unknown",
      "en:food-groups-2-unknown",
      "en:food-groups-3-unknown"
    ],
    "data_quality_tags": [
      "en:no-packaging-data",
      "en:environmental-score-extended-data-not-computed",
      "en:food-groups-1-unknown",
      "en:food-groups-2-unknown",
      "en:food-groups-3-unknown",
      "en:environmental-score-origins-of-ingredients-origins-are-100-percent-unknown",
      "en:environmental-score-packaging-packaging-data-missing",
      "en:environmental-score-production-system-no-label",
      "en:environmental-score-threatened-species-ingredients-missing"
    ],
    "data_quality_warnings_tags": [
      "en:environmental-score-origins-of-ingredients-origins-are-100-percent-unknown",
      "en:environmental-score-packaging-packaging-data-missing",
      "en:environmental-score-production-system-no-label",
      "en:environmental-score-threatened-species-ingredients-missing"
    ],

monsieurtanuki:
@g123k In order to be reactive, I guess we need to keep those entries as `String`, instead of an `enum`.

g123k:
> [<img alt="" width="16" height="16" src="https://avatars.githubusercontent.com/u/246838?u=adee4a7bd1f3a291070b46b9e8ee2ac911ced518&amp;v=4&amp;size=80">@g123k](https://github.com/g123k) In order to be reactive, I guess we need to keep those entries as `String`, instead of an `enum`.

Yes, exactly.
We will let the clients do that (and I think it will only be smooth-app) 😛

teolemon:
We also have a data quality knowledge panel accessible for logged in users, we did the groundwork for authentified read queries before @g123k

g123k:
[...]

PR #1037, opened 2025-02-10
chore: fixed 1033 and a minor test bug

monsieurtanuki:
### What
- There is a bug in #1033, and here is the fix.
- Minor test bug fixed too


monsieurtanuki:
Thank you @g123k for your quick review!

g123k:
> Thank you @g123k for your quick review!

That's strange for the `.g.dart` file, because it was generated.
But thanks for the fix

monsieurtanuki:
> > Thank you @g123k for your quick review!
> 
> That's strange for the `.g.dart` file, because it was generated. But thanks for the fix

Would be related to #1032 (or can you close it now?)
Btw unless I'm missing something #1031 could be closed too, couldn't it?

Besides, the problem with `.g.dart` files is that github somehow hide their diffs.

PR #1042, opened 2025-02-14
feat: Nutrient modifiers (Get/Send product APIs)

g123k:
Hi everyone!

Finally, a working solution for nutrient modifiers, when we get / save a product.

On the GET API, we have `_modifier` fields.
On the POST one, we must include the modifier character within the value.

### DIFF
--- a/lib/src/model/nutriments.dart
+++ b/lib/src/model/nutriments.dart
@@ -17,7 +17,9 @@ class Nutriments extends JsonObject {
       );
 
       if (modifier != null) {
-        _modifierMap[_getModifierTag(nutrient)] = modifier;
+        _modifierMap[_getModifierTag(nutrient, PerSize.serving)] = modifier;
+        _modifierMap[_getModifierTag(nutrient, PerSize.oneHundredGrams)] =
+            modifier;
       }
 
       if (modifier == NutrientModifier.notProvided) {
@@ -61,7 +63,8 @@ class Nutriments extends JsonObject {
       '${nutrient.offTag}_${perSize.offTag}';
 
   /// Returns the modifier key for a [nutrient]
-  String _getModifierTag(final Nutrient nutrient) => nutrient.offTag;
+  String _getModifierTag(final Nutrient nutrient, final PerSize perSize) =>
+      '${nutrient.offTag}_${perSize.offTag}';
 
   /// Returns the value in grams of that [nutrient] for that [perSize].
   ///
@@ -70,8 +73,9 @@ class Nutriments extends JsonObject {
       _valueMap[_getTag(nutrient, perSize)];
 
   /// Returns the modifier for a [nutrient].
-  NutrientModifier? getModifier(final Nutrient nutrient) =>
-      _modifierMap[_getModifierTag(nutrient)];
+  NutrientModifier? getModifier(
+          final Nutrient nutrient, final PerSize perSize) =>
+      _modifierMap[_getModifierTag(nutrient, perSize)];
 
   /// Sets the value in grams of that [nutrient] for that [perSize].
   ///
@@ -89,12 +93,8 @@ class Nutriments extends JsonObject {
         );
       }
 
-      /// Reset all values
-      for (final PerSize perSize in PerSize.values) {
-        _valueMap.remove(_getTag(nutrient, perSize));
-        _valueMap.remove(_getTag(nutrient, perSize));
-      }
-      _modifierMap[_getModifierTag(nutrient)] = modifier!;
+      _valueMap.remove(_getTag(nutrient, perSize));
+      _modifierMap[_getModifierTag(nutrient, perSize)] = modifier!;
       return this;
     }
 
@@ -107,7 +107,7 @@ class Nutriments extends JsonObject {
     _valueMap[_getTag(nutrient, perSize)] = value;
 
     if (modifier != null) {
-      _modifierMap[_getModifierTag(nutrient)] = modifier;
+      _modifierMap[_getModifierTag(nutrient, perSize)] = modifier;
     }
     return this;
   }
@@ -158,11 +158,17 @@ class Nutriments extends JsonObject {
   Map<String, dynamic> toJson() {
     final Map<String, dynamic> result = <String, dynamic>{};
     for (final Nutrient nutrient in Nutrient.values) {
-      String modifierChar =
-          _modifierMap[_getModifierTag(nutrient)]?.offTag ?? '';
-
       for (final PerSize perSize in PerSize.values) {
+        final NutrientModifier? modifier =
+            _modifierMap[_getModifierTag(nutrient, perSize)];
+        final String modifierChar = modifier?.offTag ?? '';
+
         final String tag = _getTag(nutrient, perSize);
+        if (modifier == NutrientModifier.notProvided) {
+          result[tag] = NutrientModifier.notProvided.offTag;
+          continue;
+        }
+
         final double? value = _valueMap[tag];
         if (value != null) {
           result[tag] = '$modifierChar$value';
--- a/test/api_get_product_test.dart
+++ b/test/api_get_product_test.dart
@@ -230,9 +230,9 @@ void main() {
       const PerSize perSize = PerSize.oneHundredGrams;
 
       expect(nutriments.getValue(Nutrient.iron, perSize), 0.00041);
-      expect(nutriments.getModifier(Nutrient.iron), isNull);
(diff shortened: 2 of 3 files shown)


lib/src/model/nutriments.dart
```diff
@@ -43,21 +58,53 @@ class Nutriments extends JsonObject {
...
+    }
+
+    if (modifier != null && value == null) {
+      throw Exception(
+        'When the modifier is set, the value must not be null.',
+      );
+    }
+
+    _valueMap[_getTag(nutrient, perSize)] = value;
+    _modifierMap[nutrient.offTag] = modifier;
```
monsieurtanuki, line 107: Could a modifier be different for different `PerSize`s?

lib/src/model/nutriments.dart
```diff
@@ -11,15 +12,27 @@ class Nutriments extends JsonObject {
   /// Nutrient map with start values from [map].
   Nutriments._fromMap(final Map<String, dynamic> map) {
     for (final Nutrient nutrient in Nutrient.values) {
+      NutrientModifier? modifier = NutrientModifier.fromValue(
```
monsieurtanuki, line 15:
```suggestion
      final NutrientModifier? modifier = NutrientModifier.fromValue(
```

lib/src/model/nutriments.dart
```diff
@@ -34,7 +47,9 @@ class Nutriments extends JsonObject {
   ///
   /// It is useful to store null values: this way we can make the difference
   /// between totally unknown values and values that have been erased.
-  final Map<String, double?> _map = <String, double?>{};
+  final Map<String, double?> _valueMap = <String, double?>{};
+  final Map<String, NutrientModifier?> _modifierMap =
+      <String, NutrientModifier?>{};
```
monsieurtanuki, line 52:
```suggestion
  final Map<String, NutrientModifier> _modifierMap =
      <String, NutrientModifier>{};
```

test/api_get_product_test.dart
```diff
@@ -1430,4 +1432,81 @@ void main() {
...
+    final Nutriments nutriments = result.product!.nutriments!;
+
+    expect(nutriments.getModifier(Nutrient.energyKJ), isNull);
+    expect(nutriments.getModifier(Nutrient.energyKCal), isNull);
+    expect(nutriments.getModifier(Nutrient.sugars), isNull);
+    expect(nutriments.getModifier(Nutrient.salt), isNull);
+    expect(nutriments.getModifier(Nutrient.fiber), isNull);
+    expect(nutriments.getModifier(Nutrient.fat), isNull);
+    expect(nutriments.getModifier(Nutrient.saturatedFat), isNull);
+    expect(nutriments.getModifier(Nutrient.proteins), isNull);
```
monsieurtanuki, line 1456: Would be more fun if at least one modifier was not null. Or get rid of that test.

test/api_get_product_test.dart
```diff
@@ -1430,4 +1432,81 @@ void main() {
       expect(productResult.product!.dataQualityWarningsTags, isNull);
     });
   });
+
+  test('check nutrients modifier (set)', () async {
```
monsieurtanuki, line 1436: Can't understand the title, more specifically: "(set)".

monsieurtanuki (commented):
Hi @g123k!

I have a doubt about the relevancy of using the same modifier for all `PerSize`s of a given nutrient:
* I'm not sure it's always true
* That makes the code harder to read (the keys of the private maps being different), with no added value

Please also read my other comments.

lib/src/model/nutriments.dart
g123k: I'm not sure, but I can support it

test/api_get_product_test.dart
g123k: My goal was to test a product with no modifier.
But I can remove it, no pb

test/api_get_product_test.dart
g123k: I forgot to include a few changes.
Indeed, it wasn't understandable

g123k:
Hi @monsieurtanuki!

Thanks for your review.
We can send a different modifier per 100g or per serving.
However, we only receive one for the two.
I've changed the code to allow this.

test/api_save_product_test.dart
```diff
@@ -505,13 +510,150 @@ Like that:
...
           nutriments: Nutriments.empty()
             ..setValue(Nutrient.salt, PerSize.oneHundredGrams, 1.0));
 
       expect(product.noNutritionData, isFalse);
       expect(product.nutriments, isNotNull);
     });
+
+    test('Nutriments modifiers (server call)', () async {
+      Product product = Product(
+        barcode: '1234567890123',
```
monsieurtanuki, line 524: The problem is that you're always setting the same values for the same product. So if you're lucky it may work the first time, and if you're not lucky for some reason it doesn't work anymore (e.g. server somehow changed) and you won't be able to see that.
You may rather start with a `getProductV3`, and change each modifier.

test/api_save_product_test.dart
```diff
@@ -505,13 +510,150 @@ Like that:
...
+        uriHelper: uriHelper,
+      );
+
+      expect(result.product!.nutriments, isNotNull);
+      expect(
+        result.product!.nutriments?.getModifier(
+          Nutrient.proteins,
+          PerSize.oneHundredGrams,
+        ),
+        NutrientModifier.notProvided,
```
monsieurtanuki, line 579: Rather keeps the nutrient x modifier in a map, then save the product, then get the product, then check that the product matches the map.

monsieurtanuki (commented):
Thank you @g123k for your work!
My bad: unfortunately I was a bit wrong in my review, as I didn't get that the server provided only a single modifier for a nutrient, regardless of the `PerSize`.
Therefore it would make much more sense to remove the `PerSize` parameter from the `_getModifierTag` method. Good thing that this method does exist, though, as we just have to remove the parameter and let the syntactic analyzer detect the related errors. 

Please have a look at my comments regarding tests: in one case you play the same tests on the same product with the same modifiers. As a consequence you don't really test the "save modifier' feature, you test that it did work once. I suggest that you always change the modifier values, and that you include the `null` modifier in the "save modifier" test, as it's really a use case.

lib/src/model/nutriments.dart
```diff
@@ -11,15 +12,31 @@ class Nutriments extends JsonObject {
...
   Nutriments._fromMap(final Map<String, dynamic> map) {
     for (final Nutrient nutrient in Nutrient.values) {
+      final NutrientModifier? modifier = NutrientModifier.fromValue(
+        map['${nutrient.offTag}_modifier'],
+      );
+
+      if (modifier != null) {
+        _modifierMap[_getModifierTag(nutrient, PerSize.serving)] = modifier;
+        _modifierMap[_getModifierTag(nutrient, PerSize.oneHundredGrams)] =
+            modifier;
```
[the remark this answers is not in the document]
g123k: The problem is that the server may have a different value per serving AND 100g.
In this case, the last one will win.

Solution 1: keep the current behavior
Solution 2: save the value depending on the `PerSize` value of the array.

monsieurtanuki:
Hi @g123k!

I've just run some tests in PROD on the website, and I don't think there's such thing as "modifier for serving" - it looks like there's only one modifier for both serving and 100g, something like `'${nutrient.offTag}_modifier'`.
A bit like "unit".

The test was on "salt" for https://world.openfoodfacts.org/api/v3/product/8712000050368/?fields=nutriments
* salt "< 0.01" for 100g
```json
      "salt": 0.01,
      "salt_100g": 0.01,
      "salt_modifier": "\u003C",
      "salt_serving": 0.025,
      "salt_unit": "g",
      "salt_value": 0.01,
```
* salt "< 0.026" for serving
```json
      "salt": 0.026,
      "salt_100g": 0.0104,
      "salt_modifier": "\u003C",
      "salt_serving": 0.026,
      "salt_unit": "g",
      "salt_value": 0.026,
```
* salt "> 0.027" for serving
```json
      "salt": 0.027,
      "salt_100g": 0.0108,
      "salt_modifier": "\u003E",
      "salt_serving": 0.027,
      "salt_unit": "g",
      "salt_value": 0.027,
```
* salt "< 0.01" for 100g
```json
      "salt": 0.01,
      "salt_100g": 0.01,
      "salt_modifier": "\u003C",
      "salt_serving": 0.025,
      "salt_unit": "g",
      "salt_value": 0.01,
```

PR #1043, opened 2025-02-17
fix: ability to empty packagings list

monsieurtanuki:
### What
- When we converted Products to/from json, we lost the distinction between `null` and empty packagings list (field `packagings`).
- As a consequence, we couldn't force the packagings to be empty, as it was considered as a null array, and therefore considered as "not changed".
- With this PR we can have `null` or empty packagings list, as they do mean something different.
- Unit tests were added in order to check the impact of that minor change.

### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/6369

### Impacted files
* `json_conversion_test.dart`: added related unit tests
* `json_helper.dart`: now we can have non-null but empty packagings


teolemon (approved):
Thank you 🙂

monsieurtanuki:

Issue #1055, opened 2025-04-02
Add flexibility to Product according to api version


monsieurtanuki:
> @teolemon We currently use an `int` version number in off-dart. It looks like in some cases (`3.1`) we'd need a `double` (or a `String`), is that right?
> 
> From what I see in the off-dart code, we use the api version in only 2 cases, is that right?
> 1. single product search - `/api/v$version/product/$barcode/`
> 2. multi product search - `/cgi/search.pl` with parameter `api_version=$version`
> 
> We have 0 flexibility in dart for breaking changes:
> * If you add fields with new names, no problem but they'll be ignored
> * If you change the values of a field - e.g. `'fr:Carrefour'` instead of `'Carrefour'` - no problem
> * if you change the types of a field - e.g. `['Carrefour']` instead of `'Carrefour'` - we're in deep trouble
> 
> The thing is that we immediately transform the JSON into a dart class instance. If the changes do not fit into the old class structure, the code will crash. That also means that we would need to change off-dart first (in order to integrate the changes in the class structure) and then Smoothie. Very problematic maintainability.
> 
> That means we need to change paradigm in the way we deal with `Product`s in off-dart:
> * just store stupid JSON
> * extract specific fields (e.g. `getBrands`) from the JSON just-in-time
> * obviously add the `schema_version` field (cf. `"schema_version":1001` in https://fr.openfoodfacts.org/api/v3.2/product/4056489970279?fields=code)
> * add an option to retrieve "raw product fields" beyond our current list (like unexpected new "brands_tags" field to query)
> 
> Could be done while keeping the current mode
> * `Product`
> * `getProduct`
> * `searchProducts`
> by adding 
> * `FlexibleProduct`
> * `getFlexibleProduct`
> * `searchFlexibleProducts`
> 
> Interesting issue.
>  

 _Originally posted by @monsieurtanuki in [#6502](https://github.com/openfoodfacts/smooth-app/issues/6502#issuecomment-2766083181)_

teolemon:
- it seems that this is also related to search-a-licious data format: https://github.com/openfoodfacts/smooth-app/issues/4844#issuecomment-3015826935

Issue #900, opened 2024-04-08
We should be more forgiving on unexpected types, format changes in the knowledge panel API.


teolemon:
### Description
- We should be more forgiving on unexpected types, format changes in the knowledge panel API.
- Any unexpected input will crash the whole product page (it will talk about a network effort) (even attributes, basic details…)

### Proposed solutions
- Handle HTML links in HTML content so that we can demo the SignalConso DGCCRF integration on Wednesday
- Make the HTML attribute optional for the ACTION element (for now we added an empty HTML attribute to avoid the bug)
- Add a new button element that accepts an URL as a parameter (in coordination with @stephanegigandet )
- Make as many things optional as possible. Kill knowledge panel if they crash, but keep the rest of the JSON alive in the view.
- Try to be granular in crashes (eg not showing invalid ones, showing others)

The new Report a product panel showed that we fail in a spectacular way for small reasons. Knowledge panels shouldn’t crash an entire product.

monsieurtanuki:
@teolemon Part of it probably concerns Smoothie and not directly off-dart.
Do you have examples of products with "unexpected" KPs?

Regarding your suggestions...

> Handle HTML links in HTML content so that we can demo the SignalConso DGCCRF integration on Wednesday

Probably in Smoothie, but as far as I can see in the code (`smooth_html_widget.dart`) we already open links (used in KP text card, action card and table card). I've just tested and it does work (e.g. on a "informations missing about the ingredient origins" KP, click on the "producer" link)

> Make the HTML attribute optional for the ACTION element (for now we added an empty HTML attribute to avoid the bug)

Easy fix, to be coded on off-dart (make the field optional) and in Smoothie (don't display the html if `null`).

> Add a new button element that accepts an URL as a parameter (in coordination with @stephanegigandet )

Could be done easily, in off-dart then Smoothie.

> Make as many things optional as possible. Kill knowledge panel if they crash, but keep the rest of the JSON alive in the view.

The thing is that we need a proper JSON structure for `toJson`/`fromJson` operations.
The easiest fix about that would be to store KPs as a simple String, to be decoded with kindness, and not as a JSON.

> Try to be granular in crashes (eg not showing invalid ones, showing others)

I had a look at the KP mandatory fields, but they are not that many of them. Would the world be a much better place just making the rare mandatory fields as optional?
* `knowledge_panel.dart`:
  * `KnowledgePanel`: nothing is mandatory
[...]

Issue #833, opened 2023-11-22
Implement the new ElasticSearch search endpoint


teolemon:
### Why - Problem description

- As part of the [search revamp project](https://github.com/openfoodfacts/search-a-licious), we will be deploying a new Elasticsearch endpoint at [https://search.openfoodfacts.net/docs#/](https://search.openfoodfacts.net/docs#/)
- It will be OpenAPI compatible so we can generate a base Flutter SDK automatically that we can draw from here.
- We will also be upgrading the classic product-opener search to use this backend in the background (with an unchanged API vs what we've already implemented), so we should not deprecate immediately classic search. 
- We should also implement direct calls to the Elasticsearch endpoint.vThis will pave the way for a fuller search implementation in Smoothie.
### URL of the doc
https://search.openfoodfacts.net/docs#/

```[tasklist]
### Routes
- [ ] /document/{identifier}
- [ ] /search
- [ ] /autocomplete
```

```[tasklist]
### Schemas
- [ ] ErrorSearchResponseExpand allobject
- [ ] HTTPValidationErrorExpand allobject
- [ ] SearchResponseDebugExpand allobject
- [ ] SearchResponseErrorExpand allobject
- [ ] SuccessSearchResponseExpand allobject
- [ ] ValidationErrorExpand allobject
```

```[tasklist]
### Sub-issues already opened
- [ ] #620 
- [ ] #619 
```

monsieurtanuki:
My first tests show some problems/questions.

#### parameter discrepancies
In the current version
1. we pass the country as a parameter
=> shouldn't we do so Elastic Search?
1. we pass `app_name` as a parameter
=> shouldn't we do so Elastic Search?
1. we pass the user name in the http header
=> shouldn't we do so Elastic Search?

#### product structure discrepancies
In the [current version](https://world.openfoodfacts.org/api/v3/product/8002270014901?fields=code,brands&lc=fr)
4. the "brands" product field is a comma-separated `String`, and in [Elastic Search](https://search.openfoodfacts.net/search?q=pizza+brands%3Acarrefour&langs=fr&page_size=5&page=1&fields=code%2Cbrands) it's a `List<String>`.
=> it probably means that we need a specific new `Product` class.

raphael0202:
Hello @monsieurtanuki!
I'm sorry, I didn't see reply to your comment sooner.

> In the current version
>
>    we pass the country as a parameter
>   => shouldn't we do so Elastic Search?
[...]

PR #1056, opened 2025-04-06
feat: 1055 - new experimental class "FlexibleProduct"

monsieurtanuki:
### What
This PR adds flexibility:
* we can now use "double" API version number, e.g. `3.2`
* we can now extract product data from json on the fly, instead of formatting the json into a `Product` before anything else

With that PR we can easily accept new product fields and fields changing format, as we can just store the json regardless of the current or future product format.

### Fixes bug(s)
- Closes: #1055

### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/6502

### Files
New files:
* `api_get_flexible_product.dart`: tests around the new class `FlexibleProduct`.
* `flexible_map.dart`: (experimental) helper class around json and maps.
* `flexible_product.dart`: (experimental) Product without predefined structure, relying mainly on the json.
* `flexible_product_result.dart`: (experimental) API answer from a call to /api/v???/product/$barcode, in a flexible manner.
* `flexible_search_result.dart`: (experimental) API answer from a call product search, in a flexible manner.

Impacted files:
* `language_helper.dart`: added 3 experimental methods for `FlexibleProduct`
* `open_food_api_client.dart`: added 2 experimental methods for `FlexibleProduct` - `getFlexibleProductResult` and `searchFlexibleProducts`
* `openfoodfacts.dart`: added the new 4 class files
* `product_image.dart`: added an experimental method to compute the image URLs
* `product_query_configurations.dart`: made it possible to call API get product method with a "double" API version number - and not only an "int"
* `product_result_v3.dart`: minor refactoring
* `product_result_v3.g.dart`: generated


monsieurtanuki:
@teolemon ping

monsieurtanuki:
> @teolemon ping

ping

monsieurtanuki:
> > @teolemon ping
> 
> ping

ping

teolemon:
Sorry @monsieurtanuki , underwater on non mobile things. I asked for a technical quick look from @PrimaelQuemerais @g123k (who also seem to be a little bit underwater).

monsieurtanuki:
@teolemon For the record it is related to the app, cf. https://github.com/openfoodfacts/smooth-app/issues/6502
I don't know if that's a priority, though.

monsieurtanuki:
@teolemon @g123k @PrimaelQuemerais ping

PrimaelQuemerais:
Hello @monsieurtanuki 
The code looks good. I'm wondering what do you plan on using this for?
You give this example : `The typical use case would be "extending" this class in order to control brand new field values.`
But if we still update a class to add new fields why not add these new fields to Product directly?

monsieurtanuki:
@PrimaelQuemerais The typical use case is that 'brands' are a comma-separated String in OxF, and a List<String> in Searchalicious.
With the FlexibleProduct we don't give a damn: we just store a JSON locally and extract the brands one way or the other. In the current system, we need the Product to match exactly what we expect, format-wise. With the JSONSerializable syntax on top, which means we cannot support both data types at the same time.

Another use case is fields that we download but ignore forever because at the time of the download they weren't part of the Product structure.

And with an 'extends' it's easy to override or add fields.

lib/src/model/flexible/flexible_product.dart
```diff
@@ -0,0 +1,190 @@
...
+  String? get abbreviatedName => json['abbreviated_product_name'];
+
+  Additives? get additives =>
+      Additives.additivesFromJson(json['additives_tags'] as List?);
+
+  Iterable<AttributeGroup>? get attributeGroups =>
+      (json['attribute_groups'] as List<dynamic>?)
+          ?.map(AttributeGroup.fromJson);
+
+  String? get brands => json['brands'] as String?;
```
PrimaelQuemerais, line 51: Following on your example with brands.
Since FlexibleProduct is supposed to be used with both the OxF API and Searchalicious, wouldn't this throw if `json['brands']` is a List instead of String?

monsieurtanuki: @PrimaelQuemerais That's correct. But it's not relevant for the moment to implement anything specific to Searchalicious.
That first PR - that already started more than 3 months ago - is only about creating a FlexibleProduct that works like a Product.

PrimaelQuemerais: But if it doesn't implement this kind of logic I don't understand where this model stands between `Product` and storing the product in a `Map`. Do you already know how this would be handled in `FlexibleProduct` and why it couldn't be implemented in `Product` directly?

JSONSerializable still allows us to implement multiple parsing logics as long as the returning type remains the same, and wouldn't you face the same issue with `FlexibleProduct`?

Let me know if there's something I'm not getting here.

monsieurtanuki: > as long as the returning type remains the same

That's the problem for "brands".

No format problem will be faced with `FlexibleProduct`, assuming it's correct, JSON-wise.
What we cannot do today with `Product` is accepting _a priori_ data that don't match exactly requirements decided months ago.
Again, currently when we download products, we _first_ check if they match the expected `Product` structure. If they don't, we ignore them by throwing an exception.
With `FlexibleProduct`, we won't have this problem because as long as it's JSON, we accept the data. We'll still have to decide _later_ how to retrieve this or that field - e.g. for `brands` - but the data won't be rejected _a priori_. And extending `FlexibleProduct` e.g. in Smoothie will make it possible to be _flexible_ about the data.

Basically that would be something like
```dart
@override
String? get brands {
  dynamic result = json['brands'];
  if (result == null) {
    return null;
  }
  try {
    return result as String;
  } catch(e) {
    return (result as List<String>).join(',');
  }
}
```

PrimaelQuemerais: We can implement the same try catch in the Product deserialization logic which I think is what we should do.

This `FlexibleProduct` approach seems to introduce a lot of runtime errors as no-one expects a getter to throw.
If we want to manipulate products from the server without flexibly I think it is wiser to keep it as a Map until we are sure of the structure, at which point it should be serialized into an object, but then I don't see what prevents us from implementing everything into `Product` directly.

That's my take but it would be nice to get more opinions @g123k @teolemon

monsieurtanuki: > seems to introduce a lot of runtime errors

Does it?

> until we are sure of the structure

That's the point of being flexible: not having to wait for a stable structure (that may never happen), and being able to manage several structure versions (e.g. API versions) at the same time.

> I think it is wiser to keep it as a Map

It is a map: `class FlexibleProduct extends FlexibleMap`.

If you have a better solution that deals with the following issues, I'd be happy to review it and to close the current PR:
* #1055
* #900 (P0)
* #841 (P0)
* #833 (P0)
* https://github.com/openfoodfacts/smooth-app/issues/6502
* https://github.com/openfoodfacts/smooth-app/issues/4844 (P1)

PrimaelQuemerais: What I mean by "we are sure of the structure" is being able to identify the types of every field at runtime. The runtime errors are introduced by getters blindly casting fields of the Map.

It is not just a Map, `FlexibleProduct` takes the responsibility of field types through getters, while not being able to ensure it, this is what the issue is for me.

Alternatively, `FlexibleProduct` could check for the type in each getter and return null in case the value is not parseable. This has its own drawbacks but at least doesn't mislead users into thinking the type is ensured.

monsieurtanuki: > Okay let's reset and let me try to understand you better.

Cool

> Is the goal of `FlexibleProduct` to provide developers a way to get products from the API regardless of the structure of what the API sends back?

Correct. In order to provide flexibility to API structure changes.

> And it is better than a function returning a Map<String, dynamic> because it offers default getters?

Correct. So that developers don't have to reinvent the wheel for the 99% for fields that don't change.

PrimaelQuemerais: I'll take a bit more time to ponder over this but here's a few thoughts :

One improvement I would make then is to allow developers to pass their own type to `getFlexibleProductResult()` like this : `static Future<FlexibleProductResult> getFlexibleProductResult<T extends FlexibleProduct>(...)`. This way `FlexibleProductResult` can contain a list of their own type. And same for `searchFlexibleProducts()`.

One concern is if I extend `FlexibleProduct` to let's say fix the issue with `brands` being Strings or Lists, I won't be able to use the brand attribute as dart doesn't allow me to both override and change the type at the same time. How to we expect developers to work around this?

I still don't really like having getters which can throw, but that's a design choice and no solution will be perfect for this problem anyway, but maybe we can find a way to let developers know that these getters can be unsafe? I can take some time to think about it.

monsieurtanuki: > One improvement I would make then is to allow developers to pass their own type to `getFlexibleProductResult()` like this : `static Future<FlexibleProductResult> getFlexibleProductResult<T extends FlexibleProduct>(...)`. This way `FlexibleProductResult` can contain a list of their own type. And same for `searchFlexibleProducts()`.

I think that would be over-engineering the code:
* we already have access to `json`, so we can build whatever class on top of that json if needed
* in `dart`  we can extend `FlexibleProduct` whenever we want, including in Smoothie, so I can't think of anything we couldn't fix with an `extension`

> One concern is if I extend `FlexibleProduct` to let's say fix the issue with `brands` being Strings or Lists, I won't be able to use the brand attribute as dart doesn't allow me to both override and change the type at the same time. How to we expect developers to work around this?
> 
> I still don't really like having getters which can throw, but that's a design choice and no solution will be perfect for this problem anyway, but maybe we can find a way to let developers know that these getters can be unsafe? I can take some time to think about it.

In order to solve this issue forever, the idea would be to deprecate `brands` and to create 2 new getters, approximately coded like that:
```dart
  String? get brandsAsString {
    final result = json['brands'];
    if (result == null) {
      return null;
    }
    if (result is String) {
      return result;
    }
    return (result as List<String>).join(',');
  }

  List<String>? get brandsAsList {
    final result = json['brands'];
    if (result == null) {
      return null;
    }
    if (result is List<String>) {
      return result;
    }
    return (result as String).split(',');
  }
```

What do you think?

monsieurtanuki: @PrimaelQuemerais Indeed, we don't need `brands` after all. With `brandsAsString` and `brandsAsList` we're good.

monsieurtanuki:
@PrimaelQuemerais With my latest push:
* added the concept of `flexibleFields` - when the server provides new fields, we can query them before off-dart's `ProductField`s include those new fields
* added SCHEMA_VERSION as a product field
* changed getter `brands` into less ambiguous getters `brandsAsString` and `brandsAsList`

monsieurtanuki:
@PrimaelQuemerais ping

monsieurtanuki:
Closed as stale. Again.

Issue #1058, opened 2025-04-16
Add flexibility to PriceTotalStats


monsieurtanuki:
### Why - Problem description
Currently in off-dart, we support less than 20 fields for `PriceTotalStats`.
And there are more than 50 fields on the server (https://prices.openfoodfacts.org/api/v1/stats).
In order to add the fields we would need to code, push the code, have it reviewed and merged and released.
That's due to the way it was coded, relying on `JsonSerializable`.

### What - Proposed solution
A flexible solution would be to keep the JSON and just let the users get what they want.
Something like `int? getInt(String tag)`.
We may also provide a list of `static const String` typical tags. If there's a new tag, we wouldn't need to release again: the users would be able to query that new tag regardless.

### Additional context
cf. `FlexibleProduct` in #1056.
cf. https://github.com/openfoodfacts/smooth-app/issues/6128
cf. https://github.com/openfoodfacts/smooth-app/pull/6553

AshutoshKhadse23:
Hello @monsieurtanuki,
I've opened a PR that addresses all the checks mentioned in the issue and is ready for review.

monsieurtanuki:
@AshutoshKhadse23 For the record, you're supposed to ask to be assigned first.

AshutoshKhadse23:
Sorry for that but could you assign this to me !

PR #1059, opened 2025-04-17
feat: Add flexibility to PriceTotalStats

AshutoshKhadse23:
Closes: #1058 

Added all 50 fields.
And provide the flexibility.
This is how it will look :
```
final stats = PriceTotalStats.fromJson(jsonData);

final priceCount = stats.priceCount;

final someNewField = stats.getInt('some_new_field');

if (stats.containsKey('another_new_field')) {
  final value = stats.getString('another_new_field');
}

final allKeys = stats.availableKeys;
```


monsieurtanuki (commented):
Hi @AshutoshKhadse23!
As you may have already understood from https://github.com/openfoodfacts/smooth-app/pull/6553, your solution is a bit too complex.
Let's focus on https://github.com/openfoodfacts/smooth-app/pull/6553 first.

AshutoshKhadse23:
@monsieurtanuki ,
I tried the simpler approach but was having issues with the autogenerated file price_total_stats.g.dart

monsieurtanuki:
> @monsieurtanuki , I tried the simpler approach but was having issues with the autogenerated file price_total_stats.g.dart

@AshutoshKhadse23 Good news: we don't need `price_total_stats.g.dart` and `JsonSerializable` in that specific case.
More generally speaking, we're sort of migrating to a less strict structure, cf. #1056. We need more flexibility.

AshutoshKhadse23:
@monsieurtanuki ,
I’ve updated the code with a simpler approach, removing price_total_stats.g.dart and JsonSerializable as they’re no longer needed.

lib/src/prices/price_total_stats.dart
```diff
@@ -3,72 +3,164 @@ import 'package:json_annotation/json_annotation.dart';
...
+  int? get priceTagCount => getInt(PRICE_TAG_COUNT);
+  int? get priceTagStatusUnknownCount => getInt(PRICE_TAG_STATUS_UNKNOWN_COUNT);
+  int? get priceTagStatusLinkedToPriceCount =>
+      getInt(PRICE_TAG_STATUS_LINKED_TO_PRICE_COUNT);
+  int? get userCount => getInt(USER_COUNT);
+  int? get userWithPriceCount => getInt(USER_WITH_PRICE_COUNT);
+
+  DateTime? get updated => _rawData.containsKey(UPDATED)
+      ? JsonHelper.nullableStringTimestampToDate(_rawData[UPDATED])
+      : null;
```
monsieurtanuki, line 127:
```suggestion
  DateTime? get updated => getDateTime('updated');
```

AshutoshKhadse23:
@monsieurtanuki, I agree with you and changed the code with the getters and without the constants.

monsieurtanuki (approved):
Thank you @AshutoshKhadse23, and congratulations for your first PR in this specific project!

Issue #1038, opened 2025-02-10
TypeError: type 'int' is not a subtype of type 'String?' in type cast


g123k:
Issue found on the OFF app: https://github.com/openfoodfacts/smooth-app/issues/6284

Sentry Issue: [SMOOTHIE-4A1](https://openfoodfacts.sentry.io/issues/5873746071/?referrer=github_integration)

```
TypeError: type 'int' is not a subtype of type 'String?' in type cast
  File "product_result_field.g.dart", line 12, in _$ProductResultFieldFromJson
  File "product_result_field.dart", line 21, in new ProductResultField.fromJson
  File "product_result_field_answer.g.dart", line 14, in _$ProductResultFieldAnswerFromJson
  File "product_result_field_answer.dart", line 26, in new ProductResultFieldAnswer.fromJson
  File "product_result_v3.dart", line 83, in ProductResultV3._fromJsonListAnswerForField
...
(9 additional frame(s) were not displayed)
```

monsieurtanuki:
Thank you @stephanegigandet!
I've just managed to reproduce the issue.
The thing is that when there's a warning or an error returned by the API, we get the related value back.
And so far we always expected `String?` values. But if you try to set a negative quantity, like `-12`, then you get a warning that returns an `int`, not a `String?`, which is an incorrect cast.
Fixed in #1074.

PR #1074, opened 2025-04-30
fix: 1038 - feedback for incorrect packaging numbers

monsieurtanuki:
### What
- When we save packagings data, we use the v3 of the save product API.
- In this version, when there'a an error or a warning, we get a detailed feedback, including the input value.
- The problem was that if we sent incorrect numbers (e.g. negative weight), those numbers were sent back but we expected `String?`, which is an incorrect cast.
- The fix is to convert explicitly those feedback numbers to `String?`.

### Fixes bug(s)
- Closes: #1038

### Impacted files
* `api_save_product_v3_test.dart`: added a test specific to incorrect numbers sent by the user
* `json_helper.dart`: added a method for an explicit conversion to `String?`
* `product_result_field.dart`: explicit conversion to `String?` for some fields; added field `valuedConverted`
* `product_result_field.g.dart`: generated

### DIFF
--- a/lib/src/model/product_result_field.dart
+++ b/lib/src/model/product_result_field.dart
@@ -1,4 +1,5 @@
 import 'package:json_annotation/json_annotation.dart';
+import 'package:openfoodfacts/openfoodfacts.dart';
 import '../interface/json_object.dart';
 
 part 'product_result_field.g.dart';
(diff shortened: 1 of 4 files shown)


monsieurtanuki:

Issue #1062, opened 2025-04-19
Implement Prices Challenges


monsieurtanuki:
### Why - Problem description
Implement the Challenges (Prices):
* https://prices.openfoodfacts.org/api/docs#/challenges/challenges_list
* https://prices.openfoodfacts.org/api/docs#/challenges/challenges_retrieve

AshutoshKhadse23:
I am attaching work, please tell me that I am  in the right direction.

PR #1075, opened 2025-04-30
add: challenge list and retrive

AshutoshKhadse23:
Closes : #1062

I had implemented the same approach we did in stats.

### DIFF
--- a/lib/src/prices/get_challenge_list.dart
+++ b/lib/src/prices/get_challenge_list.dart
@@ -1,0 +1,64 @@
+import '../utils/json_helper.dart';
+
+class GetChallengeList {
+  final Map<String, dynamic> json;
+
+  GetChallengeList(this.json);
+
+  factory GetChallengeList.fromJson(Map<String, dynamic> json) {
+    return GetChallengeList(json);
+  }
+
+  List<List<ChallengeItem>>? get items {
+    if (json['items'] == null) return null;
+    final itemsList = json['items'] as List<dynamic>;
+    return itemsList.map((itemGroup) {
+      final itemGroupList = itemGroup as List<dynamic>;
+      return itemGroupList
+          .map((item) => ChallengeItem.fromJson(item as Map<String, dynamic>))
+          .toList();
+    }).toList();
+  }
+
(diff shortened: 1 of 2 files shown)


monsieurtanuki:
> @monsieurtanuki is the code mergeable, does it need improvements ?

I don't know, I haven't reviewed it.

lib/src/prices/get_challenge_list.dart
```diff
@@ -0,0 +1,69 @@
+import '../utils/json_helper.dart';
+
+/// Get Challenge List
+class GetChallengeList {
```
monsieurtanuki, line 4: Should probably be renamed `GetChallengesResult` (file and class), in order to match existing classes like `GetPriceProductsResult`

lib/src/prices/get_challenge_list.dart
```diff
@@ -0,0 +1,69 @@
...
+  int? get page => json['page'] as int?;
+  int? get pages => json['pages'] as int?;
+  int? get size => json['size'] as int?;
+  int? get total => json['total'] as int?;
+
+  Map<String, dynamic> toJson() => json;
+}
+
+/// Individual challenge item.
+class ChallengeItem {
```
monsieurtanuki, line 33: Probably deserves a specific file.

lib/src/prices/get_challenge_response.dart
```diff
@@ -0,0 +1,39 @@
+import '../utils/json_helper.dart';
+
+class GetChallengeResponse {
```
monsieurtanuki, line 3: Duplicate of `ChallengeItem`.
In order to match the other namings, should be named something like `Challenge` or `PriceChallenge` anyway.

monsieurtanuki (commented):
Hi @AshutoshKhadse23!
Sorry for my late review! I thought @g123k had started working on it.
Please have a look at my detailed comments.
Beyond that, we need a method to get the challenges, and we need tests to test it.

monsieurtanuki:
@AshutoshKhadse23 Still working on it?

Issue #1072, opened 2025-04-29
Add Nutri-Score Details


goerlitz:
### Why - Problem description
The result of an API request only exposes the Nutri-Score grade from the field `nutrition_grade_fr`.

Any additional Nutri-Score information is not exposed, like
- Nutri-Score version
- Nutri-Score category (water, beverage, fat/oil/nuts/seeds, red meat, cheese, general)
- Nutri-Score score
- reasons why Nutri-Score is not available, e.g.
   - not applicable -> not-applicable-category
   - missing data -> category or nutrients or ingredients

However, all that information is returned in the `nutriscore` field:

```json

        "nutriscore": {
            "2021": {
                "category_available": 1,
                "data": {
                    "energy": null,
                    "fiber": 0,
                    "fruits_vegetables_nuts_colza_walnut_olive_oils": 0,
                    "is_beverage": 1,
                    "is_cheese": 0,
                    "is_fat": 0,
                    "is_water": "1",
                    "proteins": null,
                    "saturated_fat": null,
                    "sodium": null,
                    "sugars": null
                },
                "grade": "not-applicable",
                "not_applicable_category": "en:alcoholic-beverages",
                "nutrients_available": 1,
                "nutriscore_applicable": 0,
                "nutriscore_computed": 0
            },
            "2023": {
                "category_available": 1,
                "data": {
                    "energy": null,
                    "fiber": null,
                    "fruits_vegetables_legumes": 0,
                    "is_beverage": 1,
                    "is_cheese": 0,
                    "is_fat_oil_nuts_seeds": 0,
                    "is_red_meat_product": 0,
                    "is_water": "1",
                    "non_nutritive_sweeteners": 0,
                    "proteins": null,
                    "salt": null,
                    "saturated_fat": null,
                    "sugars": null
                },
                "grade": "not-applicable",
                "not_applicable_category": "en:alcoholic-beverages",
                "nutrients_available": 1,
                "nutriscore_applicable": 0,
                "nutriscore_computed": 0
            }
        },
```

### What - Proposed solution
Make the data from the `nutriscore` field available as NutriScore class.

### Alternatives you've considered
There is not really an alternative.

### Additional context
Transparency is important for users, especially, when a Nutri-Score is not available. Then it should be possible to give an explanation why it is missing. But also to explain why a Nutri-Score is 'A' or 'E' an app needs access to the details in the `nutriscore` field.

goerlitz:
[...]

PR #1079, opened 2025-05-06
feat: 1072 - Add Nutri-Score details

goerlitz:
### What

- Introduce a developer-friendly Nutri-Score class to access and interpret the structured data from the `nutriscore` field.
- Part of https://github.com/openfoodfacts/openfoodfacts-dart/issues/1072

Example Input:
```json
"nutriscore": {
    "2021": {
        "grade": "not-applicable",
        "not_applicable_category": "en:alcoholic-beverages",
        "category_available": 1,
        "nutrients_available": 1,
        "data": {
            "is_beverage": 1,
            "is_cheese": 0,
            "is_fat": 0,
            "is_water": 0,
         },
         ...
    },
    "2023": {
        ...
    }
},
```

### Why

The raw nutriscore field includes detailed metadata such as:
- the version of the algorithm (2021, 2023)
- the computation status (unknown, not applicable, computed)
- the grade, score, and input completeness
- the food category (multiple `is_*` flags)

While this is valuable information, the structure is inconsistent and difficult to work with directly in app code.

### Problem

The raw data has several structural challenges:

- grade mixes status and score logic (not-applicable, a, b, etc.)
- Food categories are derived from multiple overlapping boolean flags
- Boolean values are inconsistently typed (0, "0", false)
- Apps must implement non-trivial logic to make this data usable

### Solution

This PR introduces a high-level Nutri-Score model that:
- Parses and interprets raw data into clear domain concepts
- Encapsulates version differences (2021 vs. 2023) in a type-safe way
- Provides meaningful enum types for status, grade, input requirements, and food category
- Enables structured access to displayable info like:
  - missing input
  - reason for inapplicability
  - food category

### Structure

File | Responsibility
-- | --
model/nutriscore_raw.dart | Parses raw JSON structure from API
model/nutriscore_enums.dart | Defines enums for versioned categories, grades, etc.
model/nutriscore_category.dart | Abstracts version-specific category logic
model/nutriscore.dart | High-level domain model with rich API
test/nutriscore_from_json_test.dart | Test coverage for malformed and valid data

### Notes

This implementation uses Dart 3.0 features for clarity and safety:
- switch expressions
- pattern matching
- sealed classes
- null safety

It covers the core Nutri-Score data (status, grade, score) and versioned category (2021, 2023). Further Nutri-Score computation details included in the `nutriscore` field have not been implemented (yet)

### Example

This example demonstrates how the high-level NutriScore class makes the data accessible in a developer-friendly way.

```dart
final nutriScore = product.nutriscores?[NutriScoreVersion.v2023];

final category = switch (nutriScore?.category?.as2023) {
  NutriScoreCategory2023.water => 'Water',
  NutriScoreCategory2023.beverage => 'Beverage',
  NutriScoreCategory2023.fatOilNutsSeeds => 'Fat/Oil/Nuts/Seeds',
  NutriScoreCategory2023.cheese => 'Cheese',
  NutriScoreCategory2023.redMeatProduct => 'Red Meat Product',
  NutriScoreCategory2023.general => 'General Food',
  _ => 'Unknown Category',
};

final infoText = switch (nutriScore?.status) {
  NutriScoreStatus.unknown =>
    'Unknown (missing data: ${nutriScore?.missingData?.join(', ') ?? 'n/a'})',
  NutriScoreStatus.notApplicable =>
    'Not Applicable (category: ${nutriScore?.notApplicableCategory ?? 'n/a'})',
  NutriScoreStatus.computed =>
    'Computed (grade: ${nutriScore?.grade}, score: ${nutriScore?.score}, category: $category)',
  NutriScoreStatus.invalid =>
    'Invalid (error: ${nutriScore?.errorMessage ?? 'unknown error'})',
  null => 'No Nutri-Score available',
};

print('Nutri-Score: $infoText');
```


lib/src/model/nutriscore.dart
```diff
@@ -0,0 +1,208 @@
...
+        NutriScore.invalid('Missing required fields (grade, category, score)'),
+      NutriScoreStatus.computed =>
+        NutriScore.computed(version, grade!, category!, score!),
+      _ => NutriScore.invalid('Unexpected NutriScore: ${raw.grade}'),
+    };
+  }
+
+  @override
+  String toString() {
+    return 'NutriScore{status: ${status.name}, grade: ${grade?.name}, version: $version, category: $category, score: $score, notApplicableCategory: $notApplicableCategory, errorMessage: $errorMessage, missingData: $missingData}';
```
monsieurtanuki, line 206: I'm sure we can afford some line jumps like that:
```suggestion
    return 'NutriScore{status: ${status.name}'
      ', grade: ${grade?.name}'
      ', version: $version'
      ', category: $category'
```

lib/src/model/nutriscore.dart
```diff
@@ -0,0 +1,208 @@
...
+    this.missingData,
+  });
+
+  /// Creates a [NutriScore] with [NutriScoreStatus.unknown].
+  ///
+  /// This indicates that the Nutri-Score could not be computed due to missing data,
+  /// but partial information such as [version] and [category] may still be present.
+  ///
+  /// The [missingData] parameter indicates which data is missing.
+  factory NutriScore.unknown([
```
monsieurtanuki, line 72: Just by curiosity: what's the added value of `factory` instead of named constructor in this file?

lib/src/model/product.dart
```diff
@@ -412,6 +414,14 @@ class Product extends JsonObject {
   @JsonKey(name: 'nutrition_grade_fr')
   String? nutriscore;
 
+  /// Map of computed Nutri-Scores for different versions (e.g. 2021, 2023).
+  @JsonKey(
+    name: 'nutriscore',
+    fromJson: NutriScore.fromJson,
+    includeToJson: false,
+  )
```
monsieurtanuki, line 422: We need a `toJson` too, as in Smoothie we spend our time storing products as json strings in the local database and retrieving them when relevant.

teolemon:
@stephanegigandet perhaps some inspiration for refactoring

lib/src/model/product.dart
```diff
@@ -412,6 +414,14 @@ class Product extends JsonObject {
...
+  Map<NutriScoreVersion, NutriScore>? nutriscores;
```
monsieurtanuki, line 423: I would name it `nutriscoreDetails` too, as you did as a `ProductField`. That way it would be easier to keep the track.

lib/src/model/nutriscore_raw.dart
```diff
@@ -0,0 +1,87 @@
...
+      if (isFatOilNutsSeeds) 'isFatOilNutsSeeds', // 2023
+      if (isRedMeatProduct) 'isRedMeatProduct', // 2023
+      if (isWater) 'isWater',
+    ];
+
+    return 'data(${flags.join(', ')})';
+  }
+}
+
+bool parseBool(Map<String, dynamic> json, String key) {
```
monsieurtanuki, line 84: We already have `JsonObject.parseBool`.

monsieurtanuki (commented):
Hi @goerlitz, and thank you for your clean code!
I could add other comments, but first I have to address an important point: you must provide a `toJson` method for your new "nutriscore details" product field. Or wait until #1056 is reviewed.
Besides, a test with server data would be great, too.

lib/src/model/nutriscore.dart
goerlitz: @monsieurtanuki , I intentionally left out `toJson` because the Nutri-Score details are computed on the server side and thus `nutriscore` should be a read-only field and not sent back to the server. Or is there a technical reason to always require `toJson`? In that case I would rather implement a dummy `toJson` that provides `null` or an empty object. What do you think?

Regarding server result tests and formatting: sure, I can add this.

monsieurtanuki: You're right, we don't send back this data to the server.
But, as I explained in another comment
> We need a `toJson` too, as in Smoothie we spend our time storing products as json strings in the local database and retrieving them when relevant

I meant a meaningful `toJson`.
Actually in Smoothie we never display data from the server. We get data from the server and then store it in our local database. Then, we extract data from the local database to display it.
For that we use both fromJson and toJson.

lib/src/model/product.dart
goerlitz: @monsieurtanuki, I intentionally left out `toJson` because the Nutri-Score details are computed on the server side and thus `nutriscore` should be a read-only field and not sent back to the server.

Is there a technical reason to always require `toJson` or is this just because the Smoothie app uses the `Product` object for local storage?

Actually, the `openfoodfacts-dart` library should be agnostic of what any client app does and not pollute the code with requirements that are just needed by a single app.

Of course, I can add a `toJson`. However, only a subset of the `nutriscore` data is currently parsed by `fromJson` (because the data is complex and convoluted). Also some fields are sanitized. Thus, the output of `toJson` would be different from the input. This may lead to issues if the assumption is that `fromJson` needs to work on data retrieved from the API as well as data retrieved from local storage.

What do you think?

lib/src/model/nutriscore.dart
goerlitz: Good question. I started with factories initially, but it is probably overkill as named constructors with simple field initialization should be enough. I'll check.

lib/src/model/nutriscore_raw.dart
goerlitz: Ah, I only found `JsonHelper.boolFromJSON` which is not appropriate.
I will replace with `JsonObject.parseBool`.

goerlitz:
@monsieurtanuki, I made the requested changes:
- replaced factories with named constructors
- using JsonObject.parseBool
- added toJson() and put all json-related code into new json helper class
- added tests for toJson() and fromJson()
- added == operator and hashcode
- added tests for server data
- other minor improvements

monsieurtanuki (commented):
Hi @goerlitz!
To be honest, I find that your code looks a bit too complex for what it's actually doing.

The first thing would be to use what we already provide in the package, like `OffTagged`.

The second thing would be to code what we actually need as users: two fields, one for nutriscore2021, the other for nutriscore2023. And probably without dart 3.0 ;)
If needed I'll send you code examples of what I mean.

lib/src/model/nutriscore_enums.dart
```diff
@@ -0,0 +1,43 @@
+enum NutriScoreStatus { unknown, notApplicable, computed, invalid }
+
+enum NutriScoreGrade { A, B, C, D, E }
+
+enum NutriScoreInput { category, nutrients, ingredients }
+
+enum NutriScoreVersion {
```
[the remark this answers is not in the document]
goerlitz: `enum NutriScoreInput { category, nutrients, ingredients }` are not tags in json.

But let me check if version could be used via OffTagged.

lib/src/model/nutriscore_raw.dart
```diff
@@ -0,0 +1,58 @@
...
+
+/// Raw model for additional data used in Nutri-Score 2021/2023 calculation,
+/// indicating specific product characteristics.
+class NutriScoreDataRaw {
+  final bool isBeverage;
+  final bool isCheese;
+  final bool isFat; // 2021
+  final bool isFatOilNutsSeeds; // 2023
+  final bool isRedMeatProduct; // 2023
+  final bool isWater;
```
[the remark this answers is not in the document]
goerlitz: The underlying JSON is not "very elegant", but let me check if I can handle this with OffTagged instead.

lib/src/model/nutriscore_raw.dart
```diff
@@ -0,0 +1,58 @@
...
+  NutriScoreDataRaw.fromJson(Map<String, dynamic> json)
+      : isBeverage = JsonObject.parseBool(json['is_beverage']),
+        isCheese = JsonObject.parseBool(json['is_cheese']),
+        isFat = JsonObject.parseBool(json['is_fat']),
+        isFatOilNutsSeeds = JsonObject.parseBool(json['is_fat_oil_nuts_seeds']),
+        isRedMeatProduct = JsonObject.parseBool(json['is_red_meat_product']),
+        isWater = JsonObject.parseBool(json['is_water']);
+
+  @override
+  String toString() {
```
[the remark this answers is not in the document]
goerlitz: no problem, I can remove the toString() - but it is handy for debugging, at least for NutriScore object. Not so much for the raw data, I agree.

lib/src/model/nutriscore_raw.dart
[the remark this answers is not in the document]
monsieurtanuki: What I meant is having to comment your code so that we know which field belong to what version isn't something I would recommend.
I'll probably send you the code as I imagined it, and you'll see how simple and elegant it can be. And easier to maintain of course, that's the whole point.

lib/src/model/nutriscore.dart
```diff
@@ -0,0 +1,149 @@
...
+          _unorderedEquals(missingData, other.missingData);
+
+  @override
+  int get hashCode => Object.hash(
+        status,
+        grade,
+        version,
+        category?.value,
+        score,
+        notApplicableCategory,
```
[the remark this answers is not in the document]
goerlitz: this is not for ordering - only equality. It is (primarily= used for the toJson/fromJson test to verify that writing and reading the same object gives the same result.

Alternatively, all fields could be compared but in the test - if you prefer.

lib/src/model/nutriscore_category.dart
```diff
@@ -0,0 +1,92 @@
...
+    };
+  }
+
+  @override
+  String toString() => name;
+}
+
+/// Internal wrapper for [NutriScoreCategory2021].
+/// Used by [NutriScoreCategory] to encapsulate version-specific category types.
+class _NutriScoreCategory2021 extends NutriScoreCategory {
```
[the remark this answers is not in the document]
monsieurtanuki: > I've been going back and forth with the modeling to find a good type safe implementation for the different versions and how it differs in the details. I agree that this part is a bit complicated. Let me check if I can simplify and avoid some over-engineering.

I'll try on my side and I'll show you. Again, I assume that a good type safe implementation is based on basic OOP. We'll see ;)

goerlitz:
> Hi @goerlitz! To be honest, I find that your code looks a bit too complex for what it's actually doing.
> 
> The first thing would be to use what we already provide in the package, like `OffTagged`.
> 
> The second thing would be to code what we actually need as users: two fields, one for nutriscore2021, the other for nutriscore2023. And probably without dart 3.0 ;) If needed I'll send you code examples of what I mean.

Actually, my app needs a simple way to access all of this Nutri-Score information - that why I implemented it.

- Nutri-Score computed? -> category, score, (positive and negative factors)?
- Nutri-Score not applicable? -> why? which category?
- Nutri-Score unknown? -> why? what is missing (category?, nutrients?, ingredients?)?

Also the OFF App falls short in making that information transparent. Hence, it would be highly beneficial to have it included in an easily accessible way (the `nutriscore` field/struct has all the information but fiddling with the underlying JSON format quickly gets messy and is error-prone).

Let me try to leverage `OffTagged` and simplify some code.

lib/src/model/nutriscore/nutriscore.dart
```diff
@@ -0,0 +1,135 @@
+import 'package:openfoodfacts/openfoodfacts.dart';
```
monsieurtanuki, line 1: Nope. Please use local paths.

lib/src/model/nutriscore/nutriscore.dart
```diff
@@ -0,0 +1,135 @@
...
+
+enum NutriScoreInput { category, nutrients, ingredients }
+
+typedef NutriScore2021 = _NutriScore<NutriScoreCategory2021>;
+typedef NutriScore2023 = _NutriScore<NutriScoreCategory2023>;
+
+/// Private base class for NutriScore.
+/// It is used to define the common properties and methods for both
+/// NutriScore 2021 and NutriScore 2023.
+class _NutriScore<T extends Enum> {
```
monsieurtanuki, line 30: I don't like this `<T extends Enum>`.

What about this?
```dart
abstract class NutriScoreCategory {}

enum NutriScoreCategory2021 implements NutriScoreCategory {
  cheese,
  fat,
  beverage,
  water,
}

enum NutriScoreCategory2023 implements NutriScoreCategory {
  cheese,
  redMeatProduct,
  fatOilNutsSeeds,
  beverage,
  water,
}

abstract class NutriScore {
  NutriScore({
    required this.score,
  });

  final int? score;

  NutriScoreCategory? get category;
}

class NutriScore2021 extends NutriScore {
  NutriScore2021({
    required super.score,
    required this.category,
  });

  @override
  NutriScoreCategory2021? category;
}

class NutriScore2023 extends NutriScore {
  NutriScore2023({
    required super.score,
    required this.category,
  });

  @override
  NutriScoreCategory2023? category;
}
```

lib/src/model/nutriscore/nutriscore.dart
```diff
@@ -0,0 +1,135 @@
...
+    this.category,
+    this.score,
+    this.notApplicableCategory,
+    this.missingData = const [],
+  }) : assert(
+          (grade != null && score != null && category != null) ||
+              (grade == null && (notApplicableCategory?.isNotEmpty ?? false)) ||
+              (grade == null && missingData.isNotEmpty),
+          'Either NutriScore is computed or not applicable or unknown',
+        );
```
monsieurtanuki, line 48: Not a big fan of `assert`s in general, and specifically complicated `assert`s.

lib/src/model/nutriscore/nutriscore.dart
```diff
@@ -0,0 +1,135 @@
...
+
+  bool get isComputed => grade != null;
+  bool get isNotApplicable => notApplicableCategory?.isNotEmpty ?? false;
+  bool get isUnknown => missingData.isNotEmpty;
```
monsieurtanuki, line 52: "isUnknown"? What about `hasMissingData`?

lib/src/model/nutriscore/nutriscore.dart
```diff
@@ -0,0 +1,135 @@
...
+    case 'd':
+      return NutriScoreGrade.D;
+    case 'e':
+      return NutriScoreGrade.E;
+    default:
+      return null;
+  }
+}
+
+List<NutriScoreInput> _missingData(bool? hasCategory, bool? hasNutrients) => [
```
monsieurtanuki, line 72: Please make it a `static` method.

monsieurtanuki (commented):
Hi @goerlitz!
Thank you for your work!

lib/src/model/nutriscore/nutriscore.dart
goerlitz: `isNotApplicable` could also have missing data, e.g. when missing nutrients but having category which is not applicable.
I prefer to have `isComputed`, `isNotApplicable` and `isUnknown` to be mutually exclusive. (but I realize that this is currently not the case with this implementation - I need to fix it).

That's why I initially had the `enum status {computed, notApplicable, unknown}` to clearly distinguid between the 3 states - currently this information is distributed in the JSON data between `nutriscore_computed` field and `not-applicable`+ `unknown` values in `grade`. :-/

lib/src/model/nutriscore/nutriscore.dart
goerlitz: I also don't like the assers that much but wanted to be safe here.
Well, we could say assume that the data is always correct (which it seems to be) or say that correctness should not be our concern here. Then I will just remove it.

monsieurtanuki: @goerlitz For the record `assert` works only in debug mode anyway.
And "correctness should not be our concern _here_" 👍

lib/src/model/nutriscore/nutriscore.dart
monsieurtanuki: @goerlitz I think you're trying to be too smart here, too related to your app needs.
If the "raw" data says it's computed, not applicable, with missing data, who are you to decide otherwise or which one has the priority?
Just keep it straightforward.
Then, in your own app, add an extension if you want.

goerlitz: Yes, that makes sense. It's probably better not to introduce too much logic at this level and instead handle it in the app layer.

That said, part of the motivation behind the abstraction is that the API model is quite unclear and loosely documented. It's often hard to determine the intended logic (e.g., precedence, or how to interpret certain field combinations), which makes implementations prone to guesswork and inconsistency. I was trying to guard against that.

I'll adjust the implementation accordingly.

lib/src/model/nutriscore/nutriscore.dart
goerlitz: Ah, I didn’t realize that Dart allows overriding an (implicit) field getter with a more specific (covariant) return type. That’s really nice — thanks for pointing it out!
Thus, the generic type is not needed for type safety.

lib/src/model/nutriscore/nutriscore.dart
goerlitz: I reintroduced the `NutriScoreStatus` enum (`computed, notApplicable, unknown`) for these reasons:

- Matches the raw API: The `grade` field conflates status and value (`'unknown'`, `'not-applicable'`, `'a'` – `'e'`). The enum reflects these states while clearly separating status from actual grades.

- Avoids ambiguity: The previous `isComputed`, `isNotApplicable`, and `hasMissingData` flags could overlap. The enum enforces mutual exclusivity by design.

- Improves clarity: A single status is easier to reason about than multiple booleans with implicit precedence rules.

The raw data remains accessible; this enum is a clean abstraction on top for consistent downstream use.

lib/src/model/nutriscore/nutriscore.dart
goerlitz: actually removed and modeled differently.

goerlitz:
@monsieurtanuki , thanks for all the great support and helpful review comments!

I’ve made several updates and added more documentation. I think the code now strikes a good balance between:
- Capturing the raw JSON structure in a way that’s consistent with the rest of the API (using JsonSerializable)
- Providing a strongly typed domain model on top to make working with the data easier and less error-prone

Final verdict:
- No Dart 3 features, no generics — just clean, practical code
- Logic is kept close to the raw data — not trying to be overly clever, just focusing on the parts where strong typing provides real value
- Apps are free to choose whether to use the domain model or work directly with the raw data

lib/src/model/nutriscore/nutriscore_details.dart
```diff
@@ -0,0 +1,67 @@
...
+  NutriScore2021? get nutriScore2021 {
+    if (rawNutriScore2021 == null) return null;
+
+    return NutriScore2021(
+      category: rawNutriScore2021?.data?.category,
+      grade: rawNutriScore2021?.grade,
+      score: rawNutriScore2021?.score,
+      notApplicableCategory: rawNutriScore2021?.notApplicableCategory,
+      categoryAvailable: rawNutriScore2021?.categoryAvailable ?? false,
+      nutrientsAvailable: rawNutriScore2021?.nutrientsAvailable ?? false,
```
monsieurtanuki, line 41: If it's null, it's null.
```suggestion
      categoryAvailable: rawNutriScore2021?.categoryAvailable,
      nutrientsAvailable: rawNutriScore2021?.nutrientsAvailable,
```

lib/src/model/nutriscore/nutriscore.dart
```diff
@@ -0,0 +1,141 @@
...
+  final int? score;
+
+  /// The version-specific food category used to evaluate the Nutri-Score.
+  final _NutriScoreCategory? category;
+
+  /// Specifies the product's category for which the Nutri-Score is not applicable.
+  final String? notApplicableCategory;
+
+  /// Indicates whether the category required to compute the Nutri-Score is available.
+  final bool categoryAvailable;
```
monsieurtanuki, line 70:
```suggestion
  final bool? categoryAvailable;
```

lib/src/model/nutriscore/nutriscore.dart
```diff
@@ -0,0 +1,141 @@
...
+
+  /// Indicates whether the nutrients required to compute the Nutri-Score are available.
+  final bool nutrientsAvailable;
```
monsieurtanuki, line 73:
```suggestion
  final bool? nutrientsAvailable;
```

lib/src/model/nutriscore/nutriscore_details.dart
```diff
@@ -0,0 +1,67 @@
...
+  /// Returns a strongly typed domain model specific to Nutri-Score 2021.
+  ///
+  /// Ensures that only valid 2021-specific categories and values are used.
+  /// Returns `null` if no raw data is available.
+  NutriScore2021? get nutriScore2021 {
+    if (rawNutriScore2021 == null) return null;
+
+    return NutriScore2021(
+      category: rawNutriScore2021?.data?.category,
+      grade: rawNutriScore2021?.grade,
```
monsieurtanuki, line 37:
```suggestion
      grade: NutriScoreGrade.fromOffTag(rawNutriScore2021?.grade),
```

lib/src/model/nutriscore/nutriscore_details.dart
```diff
@@ -0,0 +1,67 @@
...
+  /// Returns a strongly typed domain model specific to Nutri-Score 2023.
+  ///
+  /// Ensures that only valid 2023-specific categories and values are used.
+  /// Returns `null` if no raw data is available.
+  NutriScore2023? get nutriScore2023 {
+    if (rawNutriScore2023 == null) return null;
+
+    return NutriScore2023(
+      category: rawNutriScore2023?.data?.category,
+      grade: rawNutriScore2023?.grade,
```
monsieurtanuki, line 54:
```suggestion
      grade: NutriScoreGrade.fromOffTag(rawNutriScore2023?.grade),
```

lib/src/model/nutriscore/nutriscore.dart
```diff
@@ -0,0 +1,141 @@
+import '../off_tagged.dart';
+
+/// Internal marker interface for version-specific Nutri-Score categories.
+///
+/// This is used internally by the abstract [NutriScore] base class to enable
+/// shared handling of 2021 and 2023 category types. It is intentionally private
+/// to prevent accidental external usage or misimplementation.
+abstract class _NutriScoreCategory {}
```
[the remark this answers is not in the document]
goerlitz: Fine with me.

That said, keeping _NutriScoreCategory private enforces a stricter and more predictable API surface. Since Nutri-Score categories for 2021 and 2023 are standardized and not open for extension by clients, a private base class helps ensure correctness and discourages misuse.

If flexibility is preferred for future-proofing or unforeseen extensions, I'm happy to make it public, but from a design perspective, encapsulating it internally provides stronger guarantees.

Let me know how you'd like to proceed.

lib/src/model/nutriscore/nutriscore.dart
goerlitz: same here, what would be the difference between `null` and `false`?

lib/src/model/nutriscore/nutriscore.dart
```diff
@@ -0,0 +1,141 @@
...
+  NutriScore({
+    String? grade,
+    _NutriScoreCategory? category,
+    this.score,
+    this.notApplicableCategory,
+    this.categoryAvailable = false,
+    this.nutrientsAvailable = false,
+  })  : grade = OffTagged.fromOffTag(grade, NutriScoreGrade.values)
+            as NutriScoreGrade?,
+        category = categoryAvailable ? category : null;
```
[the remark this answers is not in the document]
monsieurtanuki: > Thanks for the feedback. I understand your position, and I appreciate the intent to keep the API client close to the raw server response for transparency and maintainability.

Awesome!

> That said, the server response is messy and highly developer-unfriendly.

Maybe. But what we need in a first approach is access to raw data.
We may need a logical layer on top, later, and we'll need your specific experience.

monsieurtanuki:
@goerlitz Still working on it?

monsieurtanuki:
@goerlitz Perfect! Please ping me when you're done with the boring code PR.

goerlitz:
Hi @monsieurtanuki,
I’ve completely removed the Nutri-Score domain model (i.e., the higher-level logical layer) in commit [073b70a](https://github.com/openfoodfacts/openfoodfacts-dart/pull/1079/commits/073b70ad1b88414bd0077aaae4d9d37bdeb38d15)

The remaining code in this PR is now strictly a thin wrapper around the raw Nutri-Score data as returned by the API - along with all of its known flaws and inconsistencies.

If adding a domain model is still something to consider in a future PR, I’d appreciate your input on the open questions in [lib/src/model/nutriscore/nutriscore.dart](https://github.com/openfoodfacts/openfoodfacts-dart/pull/1079/files/6a8efad465e42626952855cf9f58737902169a90), especially around how to best handle the inconsistencies with minimal logic.

Thanks again! Let me know if there’s anything else you’d like to see adjusted in this PR.

lib/src/model/product.dart
```diff
@@ -412,6 +413,9 @@ class Product extends JsonObject {
   @JsonKey(name: 'nutrition_grade_fr')
   String? nutriscore;
 
+  @JsonKey(name: 'nutriscore')
+  NutriScoreDetails? nutriScoreDetails;
```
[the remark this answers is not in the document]
monsieurtanuki: @goerlitz I don't think you addressed this comment, did you?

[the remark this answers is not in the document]
monsieurtanuki: @goerlitz So let me ask _again_, have you checked that
1. if you create a `Product` named `product`
2. and set its `nutriscoreDetails` to something not null
3. then store `product.toJson()` in variable `productJson`
4. then store `Product.fromJson(productJson)` in variable `productRestored`
5. and check what you have in `productRestored.nutriscoreDetails`, you have the same thing as in `product.nutriscoreDetails`?

goerlitz:
@monsieurtanuki did my comments ensure you that everything has been tested / or did you find the time to test yourself?

What else is holding off from merging?

monsieurtanuki:
@goerlitz What shall we do now?

PR #1080, opened 2025-05-08
feat: 1072 - new nutriscore details product field

monsieurtanuki:
### What
- @goerlitz This is what I had in mind. Very boring and predictable code.

### Part of 
- #1072

### Files
New files:
* `nutriscore_data_2021.dart`: Detailed data of NutriScore version 2021.
* `nutriscore_data_2021.g.dart`: generated
* `nutriscore_data_2023.dart`: Detailed data of NutriScore version 2023.
* `nutriscore_data_2023.g.dart`: generated
* `nutriscore_detail_2021.dart`: Data of NutriScore version 2021.
* `nutriscore_detail_2021.g.dart`: generated
* `nutriscore_detail_2023.dart`: Data of NutriScore version 2023.
* `nutriscore_detail_2023.g.dart`: generated
* `nutriscore_details.dart`: NutriScore detailed info.
* `nutriscore_details.g.dart`: generated
* `nutriscore_grade.dart`: Grade of NutriScore.

Impacted files:
* `api_get_product_test.dart`: new tests around nutriscoreDetails
* `product.dart`: new nutriscoreDetails field
* `product.g.dart`: generated
* `product_fields.dart`: new nutriscore field


monsieurtanuki:
@goerlitz I understand that you would add computed fields like "category2023" (e.g. water, beverage, fat/oil/nuts/seeds, red meat, cheese, general).
Are there other features that wouldn't be available using the current PR?

goerlitz:
> ### What
> * @goerlitz This is what I had in mind. Very boring and predictable code.
> 
> ### Part of
> * [Add Nutri-Score Details #1072](https://github.com/openfoodfacts/openfoodfacts-dart/issues/1072)
> 
> ### Files
> New files:
> 
> * `nutriscore_data_2021.dart`: Detailed data of NutriScore version 2021.
> * `nutriscore_data_2021.g.dart`: generated
> * `nutriscore_data_2023.dart`: Detailed data of NutriScore version 2023.
> * `nutriscore_data_2023.g.dart`: generated
> * `nutriscore_detail_2021.dart`: Data of NutriScore version 2021.
> * `nutriscore_detail_2021.g.dart`: generated
> * `nutriscore_detail_2023.dart`: Data of NutriScore version 2023.
> * `nutriscore_detail_2023.g.dart`: generated
> * `nutriscore_details.dart`: NutriScore detailed info.
> * `nutriscore_details.g.dart`: generated
> * `nutriscore_grade.dart`: Grade of NutriScore.
> 
> Impacted files:
> 
> * `api_get_product_test.dart`: new tests around nutriscoreDetails
> * `product.dart`: new nutriscoreDetails field
> * `product.g.dart`: generated
> * `product_fields.dart`: new nutriscore field

Thank @monsieurtanuki, I'll have a look :)

goerlitz:
> @goerlitz I understand that you would add computed fields like "category2023" (e.g. water, beverage, fat/oil/nuts/seeds, red meat, cheese, general). Are there other features that wouldn't be available using the current PR?

Let me check. The data structure is complex and has some redundant fields - some are only included under some specific conditions. Unfortunately, all not at all documented.

monsieurtanuki:
> Let me check. The data structure is complex and has some redundant fields - some are only included under some specific conditions. Unfortunately, all not at all documented.

I wouldn't call that complex (as in "predicting the weather for next month"). Reagrding documentation, I guess having a look at the code may help: https://github.com/openfoodfacts/openfoodfacts-server/blob/main/lib/ProductOpener/Nutriscore.pm

goerlitz:
> > Let me check. The data structure is complex and has some redundant fields - some are only included under some specific conditions. Unfortunately, all not at all documented.
> 
> I wouldn't call that complex (as in "predicting the weather for next month"). Reagrding documentation, I guess having a look at the code may help: https://github.com/openfoodfacts/openfoodfacts-server/blob/main/lib/ProductOpener/Nutriscore.pm

You are right, the data structure itself is not complex - it's rather a complicated computation (with 1000 lines of code) that make the data structure not easy to digest. I'd still say that the whole Nutri-Score model with the two versions is complex as there are challenges with food categorization, identifying ingredients and deriving ingredient percentages is definitely not straight forward.

monsieurtanuki:
@goerlitz Quickly said:
* Specific functional classes `NutriScore2021` and `NutriScore2023` make sense
* I would add a getter for both in `NutriScoreDetails` (or even `Product`), e.g. `NutriScore2021? get nutriScore2021`
* I would add an `enum` category for each "year", e.g. `enum NutriScoreCategory2021 {water, beverage, ...}` and related getters
* Regarding "Nutri-Score unknown? -> why? what is missing (category?, nutrients?, ingredients?)?", we already have the related `bool`s, haven't we?
* Regarding "Nutri-Score not applicable? -> why? which category?", we already have the related category String, haven't we?
* I don't like your "status" - or at least I don't understand that level of abstraction.
* Regarding "Nutri-Score computed? -> category, score, (positive and negative factors)?", we have at least the category and the score. The negative factor structure do differ according to versions.
* I guess a specific `enum` regarding "interesting nutrients" like "proteins" (as opposed to "vitamin b12" that is not used in the computations) would make sense.
* Definitely we should be able to get (and compare) both 2021 and 2023 nutriscores for the same product

What next? Who codes?

I believe your previous code - arguably adapted to your app - is a bit complicated. No offense, most of the time when I code for myself I often end up with code I have hard times reading again weeks later. And you may be influenced by your app code (e.g. "we _need_ dart 3.0")
I may code more simple, which is adapted for an open source project, and you have a better understanding of what users need.
If that's OK with you I can go on with my code, while listening to your functional suggestions.
If you prefer to start again using my PR code as the base it's fine with me too.
What do you prefer?

goerlitz:
Hi @monsieurtanuki,

## TL;TR
I rebased my code on your code and adapted and simplified my implementation.

In general, I think it makes a lot of sense to to have a clear separation between
1. your (low-level) code that tightly wraps the JSON from the API - like the rest of 'openfoodfacts-dart`
2. a higher-level (optional) semantic layer that gives a better developer experience when working with Nutri-Score details.

Thus, I think there is a good synergy of both parts.

## My Next Steps

- I found a few issues in you implementation -> I will add comments
- I adapted and simplified my code on top of your -> I will update my PR with these simplifications
- I intend to reduce some further complexity which is indeed useful for my app but not necessarily needed for the API implementation.

Some more comments:

> Specific functional classes NutriScore2021 and NutriScore2023 make sense
Absolutely :)

> I would add a getter for both in NutriScoreDetails (or even Product), e.g. NutriScore2021? get nutriScore2021
Yes

> I would add an enum category for each "year", e.g. enum NutriScoreCategory2021 {water, beverage, ...} and related getters
Yes, I have this in the (optional) semantic layer.

> Regarding "Nutri-Score unknown? -> why? what is missing (category?, nutrients?, ingredients?)?", we already have the related bools, haven't we?
In my opinion this is not well modeled in the data structure: The Nutri-Score cannot be computed when either category, nutrients or ingredients are missing. Hence, if grade=="unknown" any app has to check `category_available`, `nutrients_available` and .... ooops `ingredients_available` does not exist.

Two things in terms of good developer experience:
1. an app is not interested in what is available but what is missing or wrong if Nutri-Score is unknown
2. it would be much better to check a list of missing items that checking multiple bools.

> Regarding "Nutri-Score not applicable? -> why? which category?", we already have the related category String, haven't we?
Yes, this is fine.

> I don't like your "status" - or at least I don't understand that level of abstraction.
Yes, this is debatable. I can leave it out. Instead, I'd suggest to have getters `isComputed`, `isNotApplicable`, `isUnknown`.

> Regarding "Nutri-Score computed? -> category, score, (positive and negative factors)?", we have at least the category and the score. The negative factor structure do differ according to versions.
Yes, if `nutriscore_computed` and `grade` in A-E then all other values should be filled as well.

> I guess a specific enum regarding "interesting nutrients" like "proteins" (as opposed to "vitamin b12" that is not used in the computations) would make sense.
I would leave it out for now as there is no specific need yet. Of course, this can be added later. :)

goerlitz (commented):
I really like this approach using `@JsonSerializable()` which is much more in line with rest of the library implementation.
It also allows a clean separation between the raw JSON data implementation and an (optional) domain model for Nutri-Score on top.

A few things
- bug fixes for bool parsing and recursive JSON serialization
- I'd rather suggest to have the Grade enum in the domain model and use String for grade here
- minor formatting issues and improving test readability

monsieurtanuki:
@goerlitz As you now use that code in #1079, I'm not maintaining this PR.

PR #1085, opened 2025-05-27
feat: Added Get user data function OpenPricesAPIClient

chetanr25:
### What
- Added `getUserProfile` method to fetch user data from Open Prices API and return it as `PriceUser` model.
- This will be used in User Dashboard in order to replicate [Dashboard page - OpenFoodFacts](https://prices.openfoodfacts.org/dashboard)

### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/5206
- Supports PR https://github.com/openfoodfacts/smooth-app/pull/6559

### DIFF
--- a/lib/src/open_prices_api_client.dart
+++ b/lib/src/open_prices_api_client.dart
@@ -8,6 +8,7 @@ import 'package:path/path.dart';
 import 'prices/currency.dart';
 import 'prices/maybe_error.dart';
 import 'prices/price.dart';
+import 'prices/price_user.dart';
 import 'prices/proof.dart';
 import 'prices/get_locations_parameters.dart';
 import 'prices/get_locations_result.dart';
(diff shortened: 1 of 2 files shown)


lib/src/open_prices_api_client.dart
```diff
@@ -63,6 +64,28 @@ class OpenPricesAPIClient {
         addUserAgentParameters: addUserAgentParameters,
       );
 
+  static Future<MaybeError<PriceUser>> getUserProfile(
+    final String userId,
```
monsieurtanuki, line 68:
```suggestion
    final String username,
```
For consistency with the rest of the code.

lib/src/open_prices_api_client.dart
```diff
@@ -63,6 +64,28 @@ class OpenPricesAPIClient {
...
+  ) async {
+    final Uri uri = OpenPricesAPIClient.getUri(
+      path: '/api/v1/users/$userId',
+    );
+
+    final http.Response response =
+        await HttpHelper().doGetRequest(uri, uriHelper: uriHelperFoodProd);
```
monsieurtanuki, line 75: prod?

monsieurtanuki (commented):
Hi @chetanr25!
Some comments, please read them.

lib/src/open_prices_api_client.dart
monsieurtanuki: @chetanr25 All the other methods in this class have that uri helper parameter.
So please, add it to your new method.

chetanr25: Got it. 
I had made the changes.
Please let me know if any modifications is required.

lib/src/open_prices_api_client.dart
```diff
@@ -63,6 +64,29 @@ class OpenPricesAPIClient {
         addUserAgentParameters: addUserAgentParameters,
       );
 
+  static Future<MaybeError<PriceUser>> getUserProfile(
```
monsieurtanuki, line 67: I don't think "Profile" is a name found somewhere else in prices. Looks like you just added it.
Please don't add terms that may lead to confusion.
I suggest `getUser` instead, in order to mirror `getUsers`.

monsieurtanuki (commented):
Thank you @chetanr25 for your latest changes!
I still have a problem with the naming (please read my comments), and your method deserves a test - in `api_prices_test.dart`.

lib/src/open_prices_api_client.dart
chetanr25: I avoided using getUser earlier because I thought it might clash with `OpenFoodAPIConfiguration.getUser( )` and potentially cause confusion later on. That was the only reason I didn’t use it initially.

I’ve now made the suggested changes and committed them to this PR.

chetanr25:
> Thank you @chetanr25 for your latest changes! I still have a problem with the naming (please read my comments), and your method deserves a test - in `api_prices_test.dart`.

I’ve added some basic test cases, but I’m unsure about what username I must be using in them. Please let me know, and I’ll update it accordingly.

PR #1086, opened 2025-05-31
feat: add percentage type and name field to TitleElement

PrimaelQuemerais:
### What
Some fields for #9683 have been missing from `TitleElement` in `KnowledgePanels`
This PR adds the `name` field and the value `PERCENTAGE` to `TitleElementType`


monsieurtanuki (approved):
LGTM @PrimaelQuemerais!

PR #1091, opened 2025-06-09
feat: Add `source` to the `Proof` object

g123k:
Example:

```
{
  "id": 34413,
  "owner": "g123k",
  "source": "Smoothie - OpenFoodFacts (4.20.1+1481) (ios+Version 18.5 (Build 22F76))"
}
```


lib/src/model/old_product_result.g.dart
monsieurtanuki, line 1: The point of the comment is to get rid of warnings. I know you didn't explicitly removed the comment, it's just the code generation, but we somehow need this "ignore" comment. I don't know how it passed the test!

Another solution would be to remove `OldProductResult` altogether, that is deprecated from 2024-11-28 and that we don't use in Smoothie anymore anyway.

`info: 'OldProductResult' is deprecated and shouldn't be used. Use getProductV3 and ProductResultV3 instead. (deprecated_member_use_from_same_package at [openfoodfacts] lib\src\model\old_product_result.g.dart:11)`

monsieurtanuki (approved):
Hi @g123k!
I don't know what you expect from that new "source" field, but for the record with your code you won't be able to set or update its value.
For that you'd need to change `CommonProofParameters` too.

monsieurtanuki:
> I believe it's mostly for display purposes @monsieurtanuki

If so a "read-only" comment would be appropriate.

g123k:
Sorry for my late reply.
The read-only implementation is indeed for display purposes:

Détails d'une preuve

I add the comment to my PR.

lib/src/model/old_product_result.g.dart
monsieurtanuki: Probably still to be addressed.

monsieurtanuki (approved):
LGTM

g123k:
Thanks for the review 😎

Issue #1099, opened 2025-06-24
Folksonomy "get product tag not found" format change


monsieurtanuki:
### Description
It looks like the Folksonomy response for "get product tag __is not found__" (`'product/$barcode/$key'`) changed from "null" to "[]".
That causes an error when we try to find something we cannot find.
```
❌ test/api_folksonomy_test.dart: OpenFoodAPIClient Folksonomy getProductTag - not found (failed)
  type 'List<dynamic>' is not a subtype of type 'Map<String, dynamic>' in type cast
  package:openfoodfacts/src/folksonomy.dart 198:48  FolksonomyAPIClient.getProductTag
  ===== asynchronous gap ===========================
  test/api_folksonomy_test.dart 165:36              main.<fn>.<fn>
```

PR #1102, opened 2025-06-24
fix: 1099 - additional test for "folksonomy get product tag": "not found"

monsieurtanuki:
### What
- Now a "not found" response for "folksonomy get product tag" may be "[]" (and not only "null").

### Fixes bug(s)
- Fixes: #1099


monsieurtanuki:

PR #1121, opened 2025-08-08
feat(knowledge-panels): Implement 2 new subcases for knowledge panels actions

teolemon:
### What
- feat(knowledge-panels): Implement 2 new subcases for knowledge panels actions

This commit adds two new actions to the `KnowledgePanelAction` enum:
- `editProduct` (with offTag 'edit_product')
- `reportProductToNutripatrol` (with offTag 'report_product_to_nutripatrol')

A new test case has been added to `test/knowledge_panels_from_json_test.dart` to ensure that these new actions are correctly parsed from the JSON representation of a knowledge panel.


monsieurtanuki:
@teolemon Still a draft?

monsieurtanuki (approved):
@teolemon Looks good to me!
I don't know why you kept this PR as draft, and as already said our current use of KP actions is a bit stupid, but beyond that this PR adds 2 KP actions, and that's good enough.

monsieurtanuki:
> @monsieurtanuki I'd be eager to see the fix in Smooth App, as it's degrading the XP in the "Report a problem" tab

As explained [here](https://github.com/openfoodfacts/openfoodfacts-dart/pull/1121#issuecomment-3350349890) there's no added value in this PR. It doesn't solve anything in Smoothie, where we could already have handled the extra 2 cases, and where we still have to code something.

PR #1124, opened 2025-08-20
chore: upgrade to json_serializable 6.9.5

monsieurtanuki:
### What
- We delayed the upgrade of `json_serializable` to 6.9.5 as there were unwanted changes.
- Now that we also have an issue with `test` (cf. https://github.com/monsieurtanuki/openfoodfacts-dart/actions/runs/17098201851/job/48487610527), that's the good moment to upgrade, as hopefully the `test` issue will be fixed in the process

### Part of 
- #1116
- #1114
- #1113

### Impacted files
* re-generated `.g.dart` files - most of the time no change except Mac/Window CR/LF
* `pubspec.yaml`: upgraded json_serializable 6.9.5 and the corresponding dependencies


teolemon (approved):
Approving in principle. cc @openfoodfacts/open-food-facts-mobile-team

PR #1139, opened 2025-09-17
feat: 1055 - API versions as decimal, Product.schemaVersion

monsieurtanuki:
### What
- Now we can use decimal API versions - e.g. `3.2` - instead of just `int`
- New related product field `schemaVersion`

### Part of 
- #1055

### Impacted files
* `api_get_product_test.dart`: added a test about API versions matching schema versions
* `product.dart`: new field `schemaVersion`
* `product.g.dart`: generated
* `product_fields.dart`: new product field `SCHEMA_VERSION`
* `product_query_configurations.dart`: now the version number can be decimal

### DIFF
--- a/lib/src/model/product.dart
+++ b/lib/src/model/product.dart
@@ -96,6 +96,10 @@ class Product extends JsonObject {
   @JsonKey(name: 'product_type')
   ProductType? productType;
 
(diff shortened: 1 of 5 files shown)


monsieurtanuki:

Issue #1089, opened 2025-06-02
Flutter Web support broken due to missing CORS headers from API server


Turskyi:
### Description
I'm using the latest version of the `openfoodfacts` Dart package in a Flutter app targeting Web. When trying to retrieve product information using `OpenFoodAPIClient.getProductV3(...)`, I receive a `ClientException` with a CORS error in the browser console.

The package declares Web as a supported platform, but the API responses are being blocked by CORS, making it unusable in a Flutter web app.

### Expected behavior
I expected the API call to return product data as it does on mobile platforms, or at least for the documentation to clarify that Web is not supported due to CORS limitations. Alternatively, the library could use a proxy endpoint or offer guidance for Web compatibility.

### Stacktraces
ClientException:
Message: Failed to fetch
URI: https://world.openfoodfacts.org/api/v3/product/8081744404806/?lc=en&tags_lc=en&cc=ca&app_name=ethical_scanner&app_version=0.1.6&app_platform=Web&comment=Ethical+Scanner+is+a+mobile+application+that+scans+the+barcode+of+a+product+and+tells+you+if+the+product+meets+your+ethical+standards.
Stacktrace: package:http/src/browser_client.dart 72:30 
Browser console: Access to fetch at 'https://world.openfoodfacts.org/api/v3/product/060383047184/?lc=en&tags_lc=en&cc=ca&app_name=ethical_scanner&app_version=0.1.5&app_platform=Web&comment=Ethical+Scanner+is+a+mobile+application+that+scans+the+barcode+of+a+product+and+tells+you+if+the+product+meets+your+ethical+standards.' from origin 'https://ethical-scanner.com' has been blocked by CORS policy: Request header field from is not allowed by Access-Control-Allow-Headers in preflight response.

### Package information
- `openfoodfacts` version: **3.22.0**
- Flutter version: `3.32.1`
- Platform: Flutter Web

### Additional context
This issue occurs only on Flutter Web. On Android and iOS, the same code works as expected.

I'm using the method as described in the package's example:
```dart
OpenFoodAPIClient.getProductV3(
  ProductQueryConfiguration(
    code,
    language: OpenFoodFactsLanguage.ENGLISH,
    fields: <ProductField>[ProductField.ALL],
    version: ProductQueryVersion.v3,
  ),
);
```
[...]

PR #1145, opened 2025-10-02
fix: 1089 - removed FROM from http header - for web only

monsieurtanuki:
### What
- The `From` part of the header isn't well appreciated - CORS-wise - by web implementations.
- As it's only there for stat purposes, we can get rid of it.

### Fixes bug(s)
- Fixes: #1089

Thanks to @oznecniV97 @alexgarel @Turskyi for their contributions.


monsieurtanuki:

Issue #1134, opened 2025-09-09
Get ready to implement the unwanted ingredients feature in attributes


teolemon:
### Why - Problem description
- https://github.com/openfoodfacts/openfoodfacts-server/pull/12321
- The initial proposal is linked in the PR
- Proper OpenAPI documentation is incoming

monsieurtanuki:
@teolemon Proper OpenAPI doc is needed. Your search link points to a simple "ham" search, unrelated to unwanted ingredients.
Meanwhile, just created #1135.

monsieurtanuki:
Still not working:
* [No garlic: 0](https://fr.openfoodfacts.org/cgi/search.pl?search_terms=pizza&langs=fr&page_size=5&page=1&fields=code%2Cbrands&json=1&attribute_unwanted_ingredients_tags=en:garlic)
* [No whatever: 0](https://fr.openfoodfacts.org/cgi/search.pl?search_terms=pizza&langs=fr&page_size=5&page=1&fields=code%2Cbrands&json=1&attribute_unwanted_ingredients_tags=en:gadsdsdsdsfdfdfdrlic)
* [No restriction: +6k](https://fr.openfoodfacts.org/cgi/search.pl?search_terms=pizza&langs=fr&page_size=5&page=1&fields=code%2Cbrands&json=1)

PR #1163, opened 2025-11-11
feat: 1134 - added search parameters "unwanted ingredients" and "ingredients filter"

monsieurtanuki:
### What
Added 2 search parameters:
* "unwanted ingredients", that will populate a specific attribute group and give a good (100), bad (0) or unknown (null) match to be integrated in the product score, if some "unwanted" ingredients are actually in the product
* "ingredients filter", that will exclude from the search result products that have/don't have a list of ingredients

### Fixes bug(s)
- Closes: #1134

### Files
New files:
* `ingredients_filter_parameter.dart`: List of ingredients tags to filter in and out.
* `ingredients_unwanted_parameter.dart`: "Unwanted Ingredients" search API parameter.

Impacted files:
* `api_get_product_test.dart`: added a test about "unwanted ingredients"
* `api_search_products_test.dart`: added tests about 'unwanted ingredients" and "ingredients filter"
* `openfoodfacts.dart`: added the 2 new files
* `product_query_configurations.dart`: added an optional `unwantedIngredients` parameter to single-product queries

### DIFF
--- a/lib/src/external/external_source_filter.dart
+++ b/lib/src/external/external_source_filter.dart
@@ -27,17 +27,15 @@ class ExternalSourceFilter extends JsonMap {
     final Product product,
     final OpenFoodFactsLanguage language,
     final OpenFoodFactsCountry country,
-  ) =>
-      <String>[
-        if (!acceptsString(product.productType?.offTag, productTypes))
-          'productType(${product.productType?.offTag})',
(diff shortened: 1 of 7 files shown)


teolemon (approved):
approving on my side. might need a technical review from @g123k or @PrimaelQuemerais

Issue #1158, opened 2025-11-06
New field(s) and more resilience for `PriceProduct`


monsieurtanuki:
### Why - Problem description
For `PriceProduct`:
* add a `String? quantity` field (no rush)
* change field `productQuantity` from `int?` to `num?`
* if possible, make the `PriceProduct` class more resilient regarding future unexpected changes. Could be relatively easy given that it's read-only data.

### What - Proposed solution
A solution around `JsonMap`.

### Additional context
As usual, any change on the Prices API may have a dramatic impact on off-dart, e.g. app crashing when field formats change.

cc. @raphodn

PR #1164, opened 2025-11-11
feat: 1158 - num quantity and new quantityString fields in PriceProduct

monsieurtanuki:
### What
- We're now a bit more resilient with PriceProduct fields: field data is computed on the fly at runtime only if needed.
- The typical use-case is that field "product_quantity" evolves from `int` to `num`
  - we don't use that field at all in Smoothie
  - but it can still make the app crash (a `num` may not fit into an `int`), as we tried to compute ALL fields at the same time
- Regarding field changes
  - "product_quantity" (`quantity`) is now a `num?`
  - new field "quantity" (`quantityString`) as a `String?`
  - I know, there's some collision with field names...

cc. @raphodn @raphael0202 

### Fixes bug(s)
- Closes: #1158

### Files
New file:
* `json_map.dart`: Simple reusable json map.

Deleted file:
* `price_product.g.dart`

Impacted files:
* `api_prices_test.dart`: added a test for num field quantity and new field quantityString
* `flavor.dart`: added a standard method
* `openfoodfacts.dart`: added the new file
* `price_product.dart`: now relying on `JsonMap` instead of `JsonSerializable`, so that data is computed on the fly at runtime - more resilience, no crash if a field we don't use change types; upgraded field `quantity` from `int` to `num`; added field `quantityString`

### DIFF
--- a/lib/openfoodfacts.dart
+++ b/lib/openfoodfacts.dart
@@ -40,6 +40,8 @@ export 'src/model/packaging.dart';
 export 'src/model/parameter/allergens_parameter.dart';
 export 'src/model/parameter/barcode_parameter.dart';
 export 'src/model/parameter/ingredients_analysis_parameter.dart';
+export 'src/model/parameter/ingredients_filter_parameter.dart';
+export 'src/model/parameter/ingredients_unwanted_parameter.dart';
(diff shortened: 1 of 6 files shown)


monsieurtanuki:
Closed as stale.

PR #1188, opened 2026-01-22
fix: currency fix decimals methods

monsieurtanuki:
### What
- Side-effect fix for currency decimals, given that for Prices we need price values with a number of decimals that matches the currency.
- And given that in dart `4.6 * 100 == 459.99999999999994` 😞 

### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/7379

### Impacted files
* `api_prices_test.dart`: tests for the new methods
* `currency.dart`: new tested methods `fixDecimals` and `fixDecimalsForNumbers`

### DIFF
--- a/lib/src/prices/currency.dart
+++ b/lib/src/prices/currency.dart
@@ -1,3 +1,7 @@
+import 'dart:math';
+
+import 'package:meta/meta.dart';
+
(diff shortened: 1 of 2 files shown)


g123k (approved):
👌

PR #1027, opened 2025-01-24
feat: Add excludedItems parameter to autocomplete function with dynamic size adjustment

AffanShaikhsurab:
### What
- Added `excludedItems` parameter to filter out unwanted items from auto-suggestions.  
- Dynamically adjusted the `size` parameter to request additional items based on the length of the exclusion list.  
- Ensured safety by handling potential null values in `excludedItems`.  
- Improved overall flexibility of the autocomplete functionality.

### Fixes bug(s)
#889 

### Part of 
- [#5096](https://github.com/openfoodfacts/smooth-app/issues/5096)

PR #1211, opened 2026-03-19
fix: Exclude items from autosuggestions with dynamic limit adjustment (#889)

### LINKED ISSUE #889: Exclude items from autosuggestions
### Why - Problem description
cf. https://github.com/openfoodfacts/smooth-app/issues/5096 by @g123k:
> In many parts of the edition, we have an auto-completion feature.
However, this is a nonsense to list an item… which is already selected.

### What - Proposed solution
The first thing to do would be to add a parameter to the autosuggestion methods: a list of items to exclude.
Then, if the developer asks for 10 items and we exclude 2 items, we should ask the server for 10+2 = 12 items and return the first 10 items that are not in the 2-item list.

### Alternatives you've considered
Of course if the server could accept a list of items to exclude it would be better.

Phanindra899:
## Description

- This PR fixes issue #889 by implementing proper exclusion logic for autosuggestions.

## Problem
When requesting suggestions and excluding items, the app was returning fewer items than requested because it filtered AFTER asking the API for a fixed amount.

Example:
- Request: 10 suggestions
- Exclude: 3 items (apple, banana, mango)
- Old behavior: Asked API for 10 → filtered 3 → returned 7 items ❌
- New behavior: Asked API for 13 → filtered 3 → returns 10 items ✅

## Solution
Modified `autocomplete()` method in `lib/src/open_food_search_api_client.dart`:

1. Added optional `excludedItems` parameter with default empty list
2. Calculate adjusted size: `adjustedSize = size + excludedItems.length`
3. Request extra items from API using `adjustedSize`
4. Filter out excluded items from the response
5. Limit final result back to originally requested size with `.take(size)`
6. Preserve original response metadata

## Changes
- Modified: `lib/src/open_food_search_api_client.dart`
  - Added `excludedItems` parameter to `autocomplete()` method
  - Implemented dynamic size adjustment logic
  - Added filtering for excluded items
  - Added comments explaining the logic

## Backward Compatibility
 ##Fully backward compatible
- New parameter has default value (empty list)
- Behavior unchanged when `excludedItems` is not provided
- No breaking changes to API

## References
Fixes #889
References: PR #1027 (by @AffanShaikhsurab - incomplete implementation)

### DIFF
--- a/lib/src/open_food_search_api_client.dart
+++ b/lib/src/open_food_search_api_client.dart
@@ -32,6 +32,7 @@ class OpenFoodSearchAPIClient {
     final User? user,
     final int size = 10,
     final Fuzziness fuzziness = Fuzziness.none,
+    final List<String> excludedItems = const <String>[],
     final UriProductHelper uriHelper = uriHelperFoodProd,
   }) async {
     query = query.trim();
@@ -45,6 +46,9 @@ class OpenFoodSearchAPIClient {
     if (taxonomyTags.isEmpty) {
       throw Exception('Taxonomies cannot be empty!');
     }
+    // Ask the API for enough results to keep the requested page size
+    // even after excluded suggestions are removed locally.
+    final int adjustedSize = size + excludedItems.length;
     final Uri uri = uriHelper.getUri(
       path: '/autocomplete',
       forcedHost: _getHost(uriHelper),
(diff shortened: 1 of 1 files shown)


monsieurtanuki (commented):
@Phanindra899 Don't you find it strange that you edited just one file, when the previous PR edited 8 files?

PR #1219, opened 2026-04-06
fix: 1217 - clean error when incorrect credentials

### LINKED ISSUE #1217: When the log in fails, it throws a "bad format" exception
### Description
When we log in (`login2`) with incorrect credentials, we get a lousy error message:
`type 'Null' is not a subtype of type 'Map<String, dynamic>'`

### Expected behavior
When the credentials aren't correct, we should get a normal "wrong user/password" message.

### Additional context
The problem comes from here:
```dart
  factory LoginStatus.fromJson(
    Map<String, dynamic> json, [
    Map<String, String>? headers,
  ]) {
    return LoginStatus(
      status: JsonObject.parseInt(json['status'])!,
      statusVerbose: json['status_verbose'] as String,
      userId: json['user_id'] as String?,
      cookie: headers?['set-cookie'],
      userDetails: UserDetails.fromJson(json['user']), // not populated with wrong credentials
    );
  }
```

This is an important issue, and a quick win.

monsieurtanuki:
### What
Now we have a "clean" error when logging in with incorrect credentials:
* status = 0
* statusVerbose = "user not signed-in"
 
Instead of a incomprehensible `type 'Null' is not a subtype of type 'Map<String, dynamic>'`.

The bug was introduced in #1186, package 3.30.0.

### Fixes bug(s)
- Fixes: #1217

### DIFF
--- a/lib/src/model/login_status.dart
+++ b/lib/src/model/login_status.dart
@@ -73,7 +73,7 @@ class LoginStatus {
       statusVerbose: json['status_verbose'] as String,
       userId: json['user_id'] as String?,
       cookie: headers?['set-cookie'],
-      userDetails: UserDetails.fromJson(json['user']),
+      userDetails: UserDetails.fromJson(json['user'] ?? {}),
     );
   }
 


teolemon (approved):
Thank you 🥚

monsieurtanuki:
Thank you @teolemon for your review! 🍫

PR #1233, opened 2026-06-05
fix: error handling of API calls

### LINKED ISSUE #1069: Fix error handling of API calls
### Description
Non 200 HTTP status codes returned from the API endpoint are not handled properly.
For instance, a `504 Gateway Time-out` will throw a JSON parse Exception

### Expected behavior
The API client should check the HTTP status code and handle non 200 codes properly and not attempt to parse the result as JSON.

### Package information
openfoodfacts-dart 3.20

### Additional context

Problematic code in OpenFoodAPIClient, e.g. for `search`

```dart
  static Future<SearchResult> searchProducts(
    final User? user,
    final AbstractQueryConfiguration configuration, {
    final UriProductHelper uriHelper = uriHelperFoodProd,
  }) async {
    final Response response = await configuration.getResponse(user, uriHelper);
    // --> TODO: check response status code here before trying to parse JSON
    TooManyRequestsException.check(response);
    final String jsonStr = _replaceQuotes(response.body);
    final SearchResult result = SearchResult.fromJson(
      HttpHelper().jsonDecode(jsonStr),
    );
    _removeImages(result, configuration);
    return result;
  }
```

sarthikadhawan:
# Fix handling of HTTP responses before JSON parsing

## Summary

This PR improves error handling for API calls by validating HTTP responses before attempting to parse the response body as JSON.

Previously, some non-success responses (for example, `504 Gateway Time-out`) could return an HTML error page. The client would still attempt to call `jsonDecode()`, resulting in errors such as:

```text
Exception: JSON expected, html found
```

Instead, the client should detect non-success HTTP responses and throw a dedicated exception before JSON parsing is attempted.

## Changes

- Add a new `HttpStatusException` that includes:
  - HTTP status code
  - Error message
- Introduce a shared response validation method in the API client.
- Preserve existing special-case handling for `TooManyRequestsException`.
- Throw `HttpStatusException` for non-2xx responses before calling `jsonDecode()`.
- Add regression tests covering HTML error responses such as `504 Gateway Time-out`.

## Before

```text
HTTP 504 response
    ↓
jsonDecode(response.body)
    ↓
Exception: JSON expected, html found
```

## After

```text
HTTP 504 response
    ↓
Response validation
    ↓
HttpStatusException(statusCode: 504)
```

### Success responses (2xx)

No behavior change.

### Too many requests (429)

Existing `TooManyRequestsException` handling is preserved.

### Other non-success responses

The client now throws `HttpStatusException` instead of attempting to parse a non-JSON response body.

## Motivation

This change prevents confusing JSON parsing failures when the server returns HTML error pages and provides consumers of the SDK with a structured exception that can be handled appropriately based on the HTTP status code.

## Testing

Added regression tests to verify that:

- 2xx responses continue to behave as before.
- Existing `TooManyRequestsException` behavior remains unchanged.
- HTML error responses such as `504 Gateway Time-out` result in `HttpStatusException` rather than a JSON parsing exception.

## Related Issue

Fixes #1069

### DIFF
--- a/lib/src/open_food_api_client.dart
+++ b/lib/src/open_food_api_client.dart
@@ -281,10 +281,7 @@ class OpenFoodAPIClient {
   }) async {
     final Response response = await configuration.getResponse(user, uriHelper);
     TooManyRequestsException.check(response);
-    if (response.statusCode == 404) {
-      return response.body;
-    }
-    _checkResponse(response);
+    _checkResponse(response, includeCodes: [404]);
     return response.body;
   }
 
@@ -1466,7 +1463,10 @@ class OpenFoodAPIClient {
     }
   }
 
-  static void _checkResponse(final Response response) {
-    HttpStatusException.check(response);
+  static void _checkResponse(
+    final Response response, {
+    List<int> includeCodes = const [],
+  }) {
+    HttpStatusException.check(response, includeCodes: includeCodes);
   }
 }
--- a/lib/src/open_food_api_client.dart
+++ b/lib/src/open_food_api_client.dart
@@ -280,6 +280,7 @@ class OpenFoodAPIClient {
     final UriProductHelper uriHelper = uriHelperFoodProd,
   }) async {
     final Response response = await configuration.getResponse(user, uriHelper);
+    TooManyRequestsException.check(response);
     _checkResponse(response);
     return response.body;
   }
@@ -444,6 +445,7 @@ class OpenFoodAPIClient {
     final UriProductHelper uriHelper = uriHelperFoodProd,
   }) async {
     final Response response = await configuration.getResponse(user, uriHelper);
+    TooManyRequestsException.check(response);
     _checkResponse(response);
     final String jsonStr = _replaceQuotes(response.body);
     final SearchResult result = SearchResult.fromJson(
@@ -1462,7 +1464,6 @@ class OpenFoodAPIClient {
   }
 
   static void _checkResponse(final Response response) {
-    TooManyRequestsException.check(response);
     HttpStatusException.check(response);
   }
 }
--- a/lib/src/open_food_api_client.dart
+++ b/lib/src/open_food_api_client.dart
@@ -49,6 +49,7 @@ import 'utils/tag_type.dart';
 import 'utils/taxonomy_query_configuration.dart';
 import 'utils/too_many_requests_exception.dart';
 import 'utils/uri_helper.dart';
+import 'utils/http_status_exception.dart';
 
 /// Client calls of the Open Food Facts API
 class OpenFoodAPIClient {
@@ -112,6 +113,7 @@ class OpenFoodAPIClient {
       uriHelper: uriHelper,
       addCredentialsToBody: true,
     );
+    _checkResponse(response);
     return Status.fromApiResponse(response.body);
   }
 
@@ -161,6 +163,7 @@ class OpenFoodAPIClient {
       user,
       uriHelper: uriHelper,
     );
+    _checkResponse(response);
     return ProductResultV3.fromJson(HttpHelper().jsonDecode(response.body));
   }
 
@@ -277,7 +280,7 @@ class OpenFoodAPIClient {
     final UriProductHelper uriHelper = uriHelperFoodProd,
   }) async {
     final Response response = await configuration.getResponse(user, uriHelper);
-    TooManyRequestsException.check(response);
+    _checkResponse(response);
     return response.body;
   }
 
@@ -441,7 +444,7 @@ class OpenFoodAPIClient {
     final UriProductHelper uriHelper = uriHelperFoodProd,
   }) async {
     final Response response = await configuration.getResponse(user, uriHelper);
-    TooManyRequestsException.check(response);
+    _checkResponse(response);
     final String jsonStr = _replaceQuotes(response.body);
     final SearchResult result = SearchResult.fromJson(
       HttpHelper().jsonDecode(jsonStr),
@@ -503,6 +506,8 @@ class OpenFoodAPIClient {
       uriHelper: uriHelper,
       addCredentialsToBody: false,
     );
+
+    _checkResponse(response);
 
     Map<String, dynamic> decodedJson =
         HttpHelper().jsonDecode(_replaceQuotes(response.body))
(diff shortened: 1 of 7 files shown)


test/api_get_product_test.dart
```diff
@@ -539,8 +539,10 @@ void main() {
           fields: [ProductField.ALL],
           version: version,
         );
-        ProductResultV3 result = await getProductV3InProd(configurations);
-        expect(result.product, isNull);
+        expect(
+          () async => getProductV3InProd(configurations),
+          throwsA(isA<HttpStatusException>()),
+        );
```
monsieurtanuki: You mean looking for an unknown barcode is supposed to throw an exception?

lib/src/robot_off_api_client.dart
```diff
@@ -239,4 +247,9 @@ class RobotoffAPIClient {
     }
     return result.join(',');
   }
+
+  static void _checkResponse(final Response response) {
+    TooManyRequestsException.check(response);
```
monsieurtanuki: There are no "too many requests" for robotoff.

lib/src/open_food_search_api_client.dart
```diff
@@ -61,6 +63,8 @@ class OpenFoodSearchAPIClient {
       user: user,
       uriHelper: uriHelper,
     );
+    TooManyRequestsException.check(response);
```
monsieurtanuki: There are no "too many requests" exceptions for food search.

lib/src/open_food_api_client.dart
```diff
@@ -1461,4 +1460,9 @@ class OpenFoodAPIClient {
       );
     }
   }
+
+  static void _checkResponse(final Response response) {
+    TooManyRequestsException.check(response);
```
monsieurtanuki: Actually there are "too many requests" exception only in 2 cases.

monsieurtanuki (commented):
Thank you @sarthikadhawan for your PR!
Generally speaking, that's ok.

What's bothering me:
* we somehow have two ways in the library to handle exceptions, with "your" exceptions and with MaybeError. One single way would make more sense. Nothing we can do about it in the short term given our bandwidth, so that's ok.
* it looks like you throw an exception when the user searches for an unknown barcode, which is a bit unfriendly
=> please have a look at it
* not 100% sure about the status codes returned by the server - your 2xx choice makes sense but we should have a scrutinous look at the failing tests
* "too many requests" is triggered by the server in only 2 cases, that are already handled
=> let's not spill it all over the place, and keep it in those 2 cases

cc. @goerlitz

test/api_get_product_test.dart
sarthikadhawan: server gives 404 for it, and app can catch the error and update the UI to show error message to user accordingly

lib/src/robot_off_api_client.dart
sarthikadhawan: updated

lib/src/open_food_search_api_client.dart
sarthikadhawan: updated

lib/src/open_food_api_client.dart
sarthikadhawan: updated

test/api_get_product_test.dart
monsieurtanuki: The package is already in production. We cannot decide to change a typical use case like this one, in stealth mode.

sarthikadhawan: Yes, I agree. And if I avoid HTTP status check in this function then it might again bring back same problem of json decode failing, do you have any suggestions on how can we make it backward compatible?

monsieurtanuki: @sarthikadhawan Before your PR, there was no problem with "product not found". There was a 404, but there was a valid body.
I guess the solution includes an `if(statusCode == 404)` test.

sarthikadhawan: There can be other 4xx error codes, is it sufficient to exclude only 404 from HttpStatusException's condition?

monsieurtanuki: Only 404, and only for getProduct.
We need to track 429, remember.

sarthikadhawan: updated

lib/src/utils/http_status_exception.dart
```diff
@@ -0,0 +1,21 @@
+import 'package:http/http.dart';
+
+/// Generic exception for non-success HTTP responses.
+class HttpStatusException implements Exception {
+  const HttpStatusException({required this.statusCode, required this.message});
+
+  final int statusCode;
+  final String message;
+
+  static void check(final Response response) {
```
monsieurtanuki:
```suggestion
  static void check(final Response response, {List<int> includeCodes = []}) {
```
Would be more elegant like that in order to deal with the 404 case.

lib/src/open_food_api_client.dart
```diff
@@ -278,6 +281,10 @@ class OpenFoodAPIClient {
   }) async {
     final Response response = await configuration.getResponse(user, uriHelper);
     TooManyRequestsException.check(response);
+    if (response.statusCode == 404) {
+      return response.body;
+    }
+    _checkResponse(response);
```
monsieurtanuki:
```suggestion
    _checkResponse(response, includeCodes: [404]);
```

monsieurtanuki (commented):
Thank you @sarthikadhawan for your changes!
The most important being: checking that if we're looking for an unknown barcode, we don't get an exception.

lib/src/open_food_api_client.dart
sarthikadhawan: updated, thanks!

test/http_status_exception_test.dart
monsieurtanuki, line 1: Not convinced by the "local" tests like "I put 200, I see 200", but why not.

Would be more interesting trying to get "https://fnsdfhsjdhfsdfhjsdhfsd" and catch the exception.
And more specifically, please try to get an unknown barcode like "Unknown_barcode".

[Looks like I hadn't attached that comment to the rest of them]

sarthikadhawan: Hi @monsieurtanuki, Unknown barcode test exists in https://github.com/openfoodfacts/openfoodfacts-dart/blob/ffffc20c1db4173ed16a2e6ecc54e84775b3c996/test/api_search_products_test.dart#L43

test/http_status_exception_test.dart is for testing http_status_exception.dart hence the tests just match statusCodes and messages. E2E tests for APIs exist in their specific test files and are passing like the unknown barcode one

sarthikadhawan: Hi @monsieurtanuki, let me know if any other tests are required here

monsieurtanuki (approved):
Hi @sarthikadhawan!
I have a different point of view regarding the relevancy of tests, but good enough is good enough.
Thank you and congratulations for your first contribution here!

PR #1240, opened 2026-06-29
fix: 1239 - throws attempts to set/crop image as OTHER

### LINKED ISSUE #1239: Reject attempts to set a picture as "other"
### Description
The server doesn't accept when we try to _set_ (`cgi/product_image_crop.pl`) an image as `OTHER_$language`.
Fun fact: we still can _upload_ (`cgi/product_image_upload.pl`) an image as `OTHER_$language`. It will be uploaded, but won't be considered as `OTHER_$language`.

### Expected behavior
We should explicitly reject attempts to set an image as `OTHER_$language`, instead of letting the server process that call and reject it with its own message.
In Smoothie, we shouldn't even call the _set_ method for OTHER. As we call this method only to set an existing image as `$field_$language`, we can just skip that step: for OTHER, we just want the image to be uploaded, and it already is.

### Additional context
* https://github.com/openfoodfacts/smooth-app/issues/7363
* https://github.com/openfoodfacts/openfoodfacts-server/pull/13630

monsieurtanuki:
### What
The server doesn't accept setting or cropping an existing image as OTHER.
Now, we throw an exception when the method is called with OTHER.

### Fixes bug(s)
- Fixes: #1239

### Part of 
- https://github.com/openfoodfacts/smooth-app/issues/7363
- https://github.com/openfoodfacts/openfoodfacts-server/pull/13630

### DIFF
--- a/lib/src/open_food_api_client.dart
+++ b/lib/src/open_food_api_client.dart
@@ -1281,6 +1281,9 @@ class OpenFoodAPIClient {
     required final User user,
     required final UriProductHelper uriHelper,
   }) async {
+    if (imageField == ImageField.OTHER) {
+      throw Exception('You cannot set or crop an image as OTHER');
+    }
     final String id = '${imageField.offTag}_${language.offTag}';
     final Map<String, String> queryParameters = <String, String>{
       'code': barcode,


teolemon (approved):
thank you @monsieurtanuki

monsieurtanuki:

PR #1253, opened 2026-07-27
feat: 1252 - API 3.4 support

### LINKED ISSUE #1252: Support API 3.4
### Why - Problem description
We currently support API 3.3.
There are currently 3 higher versions (up to 3.6) that we don't support.

[2025-06-11 - Product version 1002 (no change) - API version 3.4 - New attributes with parameters](https://openfoodfacts.github.io/openfoodfacts-server/api/ref-api-and-product-schema-change-log/#2025-06-11-product-version-1002-no-change-api-version-34-new-attributes-with-parameters)

Breaking changes:

* product attributes can now have parameters, such as a list of tags (used for the new Unwanted ingredients attribute)

To avoid displaying product attributes with parameters that apps cannot handle yet, the /api/v[0-2]/attributes_groups API does not return attributes with parameters.

This does not affect the product READ API, if the attribute parameters are sent, then the value of the attribute will be returned by the product READ API (e.g. /api/v2/product) even if the version is less than 3.4.

monsieurtanuki:
### What
Actually, we already used specifically version 3.4 for its added value: product attribute parameters.
No big change on that part: we use 3.4 implicitly everywhere, not just for product attribute parameters.
But that was a good opportunity to refactor around the api versions, including price, robotoff and taxonomy.

### Fixes bug(s)
- Closes: #1252

### Files
New file:
* `api_version.dart`: Api version.

Impacted files:
* `available_attribute_groups.dart`: now using an API version
* `available_preference_importances.dart`: now using an API version
* `open_prices_api_client.dart`: now using an API version
* `openfoodfacts.dart`: added new file `api_version.dart`
* `product_query_configurations.dart`: minor refactoring
* `product_query_version.dart`: now extending new class `ApiVersion`; upgrade to API 3.4
* `robot_off_api_client.dart`: now using an API version
* `taxonomy_query_configuration.dart`: now using an API version

### DIFF
--- a/lib/openfoodfacts.dart
+++ b/lib/openfoodfacts.dart
@@ -152,6 +152,7 @@ export 'src/search/fuzziness.dart';
 export 'src/search/taxonomy_name.dart';
 export 'src/search/taxonomy_name_autocompleter.dart';
 export 'src/utils/abstract_query_configuration.dart';
+export 'src/utils/api_version.dart';
 export 'src/utils/autocomplete_manager.dart';
 export 'src/utils/autocompleter.dart';
 export 'src/utils/barcodes_validator.dart';
--- a/lib/src/open_food_api_client.dart
+++ b/lib/src/open_food_api_client.dart
@@ -36,6 +36,7 @@ import 'model/taxonomy_packaging_shape.dart';
 import 'model/user.dart';
 import 'prices/maybe_error.dart';
 import 'utils/abstract_query_configuration.dart';
+import 'utils/api_version.dart';
 import 'utils/country_helper.dart';
 import 'utils/http_helper.dart';
 import 'utils/language_helper.dart';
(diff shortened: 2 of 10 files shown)


teolemon (commented):
Very nice refacto. I don't spot the API versions mapping for prices and robotoff. They are harder to spot than for OFF

monsieurtanuki:
> Very nice refacto. I don't spot the API versions mapping for prices and robotoff. They are harder to spot than for OFF

Thanks!
Not sure what you mean here: in Prices we have `static const ApiVersion defaultVersion = ApiVersion(1);`, in RobotOff we also have `static const ApiVersion defaultVersion = ApiVersion(1);` - but not the same `1` ;)

Next time I work on Prices I'll probably switch it to real OOP - as opposed to `static` methods.
That may make the API version more obvious.
Something like that:
```dart
class OpenPricesAPIClient {
  OpenPricesAPIClient({
    this.uriHelper = uriHelperFoodProd,
    this.version = defaultVersion,
  });

  final UriProductHelper uriHelper;
  final ApiVersion version;

  Future<MaybeError<GetPricesResult>> getPrices(
    final GetPricesParameters parameters, {
    final String? bearerToken,
  }) async {
    // ...
  }
```

teolemon (approved):
Ok, I was looking too quickly, and was expecting something like

monsieurtanuki:
Thank you @teolemon for your review and comments!

PR #1274, opened 2026-09-14
fix: workflow permissions for: .github/workflows/pubdev-score.yml

teolemon:
Potential fix for [https://github.com/openfoodfacts/openfoodfacts-dart/security/code-scanning/42](https://github.com/openfoodfacts/openfoodfacts-dart/security/code-scanning/42)

Add an explicit `permissions` block to the workflow root in `.github/workflows/pubdev-score.yml`, just after the trigger section (`on:`), so it applies to all jobs unless overridden.  
For this workflow, the minimal and appropriate permission is:

- `contents: read`

This supports `actions/checkout` and read-only repository access while preventing broader write scopes from default settings. No imports, methods, or dependencies are needed; this is a YAML-only change.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._


monsieurtanuki (approved):
Hi @teolemon!
I don't know anything about actions, but restricting permissions to read-only for specific actions makes sense.

